Issue #1, opened 2021-03-03
Add a testing infrastructure


hug-dev:
The few tests in `cryptoki/src/lib.rs` do not even compile. They were written with `rust-pkcs11` infrastructure in mind. We need to check if they are still valid, need to be modified and make `cargo test` pass locally and on the CI.

More generally we need to add more test and have a proper strategy for it!

PR #3, opened 2021-03-04
Add more object classes; fix tests

nickray:
I need to handle certificates in `pkcs11-uri`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,8 +8,33 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Execute CI script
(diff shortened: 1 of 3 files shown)


nickray:
Seems this ci.sh is missing, can you add it? Otherwise I might add some basic `cargo test` calls in the workflow (this is not soo bad given https://github.com/nektos/act exists, but a shell script is fine too).

hug-dev (approved):
First PR 🥳 Thank you!!
Looks good to me! We can export those commands in a `ci.sh` script if we prefer later. We did that for our other repo because we thought it was easier to maintain the script directly than the workflow commands. But until then I am ok with it!
I guess you just fixed #1 as well, great!

ionut-arm (approved):

PR #11, opened 2021-03-24
Remove armv7 bindings

ionut-arm:
This commit removes the `armv7-unknown-linux-gnueabi` bindings, since
all `arm` targets will actually share the same bindings.


hug-dev (approved):
Thanks! Maybe that means that we could allow more target triples in the `build.rs` check.

PR #18, opened 2021-04-01
Add a new way to check for supported targets

### LINKED ISSUE #15: Supported targets might not need an exact target triple check
When we select the bindings that we want, we only check `target_arch` and `target_os`. If those two parameters are the only things determining the value of the bindings, then we can simplify this check in `build.rs`:

```rust
        let supported_platforms = vec![
            "x86_64-unknown-linux-gnu".to_string(),
            "aarch64-unknown-linux-gnu".to_string(),
            "armv7-unknown-linux-gnueabi".to_string(),
            "armv7-unknown-linux-gnueabihf".to_string(),
            "arm-unknown-linux-gnueabi".to_string(),
        ];
        let target = std::env::var("TARGET").unwrap();

        // check if target is in the list of supported ones or panic with nice message
        if !supported_platforms.contains(&target) {
            panic!("Compilation target ({}) is not part of the supported targets ({:?}). Please compile with the \"generate-bindings\" feature or add support for your platform :)", target, supported_platforms);
        }
```

by parsing the target triple as `<arch><sub>-<vendor>-<sys>-<abi>` (see [here](https://doc.rust-lang.org/cargo/appendix/glossary.html#target)) and only checking for the tuple (arch, os).

That would allow much more crates to be used with the generated bindings.

edit: exact same issue on the `tss-esapi` crate.

hug-dev:
Only check the target architecture and OS if they map with pregenerated
bindings. This assumes that only those two parameters influence the
binding generation.

Also add support for x86 and PowerPC64 targets.

Fix #15


hug-dev:
Also tried to parse `aarch64-poky-linux` and `x86_64-poky-linux` and that works fine.

hug-dev:
If it's accepted, I plan to use that in the `tss-esapi` crate as well and then add the new targets in the Parsec tests.

.github/workflows/ci.yml
```diff
@@ -25,7 +25,10 @@ jobs:
           rustup target add armv7-unknown-linux-gnueabi &&
           rustup target add armv7-unknown-linux-gnueabihf &&
           rustup target add arm-unknown-linux-gnueabi &&
-          rustup target add aarch64-unknown-linux-gnu
+          rustup target add aarch64-unknown-linux-gnu &&
+          rustup target add i686-unknown-linux-gnu &&
+          rustup target add powerpc64-unknown-linux-gnu &&
+          rustup target add powerpc64le-unknown-linux-gnu
```
anta5010, line 31: Could you add x86_64-poky-linux and x86_64-poky-linux and aarch64-poky-linux

ci.sh
```diff
@@ -13,6 +13,9 @@ RUST_BACKTRACE=1 cargo build --target arm-unknown-linux-gnueabi
 RUST_BACKTRACE=1 cargo build --target armv7-unknown-linux-gnueabi
 RUST_BACKTRACE=1 cargo build --target armv7-unknown-linux-gnueabihf
 RUST_BACKTRACE=1 cargo build --target aarch64-unknown-linux-gnu
+RUST_BACKTRACE=1 cargo build --target i686-unknown-linux-gnu
+RUST_BACKTRACE=1 cargo build --target powerpc64-unknown-linux-gnu
+RUST_BACKTRACE=1 cargo build --target powerpc64le-unknown-linux-gnu
```
anta5010, line 18: as above

anta5010 (changes requested):

.github/workflows/ci.yml
ionut-arm: I think those two targets exist only in the context of Yocto? Or maybe we need to point `rustup` to a different repo? Because I can't find those targets (or any `poky` targets) through my `rustup`

cryptoki-sys/build.rs
```diff
@@ -8,18 +8,21 @@ fn main() {
...
+        match (target.architecture, target.operating_system) {
+            (Architecture::Arm(_), OperatingSystem::Linux) => {}
+            (Architecture::Aarch64(_), OperatingSystem::Linux) => {}
+            (Architecture::X86_64, OperatingSystem::Linux) => {}
+            (Architecture::X86_32(_), OperatingSystem::Linux) => {}
+            (Architecture::Powerpc64, OperatingSystem::Linux) => {}
+            (Architecture::Powerpc64le, OperatingSystem::Linux) => {}
+            (arch, os) => {
+                panic!("Compilation target (architecture, OS) tuple ({}, {}) is not part of the supported tuples. Please compile with the \"generate-bindings\" feature or add support for your platform :)", arch, os);
+            }
```
ionut-arm, line 25: 👌🏻 👌🏻 Perfect

cryptoki-sys/Cargo.toml
```diff
@@ -13,6 +13,7 @@ documentation = "https://docs.rs/crate/cryptoki"
 
 [build-dependencies]
 bindgen = { version = "0.57.0", optional = true }
+target-lexicon = "0.12.0"
```
ionut-arm, line 16: Isn't this optional?

hug-dev: It is optional when `generate-bindings` is activated but I would have to create a feature that is "not `generate-bindings`" and activate it there. I don't think that's possible currently.

Or I could create a `not-generate-bindings` feature activated by default but then it gets messy. `target_lexicon` does not have any dependencies so I am fine with it.

.github/workflows/ci.yml
hug-dev: Yes that's why I did not add it there but I tried manually and it worked.

anta5010: OK, thanks

anta5010 (approved):

ionut-arm (approved):

PR #22, opened 2021-05-28
Add support for Elliptic Curves signing

wiktor-k:
This change adds support for retrieving `EC_POINT` attribute from cards as well as `CKM_ECDSA` signing mechanism and the `EC` key type.

I'm using this patch for OpenPGP signature generation for my small proof of concept: https://gitlab.com/wiktor/pkcs11-openpgp/#pkcs11-tools-for-openpgp and tested getting EC public keys and signing using Yubico 5C with secp256r1 key stored in a PIV applet and generated through `ykman-gui`.

### DIFF
--- a/cryptoki/src/types/object.rs
+++ b/cryptoki/src/types/object.rs
@@ -27,6 +27,8 @@ pub enum AttributeType {
     Decrypt,
     /// Determines if it is possible to derive other keys from the key
     Derive,
+    /// DER-encoded Elliptic Curve point
+    EcPoint,
     /// Determines if a key supports encryption
     Encrypt,
     /// Determines if a key is extractable and can be wrapped
@@ -80,6 +82,7 @@ impl From<AttributeType> for CK_ATTRIBUTE_TYPE {
             AttributeType::Copyable => CKA_COPYABLE,
             AttributeType::Decrypt => CKA_DECRYPT,
             AttributeType::Derive => CKA_DERIVE,
+            AttributeType::EcPoint => CKA_EC_POINT,
             AttributeType::Encrypt => CKA_ENCRYPT,
             AttributeType::Extractable => CKA_EXTRACTABLE,
             AttributeType::Id => CKA_ID,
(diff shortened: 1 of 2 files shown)


ionut-arm:
Ha, perfect timing 😄  Thanks for the patch, I was in the process of adding ECC support to Parsec, which means adding ECC support to this crate. I'll be raising some PRs next week for generating the keys - and will have a look at this in the next few days.

cryptoki/src/types/object.rs
```diff
@@ -258,6 +265,7 @@ impl Attribute {
             | Attribute::Wrap(_) => 1,
             Attribute::Base(_) => 1,
             Attribute::Class(_) => std::mem::size_of::<CK_OBJECT_CLASS>(),
+            Attribute::EcPoint(bytes) => std::mem::size_of::<u8>() * bytes.len(),
```
hug-dev, line 268: I believe just `bytes.len()` would give you the same result

hug-dev (approved):
Thanks a lot for adding this 👌 It all looks pretty good to me and matching the specs. Just one tiny nitpick you can ignore if you feel too lazy 😬

cryptoki/src/types/object.rs
wiktor-k: Understood!

I see I copied a fragment from `Value` variant but the rest of them use just `bytes.len()`:

>            Attribute::Modulus(bytes) => bytes.len(),
>            Attribute::Value(bytes) => std::mem::size_of::<u8>() * bytes.len(),

wiktor-k: Ok, fixed both of them. Thanks for your review & time @hug-dev !

cryptoki/src/types/object.rs
```diff
@@ -506,6 +518,7 @@ impl TryFrom<CK_KEY_TYPE> for KeyType {
     fn try_from(key_type: CK_KEY_TYPE) -> Result<Self> {
         match key_type {
             CKK_RSA => Ok(KeyType::RSA),
+            CKK_EC => Ok(KeyType::EC),
```
ionut-arm, line 521: Could you also map `CKK_ECDSA` to `KeyType::EC`? The spec seems to treat them similarly, e.g. `EC (also related to ECDSA) private key objects (object class CKO_PRIVATE_KEY, key type CKK_EC or CKK_ECDSA) hold EC private keys`

wiktor-k: Unfortunately I can't do that because... `CKK_ECDSA` is exactly the same value as `CKK_EC`: https://github.com/parallaxsecond/rust-cryptoki/blob/main/cryptoki-sys/src/lib.rs#L133 and fighting the compiler is a futile battle!

ionut-arm: Ah, that makes sense - well, at least in terms of our interface, thanks!

ionut-arm (approved):

PR #23, opened 2021-06-01
Add ECC key generation

ionut-arm:
This commit adds generation of EC keys as a mechanism, along with the
EcParameters attribute required by this mechanism.

### DIFF
--- a/cryptoki/src/types/object.rs
+++ b/cryptoki/src/types/object.rs
@@ -13,6 +13,7 @@ use std::ffi::c_void;
 use std::ops::Deref;
 
 #[derive(Debug, Copy, Clone)]
+#[non_exhaustive]
 /// Type of an attribute
 pub enum AttributeType {
     /// List of mechanisms allowed to be used with the key
@@ -27,6 +28,8 @@ pub enum AttributeType {
     Decrypt,
     /// Determines if it is possible to derive other keys from the key
     Derive,
+    /// Parameters defining an elliptic curve
+    EcParams,
     /// DER-encoded Elliptic Curve point
     EcPoint,
     /// Determines if a key supports encryption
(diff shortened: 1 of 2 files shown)


cryptoki/src/types/object.rs
```diff
@@ -27,6 +28,8 @@ pub enum AttributeType {
     Decrypt,
     /// Determines if it is possible to derive other keys from the key
     Derive,
+    /// Parameters defining an elliptic curve.
```
wiktor-k, line 31: This is extremely nitpicky but other doc comments did not end with the dot (`. `). Maybe it'd be worth to keep it consistent one way or another?

ionut-arm: Sure, will do, I prefer following these conventions - in this case I just wasn't paying attention 😅

cryptoki/src/types/object.rs
```diff
@@ -27,6 +28,8 @@ pub enum AttributeType {
     Decrypt,
     /// Determines if it is possible to derive other keys from the key
     Derive,
+    /// Parameters defining an elliptic curve
+    EcParameters,
```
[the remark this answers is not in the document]
ionut-arm: Yeah, sure - also going to change `EcKeyPairGen` to `EccKeyPairGen`

cryptoki/Cargo.toml
```diff
@@ -16,7 +16,7 @@ libloading = "0.7.0"
 log = "0.4.14"
 derivative = "2.2.0"
 secrecy = "0.7.0"
-psa-crypto = { git = "https://github.com/parallaxsecond/rust-psa-crypto.git", rev = "8605006d34944fa880edd3d4d347f460c5585747", default-features = false, optional = true }
+psa-crypto = { git = "https://github.com/parallaxsecond/rust-psa-crypto.git", rev = "15ba7cd048923ae7193b268df51c004e8ceff48e", default-features = false, optional = true }
```
hug-dev, line 19: Do you need to use that commit? That might be a problem when you are trying to compile Parsec

ionut-arm: Oops, that's what I needed locally, but didn't pull down your recent changes in Parsec.

hug-dev (approved):
🔑

PR #26, opened 2021-06-17
Add Object ID attribute

wiktor-k:
Object ID is commonly used to identify well-known objects. For example
the Object ID `2.16.840.1.101.3.7.2.1.0` is used to identify the X.509
Certificate for Digital Signature.


hug-dev (approved):
Thanks!

ionut-arm (approved):

PR #27, opened 2021-06-17
Add get_token_info

wiktor-k:
Okay folks, I'll need some more help on this one.

* I'm not sure where to place the `TokenInfo` struct (should it be in the `-sys` crate?)
* about the struct members themselves: for example currently they are raw types that are read from the PKCS#11 driver but maybe it would be nice to clean them up (e.g. all strings - as usual in PKCS#11 - are space padded - see the unit test).
* `get_token_info` uses some raw ffi code I found that works but I'm not sure if that's the best approach.

Please ignore placeholder docs for a moment. I'll add them when other issues are fixed :)

### DIFF
--- a/cryptoki/src/types/slot_token.rs
+++ b/cryptoki/src/types/slot_token.rs
@@ -7,8 +7,9 @@
 //! Slot and token types
 
 use crate::{Error, Result};
-use cryptoki_sys::CK_SLOT_ID;
+use cryptoki_sys::{CK_SLOT_ID, CK_TOKEN_INFO};
 use std::convert::{TryFrom, TryInto};
+use std::ops::Deref;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 /// Type identifying a slot
@@ -40,5 +41,45 @@ impl TryFrom<u64> for Slot {
 impl From<Slot> for CK_SLOT_ID {
     fn from(slot: Slot) -> Self {
         slot.slot_id
+    }
+}
+
+/// Contains information about a token
+#[derive(Debug, Clone, Copy, Default)]
+pub struct TokenInfo {
+    val: CK_TOKEN_INFO,
+}
+
+impl TokenInfo {
+    pub(crate) fn new(val: CK_TOKEN_INFO) -> Self {
+        Self { val }
+    }
+
+    /// Returns the ID of the device manufacturer
+    pub fn get_manufacturer_id(&self) -> String {
+        String::from_utf8_lossy(&self.val.manufacturerID)
+            .trim_end()
+            .to_string()
+    }
+
+    /// Returns the character-string serial number of the device
+    pub fn get_serial_number(&self) -> String {
+        String::from_utf8_lossy(&self.val.serialNumber)
+            .trim_end()
+            .to_string()
+    }
+}
+
+impl Deref for TokenInfo {
+    type Target = CK_TOKEN_INFO;
+
+    fn deref(&self) -> &Self::Target {
+        &self.val
+    }
+}
+
+impl From<TokenInfo> for CK_TOKEN_INFO {
+    fn from(token_info: TokenInfo) -> Self {
+        *token_info
     }
 }
--- a/cryptoki/src/functions/slot_token_management.rs
+++ b/cryptoki/src/functions/slot_token_management.rs
@@ -4,10 +4,11 @@
 
 use crate::get_pkcs11;
 use crate::types::function::Rv;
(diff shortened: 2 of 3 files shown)


cryptoki/src/types/slot_token.rs
```diff
@@ -42,3 +42,45 @@ impl From<Slot> for CK_SLOT_ID {
         slot.slot_id
     }
 }
+
+///
+#[derive(Debug, Clone, Copy, Default)]
+#[repr(C)]
```
ionut-arm, line 48: I'm not sure we need this here. I'm thinking we could make `TokenInfo` a proper abstracted version of the FFI type, and simply have conversions between them. This would allow us to, for example, present `label`, `manufacturer_id` and `model` as `String` directly.

cryptoki/src/types/slot_token.rs
```diff
@@ -42,3 +42,45 @@ impl From<Slot> for CK_SLOT_ID {
...
+    ///
+    pub min_pin_len: u32,
+    ///
+    pub total_public_memory: u32,
+    ///
+    pub free_public_memory: u32,
+    ///
+    pub total_private_memory: u32,
+    ///
+    pub free_private_memory: u32,
```
ionut-arm, line 79: I think these should be using [`Ulong`](https://docs.rs/cryptoki/0.1.1/cryptoki/types/struct.Ulong.html)?

ionut-arm:
> I'm not sure where to place the TokenInfo struct (should it be in the -sys crate?)

I think the place where you put them is the right one. That file/module is supposed to cover section `3.2 Slot and token types` of the spec

cryptoki/src/types/slot_token.rs
hug-dev: Agree with that, when we use `repr(C)` or `repr(transparent)` in `cryptoki` is when the Rust abstracted structures must match the memory layout of their C variants. The main reason for that is that then we don't need to do any move of the data owned by the application. I can expand on that if you want, took us a while to find out through a lot of different trials and errors...
In the case of `CK_TOKEN_INFO`, it is only used for `C_GetTokenInfo` and does not need to contain any initial data from the application. That means that we can call `C_GetTokenInfo` with an empty `CK_TOKEN_INFO` and then convert that to an abstracted, Rust native, `TokenInfo` which does not need to have any specific memory layout.

There are multiple ways of how to design `TokenInfo`:
- for each field have an abstracted Rust version (`CK_UTF8CHAR`/`CK_CHAR` -> `String`, `CK_FLAGS` -> `Flags`, `CK_ULONG` -> `Ulong` but could also probably be `u32`, etc...). All the fields are put `pub`. That way we have to convert everything when `TokenInfo` is created.
- have `TokenInfo` contain a `CK_TOKEN_INFO` as a private field and have getters that perform the conversion dynamically.

I think the second option is better as it easy to add and we can only add the getters when we need them.

wiktor-k: Okay, done that for manufacturer_id and serial_number (that I need). If someone needs more they can add appropriate functions.

cryptoki/src/types/slot_token.rs
wiktor-k: Replaced all that with inner `CK_TOKEN_INFO`. Thanks for the idea!

wiktor-k:
Okay, I've vastly simplified the code when it occurred to me that `CK_TOKEN_INFO` generated by bindgen has already all members that I needed (btw is it possible somehow to see the code generated by bindgen? IDE doesn't see it just goes to macros :( ).

I've converted `TokenInfo` to be a wrapper around `CK_TOKEN_INFO` with accessor methods for properties that I use.

I've also added Deref and From implementations to be somehow similar to what you use with other types. (This can be used to access properties that don't have convenience accessor functions).

ionut-arm:
> btw is it possible somehow to see the code generated by bindgen? IDE doesn't see it just goes to macros :( 

We already have bindings commited for a number of architectures, you can see them [here](https://github.com/parallaxsecond/rust-cryptoki/tree/main/cryptoki-sys/src/bindings) - this is to have faster/easier builds on those platforms. Otherwise, you should be able to find them in something like `./target/debug/build/cryptoki-sys-.../out/cryptoki-sys-bindings.rs`, if I remember correctly.

wiktor-k:
> Is this the code you meant?

Yes! The builds for architectures are great for looking for what's available, thanks @ionut-arm! I didn't spot this previously.

hug-dev (approved):
Looks good 👌

ionut-arm (approved):

PR #33, opened 2021-07-19
EC Edward and Montgomery support

daxpedda:
Adds support for Ed25519 or X25519 for example.


ionut-arm (approved):

cryptoki-sys/pkcs11.h
```diff
@@ -884,6 +884,7 @@ typedef unsigned long ck_mechanism_type_t;
 
 /* From version 3.0 */
 #define CKM_EC_EDWARDS_KEY_PAIR_GEN	(0x1055UL)
+#define CKM_EC_MONTGOMERY_KEY_PAIR_GEN  (0x1056UL)
```
hug-dev, line 887: This is coming from version 3.0 right? Is that fine you think to mix contants from both version? Looks like it was already the case 😅 

I am wondering if generally we should add support to the 3.0 standard in this crate through a feature flag.

On another related note: [this repo](https://github.com/oasis-tcs/pkcs11) would be a better source the PKCS11 header file.

hug-dev (approved):
Thanks!

cryptoki-sys/pkcs11.h
daxpedda: > This is coming from version 3.0 right?

Yeah: https://github.com/oasis-tcs/pkcs11/blob/0c3ecef6c90dcd8249ba64cc99d50d9f8f73124e/working/3-00-current/pkcs11t.h#L1124.

> I am wondering if generally we should add support to the 3.0 standard in this crate through a feature flag.

I have no clue how version detection works in PKCS#11 or what it even means, I assumed until now that if something isn't supported it will just give a not-supported-error. In any case, if PKCS#11 support version detection, I don't think a crate feature makes a lot of sense, it would need something run-time.

hug-dev: Something run-time would be good! We need to check if the two versions do not bring too many breaking changes.

PR #35, opened 2021-07-20
Added new methods to fix issue 375 - get slots with initialized token…

Sven-bg:
…, for PKCS 11 Provider

Signed-off-by: Adriano Pallavicino <adriano.pallavicino@arm.com>

### DIFF
--- a/cryptoki/src/functions/slot_token_management.rs
+++ b/cryptoki/src/functions/slot_token_management.rs
@@ -47,6 +47,25 @@ impl Pkcs11 {
         Ok(slots)
     }
 
+    /// Get all slots available with a token
+    pub fn get_slots_with_initialized_token(&self) -> Result<Vec<Slot>> {
+        let slots = self.get_slots_with_token()?;
+
+        slots
+            .into_iter()
+            .filter_map(|slot| match self.get_token_info(slot) {
+                Ok(token_info) => {
+                    if token_info.get_flags().token_initialized() {
+                        Some(Ok(slot))
+                    } else {
+                        None
+                    }
+                }
+                Err(e) => Some(Err(e)),
+            })
+            .collect()
+    }
+
     /// Get all slots
     pub fn get_all_slots(&self) -> Result<Vec<Slot>> {
         let mut slot_count = 0;
--- a/cryptoki/src/types/slot_token.rs
+++ b/cryptoki/src/types/slot_token.rs
@@ -6,6 +6,7 @@
 
 //! Slot and token types
 
+use crate::types::Flags;
 use crate::{Error, Result};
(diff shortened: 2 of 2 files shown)


cryptoki/src/functions/slot_token_management.rs
```diff
@@ -1,5 +1,3 @@
-// Copyright 2021 Contributors to the Parsec project.
-// SPDX-License-Identifier: Apache-2.0
```
hug-dev, line 2: It's ok to keep those lines :)

cryptoki/src/functions/slot_token_management.rs
```diff
@@ -47,6 +45,47 @@ impl Pkcs11 {
...
+                slots.as_mut_ptr(),
+                &mut slot_count,
+            ))
+            .into_result()?;
+        }
+
+        let mut slots: Vec<Slot> = slots.into_iter().map(Slot::new).collect();
+
+        // This should always truncate slots.
+        slots.resize(slot_count.try_into()?, Slot::new(0));
```
hug-dev, line 75: I think you can replace all those lines with:
```rust
let slots = self.get_slots_with_token()?;
```
It should do the same using the method above!

hug-dev (commented):
Thanks for the PR! Made a few comments 👌

hug-dev:
You might need to do a `cargo fmt` before re-submitting as well!

hug-dev (approved):
Thank you!!

cryptoki/src/functions/slot_token_management.rs
```diff
@@ -47,6 +47,27 @@ impl Pkcs11 {
...
+                Ok(token_info) => {
+                    if token_info.get_flags().token_initialized() {
+                        Some(Ok(slot))
+                    } else {
+                        None
+                    }
+                }
+                Err(e) => Some(Err(e)),
+            })
+            .collect::<Result<Vec<Slot>>>()?;
```
wiktor-k, line 66: This is minor but you can either omit type declaration in `let slots: Vec<Slot>` *or* shorten the explicit type in collect to `collect::<Result<_>>()?;`

hug-dev: Since this function returns a `Result<Vec<Slot>>, I am wondering if you could also remove everything:
- the `let slots: Vec<Slot> =` at the beginning
- just have `.collect()` at the end (without `?` and `;`)

wiktor-k: Indeed. That'd make it more readable too IMHO.

hug-dev (approved):
Thank you!

PR #41, opened 2021-09-01
Multiple updates to provide additional 'info' function bindings betwe…

mjb3279:
…en rust-cryptoki and PKCS#11 v2.40

- Removed `Flags` and replaced with specific flag structures (`SessionFlags`, `SlotFlags`, `TokenFlags`, `InitialzeFlags` and `MechanismFlags`)
- Added support for `C_GetInfo` using `Info` as the structure and `get_library_info` as the function
- Added support for `C_GetSlotInfo` using `SlotInfo` as the structure and `get_slot_info` as the function
- Added support for `C_GetSessionInfo` using `SessionInfo` as the structure and `get_session_info` as the function
- Expanded `TokenInfo` to return all information from `CK_TOKEN_INFO` structure
- A number of stringify like functions
- Added tests for added functionality

### DIFF
--- a/cryptoki/src/lib.rs
+++ b/cryptoki/src/lib.rs
@@ -44,6 +44,7 @@ pub mod types;
 use crate::types::function::{Rv, RvError};
 use crate::types::session::{Session, UserType};
 use crate::types::slot_token::Slot;
+use cryptoki_sys::CK_UTF8CHAR;
 use derivative::Derivative;
 use log::error;
 use secrecy::{ExposeSecret, Secret, SecretVec};
@@ -292,3 +293,26 @@ impl Drop for Pkcs11 {
 
 /// Main Result type
 pub type Result<T> = core::result::Result<T, Error>;
+
+fn str_from_blank_padded(field: &[CK_UTF8CHAR]) -> String {
+    let decoded_str = String::from_utf8_lossy(field);
+    decoded_str.trim_end_matches(' ').to_string()
+}
+
+fn label_from_str(label: &str) -> [CK_UTF8CHAR; 32] {
+    let mut lab: [CK_UTF8CHAR; 32] = [32; 32];
+    let mut i = 0;
+    for c in label.chars() {
+        if i + c.len_utf8() <= 32 {
+            let mut buf = [0; 4];
+            let bytes = c.encode_utf8(&mut buf).as_bytes();
+            for b in bytes {
+                lab[i] = *b;
+                i += 1;
+            }
+        } else {
+            break;
+        }
+    }
+    lab
+}
--- a/cryptoki/src/lib.rs
+++ b/cryptoki/src/lib.rs
@@ -294,7 +294,7 @@ impl Drop for Pkcs11 {
 /// Main Result type
 pub type Result<T> = core::result::Result<T, Error>;
 
-fn str_from_blank_padded(field: &[CK_UTF8CHAR]) -> String {
+fn string_from_blank_padded(field: &[CK_UTF8CHAR]) -> String {
     let decoded_str = String::from_utf8_lossy(field);
     decoded_str.trim_end_matches(' ').to_string()
 }
--- a/cryptoki/src/types/mod.rs
+++ b/cryptoki/src/types/mod.rs
@@ -121,9 +121,15 @@ impl From<InitializeFlags> for CK_FLAGS {
     }
 }
 
-impl From<CK_FLAGS> for InitializeFlags {
-    fn from(flags: CK_FLAGS) -> Self {
-        Self { flags }
+impl TryFrom<CK_FLAGS> for InitializeFlags {
+    type Error = Error;
+
+    fn try_from(flags: CK_FLAGS) -> Result<Self> {
+        if flags & !(CKF_OS_LOCKING_OK | CKF_LIBRARY_CANT_CREATE_OS_THREADS) != 0 {
+            Err(Error::InvalidValue)
+        } else {
+            Ok(Self { flags })
+        }
     }
 }
 
--- a/cryptoki/src/types/mod.rs
+++ b/cryptoki/src/types/mod.rs
@@ -9,30 +9,58 @@ pub mod object;
 pub mod session;
 pub mod slot_token;
 
-use crate::{Error, Result};
+use crate::{str_from_blank_padded, Error, Result};
 use cryptoki_sys::*;
 use log::error;
 use std::convert::TryFrom;
 use std::convert::TryInto;
+use std::fmt::Formatter;
 use std::ops::Deref;
 
+trait Flags: std::fmt::Display {
+    type FlagType: Copy;
+
+    fn get_flag_value(&self) -> Self::FlagType;
+
+    fn get_flag(&self, flag: Self::FlagType) -> bool;
+
+    fn set_flag(&mut self, flag: Self::FlagType, b: bool);
+
+    fn stringify_flag(flag: Self::FlagType) -> &'static str;
+
+    fn stringify_fmt(&self, f: &mut Formatter<'_>, flags: Vec<Self::FlagType>) -> std::fmt::Result {
+        let mut first_done = false;
+        for flag in flags.iter() {
+            if self.get_flag(*flag) {
+                if first_done {
+                    write!(f, ", ")?;
+                }
+                write!(f, "{}", Self::stringify_flag(*flag))?;
+                first_done = true;
+            }
+        }
+        Ok(())
+    }
+}
+
 #[derive(Debug, Default, Clone, Copy)]
-/// Collection of boolean flags
-pub struct Flags {
+/// Collection of flags defined for [`CK_SLOT_INFO`]
+pub struct InitializeFlags {
     flags: CK_FLAGS,
 }
 
-impl Flags {
-    /// Create a new instance with all flags set to false
-    pub fn new() -> Self {
-        Flags::default()
+impl Flags for InitializeFlags {
+    type FlagType = CK_FLAGS;
+
+    fn get_flag_value(&self) -> Self::FlagType {
+        self.flags
     }
 
-    fn get_flag(&self, flag: CK_FLAGS) -> bool {
-        self.flags & flag == flag
+    fn get_flag(&self, flag: Self::FlagType) -> bool {
+        self.get_flag_value() & flag == flag
     }
 
-    fn set_flag(&mut self, flag: CK_FLAGS, b: bool) {
+    fn set_flag(&mut self, flag: Self::FlagType, b: bool) {
         if b {
             self.flags |= flag;
         } else {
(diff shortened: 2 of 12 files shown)


ionut-arm:
Oh wow, this is quite a big PR - I'll review it over the next few days, potentially in a few sessions (i.e. adding the comments in batches). I'm not that fluent in PKCS11 v2.40 but this might be a good opportunity to learn more 😄

mjb3279:
Completely understand, yes it's a big one.

Maybe one of my more 'controversial' updates was replacing `Flags` with specific flag structures (which, by the way, will break main if merged in with the other outstanding PRs).

In terms of PKCS#11, I'm more than happy to answer any question you have.  I've implemented a fully PKCS#11 v2.40 library so I'm (unfortunately) very versed in the spec.

ionut-arm:
> Maybe one of my more 'controversial' updates was replacing Flags with specific flag structures

More than happy with that sort of change - if we can get the compiler to do the work of enforcing what can be used where, why not.

> In terms of PKCS#11, I'm more than happy to answer any question you have. I've implemented a fully PKCS#11 v2.40 library so I'm (unfortunately) very versed in the spec.

Oh X_X Well I hope there won't be too many questions, but it's good to have the perspective of someone with so much experience with the spec!

cryptoki/src/types/mod.rs
```diff
@@ -9,606 +9,881 @@ pub mod object;
...
-use crate::{Error, Result};
+use crate::{str_from_blank_padded, Error, Result};
 use cryptoki_sys::*;
 use log::error;
 use std::convert::TryFrom;
 use std::convert::TryInto;
+use std::fmt::Formatter;
 use std::ops::Deref;
 
+trait Flags: std::fmt::Display {
```
ionut-arm, line 20: I think instead of a trait it could've been done with a macro instead to remove some of the code duplication, but it's alright now that it's already written out.

cryptoki/src/types/mod.rs
```diff
@@ -9,606 +9,881 @@ pub mod object;
...
+}
 
-    /// Get the RNG flag
-    pub fn rng(&self) -> bool {
-        self.get_flag(CKF_RNG)
+impl From<CK_FLAGS> for InitializeFlags {
+    fn from(flags: CK_FLAGS) -> Self {
+        Self { flags }
     }
+}
```
ionut-arm, line 128: Could this be a `TryFrom` that checks that only the flags that are used by this type are set?

cryptoki/src/types/mod.rs
```diff
@@ -9,606 +9,881 @@ pub mod object;
...
+    pub fn user_pin_final_try(&self) -> bool {
+        self.get_flag(CKF_USER_PIN_FINAL_TRY)
     }
 
-    /// Set the EC_F_P flag
-    pub fn set_ec_f_p(&mut self, b: bool) -> &mut Self {
-        self.set_flag(CKF_EC_F_P, b);
+    /// Sets value of [`CKF_USER_PIN_COUNT_LOW`]
+    pub fn set_user_pin_final_try(&mut self, b: bool) -> &mut Self {
+        self.set_flag(CKF_USER_PIN_COUNT_LOW, b);
```
ionut-arm, line 783:
```suggestion
    /// Sets value of [`CKF_USER_PIN_FINAL_TRY`]
    pub fn set_user_pin_final_try(&mut self, b: bool) -> &mut Self {
        self.set_flag(CKF_USER_PIN_FINAL_TRY, b);
```

cryptoki/src/types/mod.rs
```diff
@@ -9,606 +9,881 @@ pub mod object;
...
+    pub fn so_pin_final_try(&self) -> bool {
+        self.get_flag(CKF_SO_PIN_FINAL_TRY)
     }
 
-    /// Set the DONT_BLOCK flag
-    pub fn set_dont_block(&mut self, b: bool) -> &mut Self {
-        self.set_flag(CKF_DONT_BLOCK, b);
+    /// Sets value of [`CKF_SO_PIN_COUNT_LOW`]
+    pub fn set_so_pin_final_try(&mut self, b: bool) -> &mut Self {
+        self.set_flag(CKF_SO_PIN_COUNT_LOW, b);
```
ionut-arm, line 827:
```suggestion
    /// Sets value of [`CKF_SO_PIN_FINAL_TRY`]
    pub fn set_so_pin_final_try(&mut self, b: bool) -> &mut Self {
        self.set_flag(CKF_SO_PIN_FINAL_TRY, b);
```

cryptoki/src/types/mod.rs
```diff
@@ -699,3 +974,103 @@ impl TryFrom<usize> for Ulong {
...
+    val: CK_INFO,
+}
+
+impl Info {
+    pub(crate) fn new(val: CK_INFO) -> Self {
+        Self { val }
+    }
+
+    /// Returns the version of Cryptoki that the library is compatible with
+    pub fn get_cryptoki_version(&self) -> Version {
```
ionut-arm, line 1039: Getters don't usually have the `get_` in front

ionut-arm (commented):
Thanks again for the patch, I've managed to make my way through the whole PR!

cryptoki/src/lib.rs
```diff
@@ -292,3 +293,26 @@ impl Drop for Pkcs11 {
...
+            for b in bytes {
+                lab[i] = *b;
+                i += 1;
+            }
+        } else {
+            break;
+        }
+    }
+    lab
+}
```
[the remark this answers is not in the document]
mjb3279: I could if you want me to.  I admit that this was code taken from https://github.com/mheese/rust-pkcs11 that would space append a label that's <32 characters long.

The API spec for `C_InitToken` requires the pointer `pLabel` to point to a 32-byte, space-padded, buffer.  However, this just sets the _minimum_.  You can, technically, pass in a label that's 127 characters long, but normally the API implementer will  just take the first 32-bytes of the buffer and use that as the label, so the spec somewhat _implies_ that anything over 32 characters would be truncated anyways.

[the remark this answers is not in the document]
mjb3279: I guess that's the long way of saying, I don't think we need to return a `Result` as the spec will truncate anything >32 characters.

cryptoki/src/types/mod.rs
mjb3279: I can look into that.

cryptoki/src/types/mod.rs
mjb3279: Sure, I can do that.  It's _technically_ not disallowed (sorry double-negative) to pass in values outside of the range of the spec (as in, the spec doesn't directly say that if a value that isn't `CKF_LIBRARY_CANT_CREATE_OS_THREADS` or `CKF_OS_LOCKING_OK` to return `CKR_BAD_ARGUMENTS`), but it's definitely possible some implementer may error out.

There are some exception, e.g. the `flags` field of `CK_INFO` must be 0, or the argument to `C_Finalize` must be a `NULL_PTR`.

I can put in a `TryFrom`.

cryptoki/src/types/mod.rs
mjb3279: Ah, that's probably my `C++` based brain :)

cryptoki/src/types/session.rs
```diff
@@ -83,3 +105,105 @@ impl From<UserType> for CK_USER_TYPE {
...
+    pub const RW_USER_FUNCTIONS: SessionState = SessionState {
+        val: CKS_RW_USER_FUNCTIONS,
+    };
+
+    /// Read/write SO access
+    pub const RW_SO_FUNCTIONS: SessionState = SessionState {
+        val: CKS_RW_SO_FUNCTIONS,
+    };
+
+    fn state_to_str(&self) -> &'static str {
```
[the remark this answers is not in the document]
mjb3279: That's a fair point.  Maybe it would make more sense to make this a static method that takes in a `CK_STATE` value and make it `pub` or `pub (struct)` so it's more generalized.

cryptoki/src/types/mod.rs
```diff
@@ -699,3 +974,103 @@ impl TryFrom<usize> for Ulong {
...
+
+impl std::fmt::Display for Version {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}.{}", self.major, self.minor)
+    }
+}
+
+impl Version {
+    /// Returns the major version
+    pub fn get_major(&self) -> CK_BYTE {
```
MattDavis00, line 999:
```suggestion
    pub fn major(&self) -> CK_BYTE {
```

cryptoki/src/types/mod.rs
```diff
@@ -699,3 +974,103 @@ impl TryFrom<usize> for Ulong {
...
+        self.major
+    }
+
+    /// Returns the minor version
+    pub fn get_minor(&self) -> CK_BYTE {
```
MattDavis00, line 1004:
```suggestion
    pub fn minor(&self) -> CK_BYTE {
```

MattDavis00: (Just an example, similar cases throughout).

MattDavis00 (commented):
Thanks for the changes! Just a few comments 👨🏼‍💻
(Main thing is about `get_` prefixes as Ionut mentioned)

cryptoki/src/types/session.rs
```diff
@@ -83,3 +105,105 @@ impl From<UserType> for CK_USER_TYPE {
...
+    }
+
+    /// Returns the state of the session
+    pub fn get_session_state(&self) -> SessionState {
+        self.val.state.into()
+    }
+
+    /// Returns the slot the session is on
+    pub fn get_slot_id(&self) -> Slot {
+        self.val.slotID.try_into().unwrap()
```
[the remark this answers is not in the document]
mjb3279: Yeah the value of what a `Slot` is has given me a lot of heartburn.  This is thanks to PKCS#11's rather... annoying definition of `CK_ULONG` combined with Windows, by default, setting a `long` to 32-bits, regardless of CPU architecture.

So I think this is fine because `CK_SLOT_ID` is either a `u32` or `u64`, depending on the compiler, and, if I'm reading the code for `TryFrom`, `TryInto`, `Into`, and `From` correctly, which I may not, I think in the end it should just be assigning `CK_SLOT_ID` (which is either a `u32` or `u64`) into the same type.

cryptoki/src/types/mod.rs
mjb3279: Would you be opposed to me looking into this in a separate ticket/issue?  I'm definitely not well versed in writing Rust macros and it may take some time for me to do that.

cryptoki/src/types/mod.rs
mjb3279: I'm not opposed to doing this for the `InitializeFlags`, as it is up to the user to set them, but I don't think we should do this for other flag values that a cryptoki token may return.  The reason is that there may exist tokens that provide vendor specific flags (which is outside the PKCS#11 spec, vendors should only be defining attributes, mechanisms, key types, return types, etc., but vendors will do what vendors do) and then performing a `TryFrom` on a return may cause the function to error out, even if the P11 call succeeds.

Thoughts?

mjb3279:
Ok I think I got most of the changes.  Thank you to both @ionut-arm  and @MattDavis00 for reviewing this fairly large patch.

cryptoki/src/lib.rs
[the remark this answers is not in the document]
ionut-arm: Alright, that works for me!

cryptoki/src/types/mod.rs
ionut-arm: Sure, and it's not really that critical, don't even know if one or the other is actually idiomatic, just saw there's a bit of duplication, but I'm perfectly happy to just keep it as is.

cryptoki/src/types/mod.rs
ionut-arm: Ah, yes, that sounds sensible. Actually, are there "bit-zones" within the integer that are vendor-controlled (as opposed to strictly controlled by the spec), and which we could ignore in the `TryFrom`?

mjb3279: For thinks like `CKR_` and `CKA_` and `CKK_` and `CKM_`  and `CKO_` types, yes anything at `0x8000_0000` and above are 'vendor defined'.  Technically that doesn't exist for flags but vendors "should" be following that convention.  Maybe we have an 'VendorDefined' field in the different enums that contains values that are "unexpected".

cryptoki/src/types/session.rs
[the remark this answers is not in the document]
ionut-arm: Hmm, yes, I think that's true! Could you put a comment in there mentioning this? We try to make sure than any calls to `unwrap` or `expect` have such guarantees and are commented properly

cryptoki/src/types/mod.rs
ionut-arm: Makes sense. So with the `TryFrom` updated for `InitializeFlags` we're set. I'm guessing these extra, vendor-specific flags can be accessed by just dereferencing the integer value within?

mjb3279: Yes.  I'm thinking of at some point added a new variant in the above mentioned `enums` to provide Vendor specific definitions

mjb3279:
Ok I think I got them all now.

cryptoki/src/lib.rs
```diff
@@ -292,3 +293,26 @@ impl Drop for Pkcs11 {
 
 /// Main Result type
 pub type Result<T> = core::result::Result<T, Error>;
+
+fn str_from_blank_padded(field: &[CK_UTF8CHAR]) -> String {
```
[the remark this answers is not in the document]
ionut-arm: Is this resolved?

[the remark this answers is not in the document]
mjb3279: Yes.  This is an old diff.  Now it is called "string_from_blank_padded"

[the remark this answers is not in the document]
ionut-arm: Ah, yes, thanks, I didn't realise there are multiple commits and that I was looking at a particular one.

PR #43, opened 2021-09-08
Expanded support for remaining attributes, mechanism types (but not t…

mjb3279:
…he mechanisms themselves yet), key types, etc

- Added `AES_KEY_GEN`
- Added mechanism stringify function
- Added support for `CK_DATE`
- Added many many many more attributes (probably not all are in there yet but a lot were added)
- Added support for all PKCS#11 v2.40 object classes
- Added support for all PKCS#11 v2.40 key types
- Added support for Certificates
- Added PartialEq and Eq to ObjectHandle

### DIFF
--- a/cryptoki/src/types/mod.rs
+++ b/cryptoki/src/types/mod.rs
@@ -896,7 +896,7 @@ impl From<CK_FLAGS> for TokenFlags {
 }
 
 #[repr(u8)]
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
 /// Byte-sized boolean
 pub enum Bbool {
     /// False value
@@ -929,6 +929,15 @@ impl From<bool> for Bbool {
     }
 }
 
+impl From<Bbool> for bool {
+    fn from(val: Bbool) -> Self {
+        match val {
+            Bbool::False => false,
+            Bbool::True => true,
+        }
+    }
+}
+
 impl TryFrom<CK_BBOOL> for Bbool {
     type Error = Error;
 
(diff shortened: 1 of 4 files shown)


mjb3279:
Again, apologizes for yet another large PR, however this one is, I think, fairly straight forward.

ionut-arm:
Apologies, it's been a busy week, but I'll most certainly review your two PRs ASAP. That's likely to be next week, though.

mjb3279:
No need to apologize!  Appreciate all the time you guys set aside doing these entirely way too big PRs!

cryptoki/src/types/object.rs
```diff
@@ -430,11 +858,49 @@ impl TryFrom<CK_ATTRIBUTE> for Attribute {
                     .collect::<Result<Vec<MechanismType>>>()?;
                 Ok(Attribute::AllowedMechanisms(types))
             }
+            AttributeType::EndDate => {
+                let date = val.as_ptr() as *const CK_DATE;
+                unsafe {
+                    let year = String::from_utf8_lossy(Vec::from((*date).year).as_slice())
+                        .trim_end()
+                        .to_string();
```
ionut-arm, line 866: Since this is in a `TryFrom`, we should probably return an error if this conversion fails instead of just returning a "buggy" date. I'm also wondering if in the `new_from_str_slice` we should do something similar where we at least check that every character is a digit.

ionut-arm (commented):
Looks good overall, thanks for the massive patch! Got a couple of comments below

cryptoki/src/types/mod.rs
```diff
@@ -944,6 +953,89 @@ impl TryFrom<CK_BBOOL> for Bbool {
...
+            .trim_end()
+            .to_string();
+        let month = String::from_utf8_lossy(Vec::from(self.month).as_slice())
+            .trim_end()
+            .to_string();
+        let day = String::from_utf8_lossy(Vec::from(self.day).as_slice())
+            .trim_end()
+            .to_string();
+        // Prints this month/day/year
+        // todo - Figure out how to determine if US vs European since us Americans do it the 'right' way ;)
```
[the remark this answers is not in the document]
mjb3279: Sure, something like 13 Sept 2021?

cryptoki/src/types/object.rs
mjb3279: I've been going back and forth on how to answer this.  I think it comes down to how much 'checking' a PKCS11 library does vs the token and what the user _intends_ the date to mean.  Frankly, the `CK_DATE` type is more annoying than it's worth.  Its only use is for "reference only" and the token can safely ignore the values within it.

I think from a "helping the user" perspective, it would be helpful to bounds check the date _when the user provides it_.  The issue I think comes down to the fact that the library does not have any control over the module not providing back a valid date for an object (e.g. someone creates a key with a `CKA_END_DATE` of 99/99/9999 or maybe NO/EX/PIRE to indicate that the key never expires.  Both values are technically valid).  That may not be a _valid_ date, but it is a valid 'value' that can be held by `CK_DATE`.  The value of a `CK_DATE` is up to the application to set and interpret.

The spec states that you should only use the characters in "Table 3" "Table 3" for any `CK_CHAR` type.  `CK_UTF8CHAR` is governed by RFC2279.  Meaning as long as the values of a `CK_CHAR` field are within the bounds stated by Table 3, it's valid.

Thoughts?

cryptoki/src/types/mod.rs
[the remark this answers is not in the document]
mjb3279: Also see my comment below as to why `CK_DATE` is a very obnoxious type.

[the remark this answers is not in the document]
ionut-arm: Yeah, something like that works!

cryptoki/src/types/object.rs
ionut-arm: In that case let's just leave it as is, it doesn't seem like something critical anyway. We can implement a method on `Date` to check whether it's valid or not, if someone requests it.

cryptoki/src/types/mod.rs
```diff
@@ -1030,9 +1030,8 @@ impl std::fmt::Display for Date {
         let day = String::from_utf8_lossy(Vec::from(self.day).as_slice())
             .trim_end()
             .to_string();
-        // Prints this month/day/year
-        // todo - Figure out how to determine if US vs European since us Americans do it the 'right' way ;)
-        write!(f, "{}/{}/{}", month, day, year)
+
+        write!(f, "Month: {}\nDay: {}\nYear: {}", month, day, year)
```
mjb3279, line 1034: Per the discussion, I'm going to print out "Month: Day: Year" rather than trying to stringify month as the data coming back from the module may not nicely fit into a date structure.

mjb3279:
Hi @ionut-arm  is there anything else you need me to change on this PR?

ionut-arm (approved):
Sorry for the delay, looks good now!

PR #44, opened 2021-09-08
Added functions to utilize the "Random" PKCS#11 functions

mjb3279:
- Added `generate_random_slice` which will put random data into a provided slice
- Added `generate_random_vec` which will return random data in a `Vec<u8>`
- Added `seed_random` which will seed the RNG on the token

### DIFF
--- a/cryptoki/src/functions/random.rs
+++ b/cryptoki/src/functions/random.rs
@@ -10,6 +10,11 @@ use std::convert::TryInto;
 
 impl<'a> Session<'a> {
     /// Generates a random number and sticks it in a slice
+    ///
+    /// # Arguments
+    ///
+    /// * `random_slice` - The slice to stick the random data into.  The length of the slice represents
+    /// the number of bytes to obtain from the RBG
     pub fn generate_random_slice(&self, random_data: &mut [u8]) -> Result<()> {
         unsafe {
             Rv::from(get_pkcs11!(self.client(), C_GenerateRandom)(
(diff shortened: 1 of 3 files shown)


cryptoki/src/functions/random.rs
```diff
@@ -0,0 +1,51 @@
...
+//! Functions used to generate random numbers
+
+use crate::get_pkcs11;
+use crate::types::function::Rv;
+use crate::types::session::Session;
+use crate::Result;
+use std::convert::TryInto;
+
+impl<'a> Session<'a> {
+    /// Generates a random number and sticks it in a slice
```
ionut-arm, line 12: Could you also mention here that the length of the slice is used to decide how many bytes are returned?

cryptoki/src/functions/random.rs
```diff
@@ -0,0 +1,51 @@
...
+                random_data.as_ptr() as *mut u8,
+                random_data.len().try_into()?,
+            ))
+            .into_result()?;
+        }
+        Ok(())
+    }
+
+    /// Generates random data and returns it as a Vec<u8>
+    pub fn generate_random_vec(&self, random_len: u32) -> Result<Vec<u8>> {
```
ionut-arm, line 26: I'm guessing no one would ask for more than 4 billion bytes?

ionut-arm (approved):
Thanks, looks pretty good, only one small request.

cryptoki/src/functions/random.rs
mjb3279: I would hope not...?  That value is gated by the length parameters to `C_GenerateRandom`.

cryptoki/src/functions/random.rs
```diff
@@ -10,6 +10,11 @@ use std::convert::TryInto;
 
 impl<'a> Session<'a> {
     /// Generates a random number and sticks it in a slice
+    ///
+    /// # Arguments
+    ///
+    /// * `random_slice` - The slice to stick the random data into.  The length of the slice represents
+    /// the number of bytes to obtain from the RBG
     pub fn generate_random_slice(&self, random_data: &mut [u8]) -> Result<()> {
```
mjb3279, line 18: Updated comment block for `generate_random_slice`

ionut-arm (approved):

PR #47, opened 2021-09-23
Merge main into devel

mjb3279:
Merging `main` into `devel` to keep `devel` up-to-date


hug-dev:
You can try but I think now (with your `Write` role), you should be able to rebase `devel` on top of `main` and force-push `devel` in the repo. If that does not work I can do it myself too. That's just to avoid the merge commit and keep the history linear but if we don't care that much we can just merge this.

mjb3279:
I guess it's up to you.  The rebase will be annoying, replaying each of the commits in `devel` onto `main`.  At some point `devel` will merge back into `main`.

I tend to prefer merges, but again, up to you :)

hug-dev:
I just rebased `devel` on top of `main` and force-pushed the branch. It's currently passing CI [now](https://github.com/parallaxsecond/rust-cryptoki/runs/3688721488), hope it will pass.

I figured it will be easier when we decide to lift the code-freeze because we will just be able to fast-forward the `main` branch with `devel` and delete `devel`.

mjb3279:
So the problem with rebasing is that you lose all the merge history in `devel`.  Not sure if it matters.

hug-dev:
Yeah not sure, as long as you don't lose commit history I'd say it's ok

Issue #42, opened 2021-09-02
Provide attribute type in return from `get_attribute_info`


mjb3279:
Currently `get_attribute_info` takes in a slice of `AttributeType`'s and returns a `Vec<AttributeInfo>` ,however other than indexing into the slice and assuming the corresponding attribute information occupies the same index , the returned data does not correlate the attribute length to said attribute.

I can think of two solutions (but my limited Rust experience tells me there's probably better ways of doing this):

1. Modify `AttributeInfo` to contain the `AttributeType` by modifying the enum variants to be either:
```rust
pub enum AttributeInfo {
    Unavailable(AttributeType),
    Available(AttributeType, usize),
}
```
or
```rust
pub enum AttributeInfo {
    Unavailable { type: AttributeType},
    Available { type: AttributeType, size: usize },
}
```

2. Return a map which is keyed on the attribute, such as:
```rust
pub get_attribute_info(
&self,
object: ObjectHandle,
attributes: &[AttributeType],
) -> Result<HashMap<AttributeType, AttributeInfo>>
```

Furthermore, `get_attribute_info` does not return whether or not `C_GetAttributeValue` returns `CKR_ATTRIBUTE_SENSITIVE` or `CKR_ATTRIBUTE_INVALID`.  Operationally, these may not be the most helpful returns (frankly the spec for `C_GetAttributeValue` leaves a lot to be desired when it comes to determining why an attribute is not returned) but I believe the wrapper library should not be masking away that information.

This part of the issue may warrant a separate ticket/discussion, but one possibility would be returning the CKR as part of a tuple (along with the attribute information) in the successful part of the Result enum.

hug-dev:
Hey!

Back from holidays 🌴 
Thanks for your contribution in the crate, that's really appreciated.

I thought that because the `get_attribute_info` keeps the order, it would be ok to check which attribute type maps to which attribute info but maybe it would be better to do it in the function already.

Something like [this](https://rosettacode.org/wiki/Hash_from_two_arrays#Rust) could be possible, having `keys` as the vector of `AttributeType` and `values` as the vector returned by `get_attribute_info`. If needed we could do the same with your 2 option above. The only thing I see with that is that you would have to allocate more data but that might be fine as `AttributeType` is `Copy` and small in size.

> Furthermore, get_attribute_info does not return whether or not C_GetAttributeValue returns CKR_ATTRIBUTE_SENSITIVE or CKR_ATTRIBUTE_INVALID.

[...]

PR #49, opened 2021-09-26
Added support for `C_SetPIN`

mjb3279:

### DIFF
--- a/cryptoki/src/functions/slot_token_management.rs
+++ b/cryptoki/src/functions/slot_token_management.rs
@@ -202,4 +202,25 @@ impl<'a> Session<'a> {
             .into_result()
         }
     }
+
+    /// Changes the PIN of either the currently logged in user or of the `CKU_USER` if no user is
+    /// logged in.
+    pub fn set_pin(&self, old_pin: &str, new_pin: &str) -> Result<()> {
+        let old_pin = Secret::new(CString::new(old_pin)?.into_bytes());
+        let new_pin_secret = Secret::new(CString::new(new_pin)?.into_bytes());
+        unsafe {
+            Rv::from(get_pkcs11!(self.client(), C_SetPIN)(
+                self.handle(),
+                old_pin.expose_secret().as_ptr() as *mut u8,
+                old_pin.expose_secret().len().try_into()?,
+                new_pin_secret.expose_secret().as_ptr() as *mut u8,
+                new_pin_secret.expose_secret().len().try_into()?,
+            ))
+            .into_result()?;
+        }
+        // Remove the old pin
+        self.client().clear_pin(self.slot());
+        // Set the new pin
+        self.client().set_pin(self.slot(), new_pin)
+    }
 }
--- a/cryptoki/tests/basic.rs
+++ b/cryptoki/tests/basic.rs
@@ -539,3 +539,22 @@ fn generate_random_test() -> Result<()> {
     assert!(!random_vec.iter().all(|&x| x == 0));
     Ok(())
 }
+
+#[test]
+#[serial]
+fn set_pin_test() -> Result<()> {
+    let (pkcs11, slot) = init_pins();
+
+    let mut flags = SessionFlags::new();
+
+    let _ = flags.set_serial_session(true).set_rw_session(true);
+    let session = pkcs11.open_session_no_callback(slot, flags)?;
+
+    pkcs11.set_pin(slot, "1234")?;
+    session.login(UserType::User)?;
+    session.set_pin("1234", "5678")?;
+    session.logout();
+    pkcs11.set_pin(slot, "5678");
+    session.login(UserType::User)?;
+    Ok(())
+}


cryptoki/src/functions/slot_token_management.rs
```diff
@@ -202,4 +202,25 @@ impl<'a> Session<'a> {
             .into_result()
         }
     }
+
+    /// Changes the PIN of either the currently logged in user or of the `CKU_USER` if no user is
+    /// logged in.
+    pub fn set_pin(&self, old_pin: &str, new_pin: &str) -> Result<()> {
+        let old_pin = Secret::new(CString::new(old_pin)?.into_bytes());
```
wiktor-k, line 209: I'm not super familiar with the `Secret` API but what's the point in wrapping the PIN and then a couple of lines below `expose`ing it?

If `Secret` is protecting the bytes in memory maybe this function should take a `Secret` instead of string slice (thus the user can work with Secrets directly)... :thinking:

mjb3279: So I modeled this after the `init_pin` method in `Session` and the `init_token` method in `Pkcs11`.  In both of those, the provided PINs are converted into `Secret`s and then accessed.

You raise a fair point, though.  Maybe @hug-dev can weigh in on why there was a decision to use `Secret` in the way it was.

wiktor-k: Good point Mike. In this case I think the reasonable approach would be to create an issue about this problem and ignore it temporarily for this PR.

FTR tomorrow is a team telco with Parsec people and we could bring this matter up there.

mjb3279: I'm of the opinion that `Secret` isn't necessary for two reasons:

1. The calling application still has the `pin` held within it's memory and there's no guarantee that they protected it in any way
2. Though this is entirely dependent upon the design of the underlying PKCS#11 library, that library may also be holding onto the `pin` in memory, also unprotected.  I think it depends on how the library and token implement the session/process coherency requirements of the PKCS#11 spec.

Furthermore, I don't think we need the `set_pin` function in `Pkcs11` for the same reasons mentioned above.

However, if there are underlying requirements for this functionality, that's understandable.

wiktor-k: If this makes the library simpler to use and doesn't lower the security I'm all for that 👍

Although let's wait for @hug-dev as he probably has more context on this.

hug-dev: I think the main reason why we do that here is to zeroize the `CString` (allocated in this function) on `Drop`, `Secret` implementing that by default. Otherwise we would have to do it manually. Maybe [`Zeroizing`](https://docs.rs/zeroize/1.4.2/zeroize/struct.Zeroizing.html) in enough there?

I agree with your point 1 and 2 above, the use of `Secret` here is to try to not leak secrets in RAM but the others could still do it :/

> Furthermore, I don't think we need the set_pin function in Pkcs11 for the same reasons mentioned above.

Happy to discuss this!

hug-dev (approved):
Thanks!!

cryptoki/src/functions/slot_token_management.rs
ionut-arm: > The calling application still has the pin held within it's memory and there's no guarantee that they protected it in any way

True, if the client application doesn't care then it doesn't matter whether we use `Secret` or not, but if they do care and we don't do it, they can't do anything about it (apart from raising a PR/issue and waiting for a new release).

> Maybe Zeroizing in enough there?

We could, but it's a pain because you have to make sure, before every `?`, to clear out the memory, so 5 places in total including the final return from the function. And if we change something in the future we might forget to update that too. We could implement our own wrapper around `CString` or whatever other type we need and just derive `Zeroizing` on that, then we won't need to use `expose_secret` everywhere, but the result is probably kinda similar.

ionut-arm (approved):
👍🏻

cryptoki/src/functions/slot_token_management.rs
mjb3279: > but if they do care and we don't do it, they can't do anything about it (apart from raising a PR/issue and waiting for a new release).

That still leaves the final layer, that being the actual library, and if it's for an actual HSM, that's closed source and you just give up your secrets to the ether.

> Maybe Zeroizing in enough there?

I'd say we shouldn't store the pin at all (that is, get rid of the `pins` `HashMap` in `Pkcs11`) and just call `as_ptr()` against the `pin` `str` slice in the concerned functions (`C_InitToken`, `C_SetPIN`, `C_InitPIN`, and `C_Login`), that is if that's possible.

hug-dev: I am happy to discuss this!! I can't remember now exactly why we choose to do that, but maybe we could merge this PR as it is now (if you are ok with the rest) and open a new issue to expand on those things?

mjb3279: Sounds good!  I do think this, and my other PR, #42 will cause a merge conflict, so whichever loses the race I'll fix the conflict and commit it.

I'll open an issue to discuss the PIN handling.

Issue #50, opened 2021-09-27
PIN Handling


mjb3279:
This issue is opened to continue the discussion started in #49 in regards to how user PINs are handled by the cryptoki library wrapper.

Currently, when a user wishes to login, they must first call `set_pin` in the `Pkcs11` struct which stores the pin in a `HashMap`, keyed to the `Slot`.  Then the user may call `login` on the created `Session` object.

Any other function that directly handles pins (those that expose the `C_InitToken`, `C_InitPIN` and `C_SetPIN` functionality) wraps the provided PIN within a `Secret` structure and passes a pointer to the appropriate PKCS#11 function.  Upon completion of the function, the `Secret` structure will `drop` and zeroize the memory.

My first proposition is to drop the `pins` `HashMap` member in the `Pkcs11` struct and provide a `pins` parameter to `Session::login`

First, if the user wishes to change which user is logged in (say going from `CKU_USER` to `CKU_SO`) the must call `set_pin` to change the pin and then call `login`. This requires the implementer to track which pin is currently being stored by the library to ensure they are logging in with the correct pin.

Secondly, and this is a personal preference thing, I do not believe the library should be storing the user's passwords.  I believe the user should be the one that controls where their sensitive data is stored and be responsible for it, rather than ourselves.

Thirdly, and this is one of those weird PKCS#11 corner cases, there are situations where you may not want to provide a password to `C_Login`.  For example, if the flag `CKF_PROTECTED_AUTHENTICATION_PATH` is set in token flags, that indicates there's a separate way of logging into the token (e.g. finger print reader, smart card reader, etc).  In this situation, you must pass a `NULL_PTR` to `C_Login` to take advantage of that alternative authentication path.

My second proposition is to drop `Secret` and instead just pass the pointer from the provided `str` slice to the appropriate `C_` function.  If this isn't possible, then anytime a pin is provided and needs to be copied somewhere on the stack that is controlled by the library, then use `Secret`.

mjb3279:
To just add one more thing, I think we should wrap any provided attributes used in object creation (e.g. `Session::create_object`) in `Secret` to ensure we do erase any sensitive data the user provides.  An example would be creating an AES key using `Session::create_object`.  In this situation, the operator is required to provide the value of the AES key in the `CKA_VALUE` attribute.

hug-dev:
Thanks for the issue!

[...]

PR #52, opened 2021-09-28
Fixes to address #50

mjb3279:
- Removed the PIN and session storage from `Pkcs11`
  - Allow the cryptoki token to maintain state
- Removed methods in `Pkcs11` that handled PIN storage
- Added `pin` as an argument to `login`
- Removed calling `logout` when a `Session` is `Drop`ed
- Updated tests

### DIFF
--- a/cryptoki/src/functions/object_management.rs
+++ b/cryptoki/src/functions/object_management.rs
@@ -134,9 +134,8 @@ impl<'a> Session<'a> {
     /// let mut flags = SessionFlags::new();
     /// let _ = flags.set_rw_session(true).set_serial_session(true);
     ///
-    /// pkcs11.set_pin(slot, "1234").unwrap();
     /// let session = pkcs11.open_session_no_callback(slot, flags).unwrap();
-    /// session.login(UserType::User);
+    /// session.login(UserType::User, "fedcba");
     ///
     /// let empty_attrib= vec![];
     /// if let Some(object) = session.find_objects(&empty_attrib).unwrap().get(0) {
--- a/cryptoki/src/functions/object_management.rs
+++ b/cryptoki/src/functions/object_management.rs
@@ -135,7 +135,7 @@ impl<'a> Session<'a> {
     /// let _ = flags.set_rw_session(true).set_serial_session(true);
     ///
     /// let session = pkcs11.open_session_no_callback(slot, flags).unwrap();
-    /// session.login(UserType::User, "fedcba");
+    /// session.login(UserType::User, Some("fedcba"));
     ///
     /// let empty_attrib= vec![];
     /// if let Some(object) = session.find_objects(&empty_attrib).unwrap().get(0) {
--- a/cryptoki/src/functions/session_management.rs
+++ b/cryptoki/src/functions/session_management.rs
@@ -43,17 +43,25 @@ impl<'a> Session<'a> {
     }
 
     /// Log a session in.
-    pub fn login(&self, user_type: UserType, pin: &str) -> Result<()> {
-        let pin_ptr = match pin.len() {
-            0 => std::ptr::null_mut(),
-            _ => pin.as_ptr() as *mut u8,
+    ///
+    /// # Arguments
+    ///
+    /// * `user_type` - The type of user to log in as
+    /// * `pin` - The PIN to use, or `None` if you wish to use the protected authentication path
+    ///
+    /// _NOTE: By passing `None` into `login`, you must ensure that the
+    /// [CKF_PROTECTED_AUTHENTICATION_PATH] flag is set in the `TokenFlags`._
+    pub fn login(&self, user_type: UserType, pin: Option<&str>) -> Result<()> {
+        let (pin, pin_len) = match pin {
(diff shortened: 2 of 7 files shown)


cryptoki/src/functions/session_management.rs
```diff
@@ -42,20 +42,26 @@ impl<'a> Session<'a> {
...
-    ///
-    /// Do not fail if the user is already logged in. It happens if another session on the same slot
-    /// has already called the log in operation. Record the login call and only log out when there
-    /// aren't anymore sessions requiring log in state.
-    pub fn login(&self, user_type: UserType) -> Result<()> {
-        self.client().login(self, user_type)
+    /// Log a session in.
+    pub fn login(&self, user_type: UserType, pin: &str) -> Result<()> {
+        let pin_ptr = match pin.len() {
+            0 => std::ptr::null_mut(),
```
hug-dev, line 48: Should we distinguish between no pin and a pin of size 0? As in between `pPin` with a valid address but the len being 0 and `pPin` with the NULL value.

cryptoki/tests/basic.rs
```diff
@@ -27,7 +28,9 @@ type Result<T> = std::result::Result<T, ErrorWithStacktrace>;
 #[test]
 #[serial]
 fn sign_verify() -> Result<()> {
-    let (pkcs11, slot) = init_pins();
+    let user_pin = "fedcba";
+    let so_pin = "abcdef";
+    let (pkcs11, slot) = init_pins(so_pin, user_pin);
```
hug-dev, line 33: Maybe we could defined `USER_PIN` and `SO_PIN` as static values in the Rust code and `init_pins` not taking any parameters (since we don't really care what those values are for tests I think).

hug-dev (commented):
Thanks!! Feels good to remove this logic. Even if I brace myself for the Parsec changes that will use this 💪

mjb3279:
Yeah I'll throw `Option` around the `&str` and use that to distinguish between pin and no pin

cryptoki/src/functions/session_management.rs
```diff
@@ -43,17 +43,25 @@ impl<'a> Session<'a> {
...
-            0 => std::ptr::null_mut(),
-            _ => pin.as_ptr() as *mut u8,
+    ///
+    /// # Arguments
+    ///
+    /// * `user_type` - The type of user to log in as
+    /// * `pin` - The PIN to use, or `None` if you wish to use the protected authentication path
+    ///
+    /// _NOTE: By passing `None` into `login`, you must ensure that the
+    /// [CKF_PROTECTED_AUTHENTICATION_PATH] flag is set in the `TokenFlags`._
```
hug-dev, line 53: 👍

hug-dev (approved):
Thanks!

mjb3279:
Ok sorry missed a comment and it bugged out.  All fixed now

wiktor-k:
Yep, this looks good 👍 I can run tests on my tokens tomorrow morning but OTOH this PR looks simple and straightforward. Thanks Mike!

mjb3279:
Glad to help!  And if you guys need any assistance with the changes on your other projects, feel free to reach out!

wiktor-k (approved):

wiktor-k:
> And if you guys need any assistance with the changes on your other projects, feel free to reach out!

What is your area of interest except for PCKS11? Parsec has quite a diverse set of crypto backends... not sure if issues/PRs is the right medium to discuss this though :)

cryptoki/src/functions/object_management.rs
```diff
@@ -135,7 +135,7 @@ impl<'a> Session<'a> {
     /// let _ = flags.set_rw_session(true).set_serial_session(true);
     ///
     /// let session = pkcs11.open_session_no_callback(slot, flags).unwrap();
-    /// session.login(UserType::User, "fedcba");
+    /// session.login(UserType::User, Some("fedcba"));
```
wiktor-k, line 138: By the way, it seems that if PIN instead of being `&str` would be generic `T` and the parameter was specialized as `where T: Into<Option<&str>` the API could stay the same while allowing `None`:

```rust
fn just_str(s: &str) {
    println!("just s: {}", s);
}

fn into_str<'a, T>(s: T) where T: Into<Option<&'a str>> {
    println!("into s: {:#?}", s.into());
}

fn main() {
    just_str("test");
    
    into_str(Some("test"));
    into_str("test"); // <-- better ergonomics, no need for explicit Some("test")
    into_str(None);
}
```

[Link to playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9ee2a5632ea8e33e6d9781bf0d14e0a0)

hug-dev: True, could be an `Option`. Happy for this to be added in a subsequent PR 😄 Didn't know about it!

hug-dev:
> And if you guys need any assistance with the changes on your other projects, feel free to reach out!

If you are interested and have some time, you could look at importing the latest `cryptoki` `main` (so containing this changes) to Parsec! Only if you have time, otherwise we will do it one day or another. It would be nice to have your views on how we use PKCS11 there 😃

PR #56, opened 2021-10-31
Module tree hygiene

### LINKED ISSUE #55: Module tree structure makes docs difficult to navigate
The overall structure of the cryptoki crate's module tree consists of several relatively long branches with often empty intermediate modules. The actual content is largely buried on distant leaves in counterintuitive locations (e.g., Session declared among data types). When I first encountered this crate, I genuinely assumed someone was just reserving the name and it was mostly empty because the places I assumed things would be (e.g., definitions of object types) were empty. That's a shame, because it exists for the same reason I was looking in the first place: the current pkcs11 crate isn't idiomatic and exposes an unsafe, easy to misuse interface.

This crate shows a lot of promise toward addressing that problem, but its documentation structure (which is a direct result of its internal structure) is (IMO) a serious barrier to it being adopted widely.

I'll be submitting a PR shortly that attempts to address this issue.

vkkoskie:
This PR introduces breaking changes. It attempts to resolve Issue #55, which has a more complete description of its rationale.

The changes are almost entirely structural. No function bodies or prototypes have changed, and the overall API is identical. The break comes only in the paths through which external users of the crate access its items. (e.g., `use cryptoki::types::session::Session` is now `use cryptoki::session::Session` but the Session struct is unchanged). The overall effect is a wider, shallower tree, where each module contains conceptually linked types. This produces a much more intuitive "front door" for documentation.

Because file move diffs are often difficult to read, the PR is broken up into a series of commits each with its own detailed commit message to make review easier.

As I was making these changes, I also noticed a general misuse of visibility markers, and I suspect that more of the crate is exposed as pub than is ideal. While I did make several necessary changes from private to pub(crate) in order for moved items to be usable with the new structure, I only made one visibility change unrelated to this PR's primary goal. This is the final commit to the branch and removes the macro_export directive to take the get_pcks11 macro out of the public API (where it wouldn't have compiled anyway).

closes #55


mjb3279:
Any reason not to just delete the `objects` module all together as it's empty and I, at the moment, cannot think of a good reason to keep it?

ionut-arm (approved):
The changes look sensible to me! Also, I'm happy with the `objects` module being removed.

The imports for that failing doctest should also be fixed, though

vkkoskie:
New commits. More file moves, which make diffs look much more drastic than they really are. Here's the summary:

To try and respect the prior organization as much as possible, the methods previously located in the `functions` module all still occupy a file of the same name, indicating which section of the specification they belong to. However, all have been changed from methods to private functions in a submodule of the struct they belong to. The struct definition itself exposes these functions through method stubs, which are now also the collection point for method documentation.

This provides what I think is a nice balance between the needs of developers and users of the crate. Users see a uniform struct uncluttered by impl annotations. This makes the struct itself monolithic which would be cumbersome to developers, but instead they get to work in the broken out files nearby.

I wasn't sure if performance was of any concern, but I went ahead and removed the layer of indirection introduced by stubbing by forcing explicit inlining. Even if performance isn't a big concern, it does no harm.

I also found what I *think* is a bug in `Session::close()`, which I believe was intended to take ownership of `self` to trigger a call to `drop()` the same way the library context does. It probably should have been its own Issue/PR, but I didn't think it would be a big deal to include it.

cryptoki/src/session/mod.rs
```diff
@@ -88,7 +88,7 @@ impl Session<'_> {
 
     /// Close a session
     /// This will be called on drop as well.
-    pub fn close(&self) {}
+    pub fn close(self) {}
```
mjb3279, line 91: That's a good catch, otherwise `close()` really didn't do anything.

mjb3279:
I think I buy why you made those changes in `b711d0c`.  It's definitely drastic but the docs that are generated are significantly improved.

cryptoki/src/session/mod.rs
```diff
@@ -0,0 +1,478 @@
...
+    ///
+    /// This function will return a Vector of [AttributeInfo] enums that will either contain
+    /// the size of the requested attribute, [AttributeInfo::TypeInvalid] if the attribute is not a
+    /// valid type for the object, or [AttributeInfo::Sensitive] if the requested attribute is
+    /// sensitive and will not be returned to the user.
+    ///
+    /// The list of returned attributes is 1-to-1 matched with the provided vector of attribute
+    /// types.  If you wish, you may create a hash table simply by:
+    ///
+    /// ```rust
```
mjb3279, line 148: Let's replace this `rust` with `no_run`

mjb3279 (changes requested):

cryptoki/src/session/mod.rs
mjb3279:
```suggestion
    /// ```no_run
```

vkkoskie:
Hmm. Github CI is unhappy with a part of the code I didn't touch. Maybe the rust image updated to a new version? Just a guess.

cryptoki/src/session/mod.rs
```diff
@@ -0,0 +1,478 @@
...
+    /// use cryptoki::Pkcs11;
+    /// use cryptoki::types::locking::CInitializeArgs;
+    /// use cryptoki::types::object::AttributeType;
+    /// use cryptoki::types::session::UserType;
+    /// use cryptoki::types::SessionFlags;
+    /// use std::collections::HashMap;
+    /// use std::env;
```
mjb3279, line 155: I think because of your changes to the layout of the crate, these `use` statements are failing

mjb3279, line 155:
```suggestion
    /// use cryptoki::context::Pkcs11;
    /// use cryptoki::context::CInitializeArgs;
    /// use cryptoki::object::AttributeType;
    /// use cryptoki::session::UserType;
    /// use cryptoki::session::SessionFlags;
    /// use std::collections::HashMap;
    /// use std::env;
```

mjb3279:
Went ahead and made the changes and force pushed an updated commit to contain the 'magic' 'Signed-off-by' line.

mjb3279 (approved):

PR #58, opened 2021-11-03
Adds management session functionality

### LINKED ISSUE #57: Solution for `Session` lifetimes
The `Session` struct contained an explicit lifetime requirement on an instance of the `Pkcs11` struct.  Overall, this makes sense; you do not want an instance of a `Session` to outlive the instance of `Pkcs11`.  However, this explicit lifetime requirement presents some interesting problems during implementation.

Take for example, a certificate authority that utilizes a PKCS#11 compliant token that contains a private key used to sign requests coming in.  This certificate authority is multi-threaded so it can handle multiple requests at once.  Each thread is required to open a session on the token in order to access the underlying object and wield it in order to sign the request.  Now in order to wield this object, that session needs to be logged in.

One way of handling this is to simply login each session when it's created.  If the session is already logged in, the login call will fail with `CKR_USER_ALREADY_LOGGED_IN`.  That's annoying, but it can be handled in code.  However, some tokens can't be logged in simply by password, some require smart cards inserted into the token, some require biometerics, etc.  The point is that sometimes, login is interactive.  And once the final session closes, that token is now logged out and a user is required to log back in.

To get around this, many CA application store off a 'management' session.  This session is opened shortly after the application is started and the PKCS#11 library is initialized.  This session is logged in and lives for as long as the application runs.  The threads simply need to open a new session, do what they need to do, close the session and end.

The issue now, is say this CA stores the instance of `Pkcs11` in a struct.  Where can the instance of the 'management' session live?

```rust
struct MyCA {
  p11: Pkcs11,
  mgmt_session: Session,
}
```

This won't work, due to the explicit lifetime requirement of `Session`.  This [stack overflow](https://stackoverflow.com/questions/32300132/why-cant-i-store-a-value-and-a-reference-to-that-value-in-the-same-struct) question goes into more detail.

One possible solution is to add functions to the crate that allow the user to create the session without having to hold onto it.  Something like `create_management_session(slot: Slot, read_write: bool, user: UserType, pin: Option<&str>)` and `close_management_session(slot: Slot)`.

mjb3279:
Resolves Issue #57 

- Added `open_management_session`
- Added `close_management_session`
- Added test to test newly added functionality

closes #57 

### DIFF
--- a/cryptoki/src/functions/session_management.rs
+++ b/cryptoki/src/functions/session_management.rs
@@ -82,9 +82,9 @@ impl Pkcs11 {
                     // Close the session if we can't login and return the resulting error
                     // Not using `get_pkcs11` here because the macro uses a '?' which is not
                     // allowed in a closure that does not return a Result.
-                    // Justification for unwrap -> If C_CloseSession doesn't exist, the
-                    // library is frankly unusable.
-                    let _ = self.function_list.C_CloseSession.unwrap()(session_handle);
+                    if let Some(func) = self.function_list.C_CloseSession {
+                        let _ = func(session_handle);
+                    }
                     e
                 })?;
             }
--- a/cryptoki/src/functions/session_management.rs
+++ b/cryptoki/src/functions/session_management.rs
@@ -9,12 +9,11 @@ use crate::types::slot_token::Slot;
 use crate::types::SessionFlags;
 use crate::Pkcs11;
 use crate::Result;
-use cryptoki_sys::CK_SESSION_INFO;
+use cryptoki_sys::{CK_SESSION_HANDLE, CK_SESSION_INFO};
 use std::convert::TryInto;
 
 impl Pkcs11 {
-    /// Open a new session with no callback set
-    pub fn open_session_no_callback(&self, slot_id: Slot, flags: SessionFlags) -> Result<Session> {
+    fn open_session(&self, slot_id: Slot, flags: SessionFlags) -> Result<CK_SESSION_HANDLE> {
         let mut session_handle = 0;
 
         unsafe {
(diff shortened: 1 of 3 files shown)


cryptoki/src/functions/session_management.rs
```diff
@@ -28,8 +27,93 @@ impl Pkcs11 {
...
+                    pin_len.try_into()?,
+                ))
+                .into_result()
+                .map_err(|e| {
+                    // Close the session if we can't login and return the resulting error
+                    // Not using `get_pkcs11` here because the macro uses a '?' which is not
+                    // allowed in a closure that does not return a Result.
+                    // Justification for unwrap -> If C_CloseSession doesn't exist, the
+                    // library is frankly unusable.
+                    let _ = self.function_list.C_CloseSession.unwrap()(session_handle);
```
ionut-arm, line 87: I think even if the library is unusable we shouldn't bomb out. Maybe something like:
```rust
self.function_list.C_CloseSession.ok_or(Error::NullFunctionPointer)?(session_handle);
```
?

cryptoki/src/functions/session_management.rs
```diff
@@ -28,8 +27,93 @@ impl Pkcs11 {
...
+        slot_id: Slot,
+        flags: SessionFlags,
+        user_type: UserType,
+        pin: Option<&str>,
+    ) -> Result<()> {
+        let mut map = self
+            .management_sessions
+            .lock()
+            .map_err(|_e| crate::Error::MutexPoisonError)?;
+        if let std::collections::hash_map::Entry::Vacant(_) = map.entry(slot_id) {
```
ionut-arm, line 67: Is it possible to replace this big `if ... else ...` by using [`contains`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.contains_key) and just returning if the key is already there?

cryptoki/src/functions/session_management.rs
```diff
@@ -28,8 +27,93 @@ impl Pkcs11 {
...
-        Ok(Session::new(session_handle, self))
+    /// Closes the management session on the slot
+    ///
+    /// # Errors
+    ///
+    /// If the session fails to close, the typical [`Error::Pkcs11`] will return.  If a management
+    /// session does not exist on this token, then [`Error::ManagementSessionDoesNotExist`] will
+    /// be returned.
+    ///
+    /// Even if the close session call fails, the value will still be removed from the map
```
ionut-arm, line 107: What happens in that case? Does the app remain logged in?

cryptoki/src/lib.rs
```diff
@@ -70,6 +75,7 @@ pub struct Pkcs11 {
     #[derivative(Debug = "ignore")]
     _pkcs11_lib: cryptoki_sys::Pkcs11,
     function_list: cryptoki_sys::_CK_FUNCTION_LIST,
+    management_sessions: ManagementSessionMap,
```
ionut-arm, line 78: I think you should add logic in `Drop`, below, to close any outstanding sessions before finalizing (unless you think it doesn't make any difference?)

ionut-arm (approved):

cryptoki/src/functions/session_management.rs
mjb3279: The problem is that you can't use '?' in this closure because the closure returns an `Error` not a `Result`.

Maybe a better solution would be to log the error (which I personally don't like because I don't like libraries logging anything unless explicitly told to do so) or maybe just ignore it?

cryptoki/src/functions/session_management.rs
mjb3279: Possibly, yes.  The user could then create another session and call logout on that.

I don't think this is really that much of an issue as C_CloseSession only really fails if the session doesn't exist anymore or something really catastrophic occurred.

cryptoki/src/lib.rs
mjb3279: YES!  Forgot to do that!

mjb3279: Technically it shouldn't make a difference, calling Finalize should close all sessions anyways, but that would be cleaner for sure.

cryptoki/src/functions/session_management.rs
ionut-arm: ah, yes, that's true, and it might not be desirable to return that error instead of the "original" one anyway. Just ignore it if `C_CloseSession` doesn't exist

cryptoki/src/functions/session_management.rs
mjb3279: So the reason for this was that `clippy` threw a fit when I used `contains_key` and then after that called `insert` -> https://rust-lang.github.io/rust-clippy/master/#map_entry

mjb3279: This was clippy's suggestion on how to fix the lint error

mjb3279: Actually I should be using the result of Vacant() to insert something into that entry

mjb3279:
Due to the changes recently merged into `main`, I'm going to close this PR and start a new one

PR #59, opened 2021-11-04
Implemented new way of holding the context within the session

### LINKED ISSUE #57: Solution for `Session` lifetimes

mjb3279:
- Session now holds an Arc pointing to a Pkcs11 instance
- Pkcs11::new now returns an Arc<Self>

Resolves #57

### DIFF
--- a/cryptoki/src/context/mod.rs
+++ b/cryptoki/src/context/mod.rs
@@ -42,8 +42,8 @@ pub(crate) struct Pkcs11Impl {
     // Even if this field is never read, it is needed for the pointers in function_list to remain
     // valid.
     #[derivative(Debug = "ignore")]
-    _pkcs11_lib: Arc<cryptoki_sys::Pkcs11>,
-    pub(crate) function_list: Arc<cryptoki_sys::_CK_FUNCTION_LIST>,
+    _pkcs11_lib: cryptoki_sys::Pkcs11,
+    pub(crate) function_list: cryptoki_sys::_CK_FUNCTION_LIST,
 }
 
 /// Main PKCS11 context. Should usually be unique per application.
@@ -69,8 +69,8 @@ impl Pkcs11 {
 
             Ok(Pkcs11 {
                 impl_: Arc::new(Pkcs11Impl {
-                    _pkcs11_lib: Arc::new(pkcs11_lib),
-                    function_list: Arc::new(*list_ptr),
+                    _pkcs11_lib: pkcs11_lib,
+                    function_list: *list_ptr,
                 }),
             })
         }
--- a/cryptoki/src/context/mod.rs
+++ b/cryptoki/src/context/mod.rs
@@ -6,6 +6,7 @@
 macro_rules! get_pkcs11 {
     ($pkcs11:expr, $func_name:ident) => {
         ($pkcs11
+            .impl_
             .function_list
             .$func_name
             .ok_or(crate::error::Error::NullFunctionPointer)?)
@@ -34,20 +35,26 @@ use std::mem;
 use std::path::Path;
 use std::sync::Arc;
 
-/// Main PKCS11 context. Should usually be unique per application.
 #[derive(Derivative)]
 #[derivative(Debug)]
-pub struct Pkcs11 {
+// Implementation of Pkcs11 class that can be enclosed in a single Arc
+pub(crate) struct Pkcs11Impl {
     // Even if this field is never read, it is needed for the pointers in function_list to remain
     // valid.
     #[derivative(Debug = "ignore")]
-    _pkcs11_lib: cryptoki_sys::Pkcs11,
-    pub(crate) function_list: cryptoki_sys::_CK_FUNCTION_LIST,
+    _pkcs11_lib: Arc<cryptoki_sys::Pkcs11>,
+    pub(crate) function_list: Arc<cryptoki_sys::_CK_FUNCTION_LIST>,
+}
+
+/// Main PKCS11 context. Should usually be unique per application.
+#[derive(Clone, Debug)]
+pub struct Pkcs11 {
+    pub(crate) impl_: Arc<Pkcs11Impl>,
 }
 
 impl Pkcs11 {
     /// Instantiate a new context from the path of a PKCS11 dynamic llibrary implementation.
-    pub fn new<P>(filename: P) -> Result<Arc<Self>>
+    pub fn new<P>(filename: P) -> Result<Self>
     where
         P: AsRef<Path>,
     {
@@ -60,10 +67,12 @@ impl Pkcs11 {
 
             let list_ptr = *list.as_ptr();
 
-            Ok(Arc::new(Pkcs11 {
-                _pkcs11_lib: pkcs11_lib,
-                function_list: *list_ptr,
-            }))
+            Ok(Pkcs11 {
+                impl_: Arc::new(Pkcs11Impl {
+                    _pkcs11_lib: Arc::new(pkcs11_lib),
+                    function_list: Arc::new(*list_ptr),
+                }),
+            })
         }
     }
 
@@ -124,11 +133,7 @@ impl Pkcs11 {
     }
 
     /// Open a new session with no callback set
-    pub fn open_session_no_callback(
-        self: &Arc<Self>,
-        slot_id: Slot,
-        flags: SessionFlags,
-    ) -> Result<Session> {
+    pub fn open_session_no_callback(&self, slot_id: Slot, flags: SessionFlags) -> Result<Session> {
         session_management::open_session_no_callback(self, slot_id, flags)
     }
 }
--- a/cryptoki/src/context/mod.rs
+++ b/cryptoki/src/context/mod.rs
@@ -33,6 +33,7 @@ use derivative::Derivative;
 use log::error;
 use std::mem;
 use std::path::Path;
+use std::ptr;
 use std::sync::Arc;
 
 #[derive(Derivative)]
@@ -44,6 +45,30 @@ pub(crate) struct Pkcs11Impl {
     #[derivative(Debug = "ignore")]
     _pkcs11_lib: cryptoki_sys::Pkcs11,
     pub(crate) function_list: cryptoki_sys::_CK_FUNCTION_LIST,
+}
+
+impl Pkcs11Impl {
+    // Private finalize call
+    #[inline(always)]
+    fn finalize(&self) -> Result<()> {
+        unsafe {
+            Rv::from(self
+                .function_list
+                .C_Finalize
+                .ok_or(Error::NullFunctionPointer)?(
+                ptr::null_mut()
+            ))
+            .into_result()
+        }
+    }
+}
+
+impl Drop for Pkcs11Impl {
+    fn drop(&mut self) {
+        if let Err(e) = self.finalize() {
+            error!("Failed to finalize: {}", e);
+        }
+    }
 }
 
 /// Main PKCS11 context. Should usually be unique per application.
@@ -135,13 +160,5 @@ impl Pkcs11 {
     /// Open a new session with no callback set
     pub fn open_session_no_callback(&self, slot_id: Slot, flags: SessionFlags) -> Result<Session> {
         session_management::open_session_no_callback(self, slot_id, flags)
-    }
-}
-
-impl Drop for Pkcs11 {
-    fn drop(&mut self) {
-        if let Err(e) = general_purpose::finalize_private(self) {
-            error!("Failed to finalize: {}", e);
-        }
     }
 }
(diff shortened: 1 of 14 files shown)


vkkoskie:
This works, but just for the sake of clarity I was thinking an internal `Arc` around just the `function_list` element, which is what gets cloned into the `Session`. An outer `Arc` allows user code to clone the context. That wasn't possible before this change, and I don't know if that behavior goes against any threading restrictions you were assuming. I'm also not totally clear on what role the `_pkcs11_lib` member is actually playing that might require it to be transferred too.

mjb3279:
> threading restrictions you were assuming

My only thought on threading restrictions would be that the user would be able to create new sessions in threads they create, based off of the same context.  So the context itself would need to be shared across any thread opened for this purpose.  I made the entire thing an `Arc` in order to capture all elements of the context.

>  I'm also not totally clear on what role the _pkcs11_lib member is actually playing that might require it to be transferred too.

This effectively holds open the shared library (either a `.so` or a `.dll`)

vkkoskie:
I was looking at the cryptoki-sys crate for the first time. Curious why doesn't it hold its own function list? It just seems fragile in ways that it doesn't need to be.

And this still works with this somewhat common pattern if you don't have a strong need to expose the Arc:
```rust
#[derive(Clone)]
#[repr(transparent)]
pub struct Pcks11(Arc<Inner>);
struct Inner { // private
   // old struct contents
}
```

mjb3279:
> I was looking at the cryptoki-sys crate for the first time. Curious why doesn't it hold its own function list? It just seems fragile in ways that it doesn't need to be.

The majority of that crate was auto-generated by bindgen.

> And this still works with this somewhat common pattern if you don't have a strong need to expose the Arc:

I can look into this

ionut-arm:
> I was looking at the cryptoki-sys crate for the first time. Curious why doesn't it hold its own function list? It just seems fragile in ways that it doesn't need to be.

Because it's supposed to be an FFI-only layer crate, as described [here](https://kornel.ski/rust-sys-crate): 

> The task of the sys crates is expose a minimal low-level C interface to Rust (FFI) and to tell Cargo how to link with the library. Adding higher-level, more Rust-friendly interfaces for the libraries is left to "wrapper" crates built as a layer on top of the sys crates

vkkoskie:
Yes, I understand its role. But doesn't that role include owning the function list it exposes? They have the same lifetime, but rust can't detect that because of the FFI boundary. So the list should either need explicit lifetime annotations to prevent it from being used unsafely or the library should maintain ownership so they're invalidated at the same time.

mjb3279:
The library does own the function list it exposes.  It's returning a pointer to it.  The 'lifetime', if you will, is maintained by keeping an instance of `cryptoki_sys::Pkcs11` as `_pkcs11_lib`.

vkkoskie:
Maybe instead of "owns" I should have said "manages". As in, the callbacks are cached inside the FFI layer struct and accessed through it, rather than having two members in the wrapper. This is effectively the self-reference problem described in the stackoverflow link in #57. The wrapper struct contains both a thing (the FFI lib struct) and a "reference" into that thing. Except instead of an actual reference that the borrow checker could get angry about, it's a raw pointer whose safety you have to guarantee manually.

Also, in case I"m coming across as argumentative, I'm really just curious. I haven't had the chance to work with bindgen or deal with FFI issues, so I'm trying to make sure I understand why things that look "bad" are maybe necessary.

mjb3279:
I think this is one of those things that will always look bad because of how the C library implements things.  Normally, the C library contains a static struct that gets defined at compile time and you are just receiving a pointer to some piece of memory within the library.  I guess you could change how the crate is implemented to hold it by creating a wrapper struct around the `crypto_sys::Pkcs11` struct, but then I think it would like the same as `cryptoki::Pkcs11`.  But I think that gets away from the pure FFI only crate ideal.

ionut-arm:
Indeed, I think a big part of the issue is down to how the spec defined things, which in the end puts limits on how you can (sanely and nicely) deal with it in Rust. Something like the TPM spec, where you don't have to retrieve a list of pointers to functions that form the API would've been nice, alas such is PKCS11.

> Also, in case I"m coming across as argumentative, I'm really just curious. I haven't had the chance to work with bindgen or deal with FFI issues, so I'm trying to make sure I understand why things that look "bad" are maybe necessary.

No, it's always good to have people challenge the design/interface/implementation/... - it might not always get a speedy fix if we want to maintain things stable for a while, but at least being aware of alternatives and best practices is certainly better than not being aware :) And this is a prime reason for having a project open sourced anyway. Apologies if I've been curt, there's been a deluge of stuff coming up this week, difficult to keep track of everything.

cryptoki/src/session/mod.rs
```diff
@@ -35,37 +35,37 @@ pub use flags::*;
 /// threads. A Session needs to be created in its own thread or to be passed by ownership to
 /// another thread.
 #[derive(Debug)]
-pub struct Session<'a> {
+pub struct Session {
     handle: CK_SESSION_HANDLE,
-    client: &'a Pkcs11,
+    client: Pkcs11,
     // This is not used but to prevent Session to automatically implement Send and Sync
     _guard: *mut u32,
```
ionut-arm, line 42: Could you get rid of this and simply do an `impl !Sync for Session {}` below?

cryptoki/src/context/mod.rs
```diff
@@ -32,18 +33,50 @@ use derivative::Derivative;
...
+impl Drop for Pkcs11Impl {
+    fn drop(&mut self) {
+        if let Err(e) = self.finalize() {
+            error!("Failed to finalize: {}", e);
+        }
+    }
+}
+
+/// Main PKCS11 context. Should usually be unique per application.
+#[derive(Clone, Debug)]
```
ionut-arm, line 75: Is it needed/safe to make this clonable? I think we only have methods that require a non-mutable reference, so one instance could be used across threads. For creating sessions we could just build another `Pkcs11` on the fly.

cryptoki/src/context/mod.rs
```diff
@@ -32,18 +33,50 @@ use derivative::Derivative;
...
+
+impl Drop for Pkcs11Impl {
+    fn drop(&mut self) {
+        if let Err(e) = self.finalize() {
+            error!("Failed to finalize: {}", e);
+        }
+    }
+}
+
+/// Main PKCS11 context. Should usually be unique per application.
```
ionut-arm, line 74: This should probably be expanded, but let's leave it for a separate PR.

ionut-arm (commented):
Thanks for the patch, a few comments below!

cryptoki/src/session/mod.rs
mjb3279: I think this should be part of a larger discussion.  The original purpose, afaik, for this `_guard` was to force a compile check to prevent a `Session` be used across multiple threads.  Should we maintain that for `Session` or allow a user to, in effect, do whatever they want with the Session, even if it's not in the best "pkcs11" way?

For example I started making changes to Parsec to support these changes but immediately ran into a problem where the PKCS11 provider was unable to be used within an `Arc` because I put `Session` within it to support a `management` session.

cryptoki/src/context/mod.rs
mjb3279: I believe it is safe to make `Pkcs11` clonable because it will clone the instance of `Pkcs11Impl`.  I dont' think you want to make a new `Pkcs11` per slot as that will attempt to reload the PKCS11 C library and call `C_GetFunctionList` again.

cryptoki/src/session/mod.rs
ionut-arm: > Should we maintain that for Session or allow a user to, in effect, do whatever they want with the Session, even if it's not in the best "pkcs11" way?

Isn't that a no-no, based on the spec? As in, multithreaded use of a session _may_ lead to inconsistencies. I was only suggesting to replace the "indirect" way of making `Session` not `Sync` with an explicit one. The broader discussion can be left for another issue.

mjb3279: The spec states that:
>  the behavior of Cryptoki is _undefined_ if multiple threads of an application attempt to access a common Cryptoki session simultaneously

Is it disallowed?  It's highly suggested that you do _not_ use >1 thread to access the same session, but it's not a hard 'no'.

I like that we attempt to protect the user, I think that makes a lot of sense and I'm not against keeping it.  I think user programs (e.g. Parsec) may have a hard time using a management type session without some architecting around it.

mjb3279: Oh so you can't do negative trait bounds yet:

```
error[E0658]: negative trait bounds are not yet fully implemented; use marker types for now
  --> cryptoki\src\session\mod.rs:64:6
   |
64 | impl !Sync for Session {}
   |      ^^^^^
   |
   = note: see issue #68318 <https://github.com/rust-lang/rust/issues/68318> for more information
```

ionut-arm: Damn, ok, thanks for trying! We can just keep it as is for now

cryptoki/src/context/mod.rs
ionut-arm: Hmm, true. I mean `Pkcs11` can be used across threads safely, I assume, but cloning shouldn't be that bad since we'd still have one "actual" context per app.

cryptoki/src/session/mod.rs
vkkoskie: I think you probably want [PhantomData](https://doc.rust-lang.org/nomicon/phantom-data.html) here. Also, you can't implement negative traits (yet), but I think you can test for them to confirm you're getting the effect you want. (Note: I'm not near enough to real code to check that this is even syntactically correct, so it may require some massaging.)

```rust
#[cfg(test)]
mod test {
    use super::MyType;

    fn not_sync<T:!Sync>(_unused: T) {}

    #[test]
    fn my_type_not_sync() {
        // Won't compile if MyType is Sync
        not_sync(MyType);
    }
}
```

ionut-arm (approved):

cryptoki/src/session/mod.rs
ionut-arm: That does look pretty nice. @mjb3279 - would you prefer keeping as is or using `PhantomData`? The test would also be good

mjb3279: I can do that.  Been busy the last few days so haven't had a chance to integrate this idea.  Will try to get to it by the end of the week.

ionut-arm: No rush on my side, I just got back to this.

mjb3279: Looks like you still can't use negative bounds as you described above.

ionut-arm (approved):
👍🏻 thanks!

ionut-arm:
Shall I merge this? I don't think we'll get any other reviews for a while

wiktor-k (approved):
This looks better than what we had previously 👍

PR #60, opened 2021-11-04
Switch to inclusive bindgen naming

vkkoskie:
The bindgen crate has deprecated several methods in favor of more inclusive alternative spellings. Introduces no functional change.


mjb3279:
Just to verify it does not change the output of the generated `.rs` files, correct?

vkkoskie:
Correct. bindgen refers to these as aliases to the items they deprecated. You're already using the most recent version of bindgen, so it doesn't even require a change to Cargo.toml.

mjb3279 (approved):

ionut-arm (approved):
Thanks! I noticed these some time ago but forgot to update/raise an issue

PR #61, opened 2021-11-05
Use rust's own bool type in abstraction crate

vkkoskie:
The Bbool type was translating between the FFI's CK_BBOOL and rust's own bool, by providing a backing intermediate type. But the storage for a bool is guaranteed to be large enough for the job, and the point of conversion from byte back to bool is already structured to check for validity.

### DIFF
--- a/cryptoki/src/object.rs
+++ b/cryptoki/src/object.rs
@@ -762,17 +762,14 @@ impl From<&Attribute> for CK_ATTRIBUTE {
 /// Private function standing in for TryInto<bool> for &[u8]
 /// which can't be implemented through the actual trait because
 /// it and both types are external to this crate.
+/// NB from the specification: "In Cryptoki, the CK_BBOOL data type
+/// is a Boolean type that can be true or false. A zero value means
+/// false, and a nonzero value means true." so there is no invalid
+/// byte value.
 fn try_u8_into_bool(slice: &[u8]) -> Result<bool> {
     let as_array: [u8; std::mem::size_of::<CK_BBOOL>()] = slice.try_into()?;
     let as_byte = CK_BBOOL::from_ne_bytes(as_array);
-    match as_byte {
-        0u8 => Ok(false),
-        1u8 => Ok(true),
-        other => {
-            error!("Byte value {} invalid as bool.", other);
-            Err(Error::InvalidValue)
-        }
-    }
+    Ok(!matches!(as_byte, 0u8))
 }
 
 impl TryFrom<CK_ATTRIBUTE> for Attribute {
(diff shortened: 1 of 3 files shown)


vkkoskie:
Ref for the data layout, which is also present in a comment in the commit: https://doc.rust-lang.org/reference/type-layout.html#primitive-data-layout

cryptoki/src/object.rs
```diff
@@ -617,7 +617,7 @@ impl Attribute {
             | Attribute::Verify(_)
             | Attribute::VerifyRecover(_)
             | Attribute::Wrap(_)
-            | Attribute::WrapWithTrusted(_) => 1,
+            | Attribute::WrapWithTrusted(_) => std::mem::size_of::<bool>(),
```
mjb3279, line 620: Extreme corner case -> `CK_BBOOL` is defined as an `unsigned char`.  What if the architecture defines a `char` to not be 8-bits (e.g. not POSIX compliant).

Googling this question turns up two TI DSPs that if someone implemented a cryptoki token on I'd question their sanity.

I think this is fine, but just wanted to pose the question.

mjb3279:
This looks fine by me. And since Rust's `true` is equivalent to `CK_TRUE` this should work.

mjb3279 (approved):

cryptoki/src/object.rs
vkkoskie: Good catch. It may be possible at an architectural level, but the standard itself excludes that possibility by explicitly saying "unsigned 8-bit value".

At this particular location, the type being sized is exactly the type that the raw pointer points to. So if, say, CK_BBOOL somehow gets bigger because of a weird backing c_uchar, the ptr and len calls still match each other. And the provider should return `CKR_BUFFER_TOO_SMALL` or some other error code.

vkkoskie:
> This looks fine by me. And since Rust's `true` is equivalent to `CK_TRUE` this should work.

Funny enough, there is some ambiguity here. CK_TRUE is defined to 1, but the range of valid types is much more lax:
> In Cryptoki, the CK_BBOOL data type is a Boolean type that can be true or false.  A zero value means 
false, and a nonzero value means true.

Should the range error condition on the u8->bool be removed?

mjb3279:
> In Cryptoki, the CK_BBOOL data type is a Boolean type that can be true or false. A zero value means false, and a nonzero value means true.

That's why I _love_ this spec 🙄 

> Should the range error condition on the u8->bool be removed?

Yeah probably to be as spec-compliant as possible (which I have droned on about in the past).

vkkoskie:
> That's why I _love_ this spec :roll_eyes: 

If it's any consolation, I don't like it either.

ionut-arm (approved):
Makes sense, thanks! 👍🏻

PR #62, opened 2021-11-06
Suppress null pointer deref warnings

vkkoskie:
This is a known issue with bindgen under heavy discussion, but doesn't look like it will be resolved soon. This attribute suppresses genuine warnings about undefined behavior in a pattern generated by bindgen many, many times. It should be removed when the bindgen issue is resolved.

Until then, it obscures other helpful warnings and errors.

See: rust-lang/rust-bindgen#1651

### DIFF
--- a/cryptoki-sys/src/lib.rs
+++ b/cryptoki-sys/src/lib.rs
@@ -8,6 +8,10 @@
 #![allow(clippy::string_lit_as_bytes)]
 // Public items exportedby this crate should match the C style
 #![allow(clippy::upper_case_acronyms)]
+// Suppress warnings from bindgen-generated code
+// Remove on resolution of
+// https://github.com/rust-lang/rust-bindgen/issues/1651
+#![allow(deref_nullptr)]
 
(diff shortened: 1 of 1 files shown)


mjb3279 (approved):

mjb3279:
Thanks for helping clean up the build!  Quickly read through that issue and something tells me that `#![allow(deref_nullptr)]` will be in there for a while...

ionut-arm (approved):
Nice, thanks! Anyone have any idea of ways to get reminded to check on that issue from time to time? It's so easy to forget these TODOs exist.

vkkoskie:
> Nice, thanks! Anyone have any idea of ways to get reminded to check on that issue from time to time? It's so easy to forget these TODOs exist.

I looked but couldn't find any. Within the language itself, every single function has required doc annotations that hint at it being possible. There, a feature is either stable and lists the "since" semver of the release it was stabilized in OR it's unstable and lists the issue number where the feature is being discussed. I couldn't identify a way to use those issue tags in code by mere mortals though.

PR #64, opened 2021-11-11
Add a missing flag value used by TokenInfo

vkkoskie:
This flag indicates a failure in a token self-test. The addition to pkcs11.h currently has no effect beyond documentation.

See: PKCS#11 2.40, Table 6


mjb3279:
Should the bindings be regenerated due to this change?

vkkoskie:
> Should the bindings be regenerated due to this change?

It has no effect on the bindings since bindgen ignores preprocessor definitions. It *should* affect the bindings, but that's a different Issue/PR.

mjb3279 (approved):

mjb3279 (approved):

vkkoskie:
Closing for two reasons:

1) PR #63 had to be rebased onto this one, so it will be included there anyway.
2) PR #66 moots these changes because it replaces the headers entirely.

PR #66, opened 2021-11-14
Replace third party header with OASIS official

### LINKED ISSUE #65: Current `pkcs11.h` is not up-to-date
The `pkcs11.h` header file in `crypto-sys` is not in conformance with [PKCS#11 v2.40 + errata 01](http://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html).  This should be updated to match what the current spec is.  We have two options:

1. Go through the file and add in the missing fields
2. Download latest header files from [oasis](https://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/include/pkcs11-v2.40/)
2a. Will need to either modify `pkcs11.h` or add in our own header files for the different architectures (really the big one is Windows with its `pragma` directives).

vkkoskie:
Not quite ready for merge. Read each commit message, which explain some potential areas of concern.

I was unable to update bindings for two targets, so this will fail normal CI tests until that's corrected in some way.

closes #65
moots #64


vkkoskie:
Hmm. CI is seg faulting. My guess is the struct packing, but I won't have time to dig into this again for a while. One of the things I was scratching my head about while making these changes was how anything was working before with unpacked structs. Updating the packing is technically independent of the header swap, but still needed.

mjb3279:
What prompted the change in 8226a25?  I'm of the opinion we shouldn't touch the headers unless we have to.

cryptoki-sys/oasis-headers/pkcs11.h
```diff
@@ -181,6 +181,12 @@ extern "C" {
  * #endif
  */
 
+/* Recommended Unix definitions from above */
+#define CK_PTR *
+#define CK_DECLARE_FUNCTION(returnType, name) returnType name
+#define CK_DECLARE_FUNCTION_POINTER(returnType, name) returnType (* name)
+#define CK_CALLBACK_FUNCTION(returnType, name) returnType (* name)
+
```
mjb3279, line 189: My recommendation for this is to add a new header file (something like `pkcs11-rust.h`) that defines these rather than doing them in-line.

The following is copied out of a header file I've used to define cryptoki types that has worked on Linux, MacOS and Windows
```C
#pragma once

#ifdef _MSC_VER
#pragma pack(push, cryptoki, 1)
#endif

#if defined _WIN32
#if defined P11_STATIC
#define P11_EXPORT
#elif defined DLL_EXPORT
#define P11_EXPORT    __declspec(dllexport)
#else
#define P11_EXPORT    __declspec(dllimport)
#endif
#else
#define P11_EXPORT
#endif

// Define UNIX defaults for PKCS11 interface

/// 1. CK_PTR: The indirection string for making a pointer to an object.
#define CK_PTR    *

/// 2. CK_DEFINE_FUNCTION(returnType, name): A macro which makes
/// an exportable Cryptoki library function definition out of a
/// return type and a function name.
#define CK_DEFINE_FUNCTION(returnType, name) \
    returnType name

/// 3. CK_DECLARE_FUNCTION(returnType, name): A macro which makes
/// an importable Cryptoki library function declaration out of a
/// return type and a function name.
#define CK_DECLARE_FUNCTION(returnType, name) \
    returnType P11_EXPORT name

/// 4. CK_DECLARE_FUNCTION_POINTER(returnType, name): A macro
/// which makes a Cryptoki API function pointer declaration or
/// function pointer type declaration out of a return type and a
/// function name.
#define CK_DECLARE_FUNCTION_POINTER(returnType, name) \
    returnType P11_EXPORT(*name)

/// 5. CK_CALLBACK_FUNCTION(returnType, name): A macro which makes
/// a function pointer type for an application callback out of
/// a return type for the callback and a name for the callback.
#define CK_CALLBACK_FUNCTION(returnType, name) \
    returnType(*name)

/// 6. NULL_PTR: This macro is the value of a NULL pointer.
#ifndef NULL_PTR
#define NULL_PTR    0
#endif

#include <pkcs11.h>

#ifdef _MSC_VER
#pragma pack(pop, cryptoki)
#endif
```

cryptoki-sys/oasis-headers/pkcs11.h
```diff
@@ -261,4 +264,7 @@ struct CK_FUNCTION_LIST {
 }
 #endif
 
+// Restore prior pack setting
+#pragma pack(pop)
+
```
mjb3279, line 269: I believe explicit struct packing is really only needed for MSC based compilers.

vkkoskie:
> What prompted the change in [8226a25](https://github.com/parallaxsecond/rust-cryptoki/commit/8226a25a78ffcdd8de4dbfae8daf8fc0c6a7a6b6)? I'm of the opinion we shouldn't touch the headers unless we have to.

This was a byproduct of me misinterpreting [this comment](https://github.com/parallaxsecond/rust-cryptoki/blob/f2e7573bfe016e528ddbf9d07d630923f38a3206/cryptoki-sys/src/lib.rs#L98) or maybe just trusting it too much. Much later, when I started messing with the build script, I realized that preprocessor definitions were being excluded because they needed to be explicitly added with `allowlist_var`.

I didn't revert at that time because I liked the fact that const definitions caused the types to be propagated up into the bindings. But Rust's type aliasing really does just make these synonyms with whatever's backing `CK_ULONG` and is really only functioning as a documentation method.

It's fairly weak justification to keep them if having the headers be as pure as possible is a goal. That purity wasn't present in the headers I was replacing, though, so I figured I had a bit more freedom to customize.

> EDIT: If we do decide to change the headers because of bindgen, I believe that a script should be provided that shows how (and why) these changes are made.

Yeah, agreed this wasn't my best commit message.

cryptoki-sys/oasis-headers/pkcs11.h
vkkoskie: +1. I don't have an environment to test for anything like platform-specific preprocessor stuff, so I opted to make it as small and simple as possible. I'll be sure to steal this snippet when I start committing again.

cryptoki-sys/oasis-headers/pkcs11.h
vkkoskie: Not sure what you mean by this exactly. Are you saying that elsewhere, the default packing is what PKCS11 requires?

If so, then I'm not sure what to make of the bindings I generated. I was curious enough about that detail that I split the bindings updates into before- and after-pack commits to create a nice diff.

vkkoskie:
Okay, I basically started over, cherry-picking the first draft of commits as needed. The branch now touches the headers as little as possible. There are still some modifications to look at though.

And again, I've disabled Darwin and Windows support because I don't have those toolchains available, and they'd just fail. Not exactly sure what the best way forward is there.

The only struct packing left is the MSC stuff you provided, which doesn't get used on my end. I'm still very confused about when this does and doesn't matter. But I'm comfortable not making it a condition of merging because it wasn't present before and so you're no worse off keeping it as is.

vkkoskie:
> EDIT: If we _do_ decide to change the headers because of `bindgen`, I believe that a script should be provided that shows how (and why) these changes are made.

I've now managed to get all but one of the changes needed from the base header implemented in build.rs at generation time. The last one is a disagreement that can crop up between bindgen's preprocessor and compiler passes that I don't see a clean way around. I think it's minor enough to leave as is though.

ionut-arm:
Hey,

Finally got some time to look through this.

1) could you please expand the comment in https://github.com/parallaxsecond/rust-cryptoki/pull/66/commits/eff891738d3c669030ec7d8a09a01bea1d2d0191 to link to the place where those header files were taken from?
2) @mjb3279 - after this is merged could you regenerate the Windows bindings? I'll try to do the same for Mac OS

vkkoskie:
> Hey,
> 
> Finally got some time to look through this.
> 
>     1. could you please expand the comment in [eff8917](https://github.com/parallaxsecond/rust-cryptoki/commit/eff891738d3c669030ec7d8a09a01bea1d2d0191) to link to the place where those header files were taken from?

Sure. Just update the commit message or drop a README in the oasis-headers directory? Or both? Also, is the licensing information at the top of the headers sufficient? I'm trying to parse their [legal pages](https://www.oasis-open.org/policies-guidelines) but am having trouble finding anything directed at derivative products instead of OASIS official notices.

>     2. @mjb3279 - after this is merged could you regenerate the Windows bindings? I'll try to do the same for Mac OS

If you place these in any commit I have access to, I'll be able to cherry-pick them into this PR pre-merge.

>     3. `One of the things I was scratching my head about while making these changes was how anything was working before with unpacked structs.` - what do you mean? I thought `bindgen` takes care of this (or maybe I'm thinking of the wrong thing)

It's the difference between `#[repr(C)]` and `#[repr(C,packed)]` in the generated bindings. It's been rebased out of the current PR, but if you look at 60a3a3f you can see the difference between the two. In particular, struct sizes shrink in many places. It also makes the tests seg fault. But the packed representation is an explicit feature of the standard because that's the only one with any guarantee of ABI consistency.

There's definitely more investigating to do, and I haven't looked at it much myself.  But it can be treated separately from this PR, and should probably have its own tracking issue.

ionut-arm:
Sorry for the very sluggish replies.

> 
> Sure. Just update the commit message or drop a README in the oasis-headers directory? Or both? Also, is the licensing information at the top of the headers sufficient? I'm trying to parse their [legal pages](https://www.oasis-open.org/policies-guidelines) but am having trouble finding anything directed at derivative products instead of OASIS official notices.
> 

Just the commit message should be fine. As for the license... hmm, good point, I'll have to find out.

> 
> If you place these in any commit I have access to, I'll be able to cherry-pick them into this PR pre-merge.
> 

Should be ok to just do it in separate PRs, no need to complicate this one.

> 
> It's the difference between `#[repr(C)]` and `#[repr(C,packed)]` in the generated bindings. It's been rebased out of the current PR, but if you look at [60a3a3f](https://github.com/parallaxsecond/rust-cryptoki/commit/60a3a3f7213dc983117a805ce0c8699217cff4ea) you can see the difference between the two. In particular, struct sizes shrink in many places. It also makes the tests seg fault. But the packed representation is an explicit feature of the standard because that's the only one with any guarantee of ABI consistency.
> 
> There's definitely more investigating to do, and I haven't looked at it much myself. But it can be treated separately from this PR, and should probably have its own tracking issue.

Ah, that makes sense. Ok, a new issue it is - should I create it? I think you have the details lined up more finely.

I'll have another read through the PR, though I need to figure out if there are any restrictions/barriers for us to include those header files. Maybe we can just pull them at build time if need be?

ionut-arm:
Hi @vkkoskie !

I'm not sure if you managed to find more information on the issue regarding Oasis licenses.

I've queried internally and it seems the headers are indeed under Oasis IPR, however the IPR is incompatible with our Apache-2.0 license (as it requires limiting distribution to members of Oasis only), so the PR can't be merged in its current form. However, I still don't know if pulling them in at build time to generate the FFI code is any better. Do you happen to have any more insight into that?

The Oasis headers appear to be more of a spec in code form. At the same time, I don't know which other headers we should be using for our goals...

vkkoskie:
Interesting. I followed a few links down from yours to [this page](https://www.oasis-open.org/open-repositories/#licensingRules) that seems to at least acknowledge that someone might actually want to _use_ OASIS work product instead of just contribute to it.

That led me to search github to see of a repo existed, which [does exist](https://github.com/oasis-tcs/pkcs11). (What's currently committed came from from OASIS's own domain.) The license there is boilerplate (as noted in its commit message) and links to other places in the OASIS legal docs that are also beyond the limits of my legal knowledge. My best guess is that the headers can be used under [BSD-3-Clause](https://opensource.org/licenses/BSD-3-Clause), but I"ve filed an issue on that repo to get clarification.

ionut-arm:
Thanks for following that up, it sounds promising! I'll subscribe to the issue we raise, let's see what they reply :)

mjb3279:
Hey everyone.  Sorry for disappearing as of late.  I normally get to work on this repo when there is a lull in work, however lately I've been absolutely swamped with work and family stuff.  Hopefully early next year things will calm down and I can get back to contributing.

ionut-arm:
> Hey everyone. Sorry for disappearing as of late. I normally get to work on this repo when there is a lull in work, however lately I've been absolutely swamped with work and family stuff. Hopefully early next year things will calm down and I can get back to contributing.

I'm in pretty much the same situation, hope we're not slowing @vkkoskie down too much 😅

jhagborgftx:
It seems that structure packing should be limited to Windows.  This is a known inaccuracy of the spec, and there is a [proposal](https://www.oasis-open.org/committees/document.php?document_id=69505&wg_abbrev=pkcs11) to change this in v3.2.  This was accepted without objection according to these [meeting minutes](https://wiki.oasis-open.org/pkcs11/MeetingMinutes/Minutes01262022).

Even in older versions of the spec, it seems nobody was actually packing structures on Unix.  So following the spec literally will cause ABI incompatibility with every existing PKCS#11 implementation.

wiktor-k:
I propose we close this PR. It doesn't see much progress lately and there are parallel efforts to update the PKCS version in other PRs/issues. Also closing doesn't hurt since the code is still here and in case something happens we can re-open.

Thoughts?

ionut-arm:
> I propose we close this PR. It doesn't see much progress lately and there are parallel efforts to update the PKCS version in other PRs/issues. Also closing doesn't hurt since the code is still here and in case something happens we can re-open.
> 
> Thoughts?

Agree, it's gone somewhat stale, I'd be happy to close. If we need to revisit the topic, we can re-open or link to this one from a new PR.

PR #67, opened 2021-11-16
Revert "Draft: SlotInfo refactor and reimplement flags types"

ionut-arm:
Apologies for the confusion, it's been quite hectic lately. Does this work as a way to backtrack or should I nuke the commits completely?

cc @vkkoskie


vkkoskie:
I have no objections to resetting any branch instead of piling revert commits on top. I've got local versions of everything that I can reconstruct with. If you do that and it's possible, I'd also recommend reopening #63 so the comment history etc. can continue to be built on.

ionut-arm:
> I'd also recommend reopening #63 so the comment history etc. can continue to be built on.

Done, but doesn't look possible to reopen a merged PR. Will have to do this the painful way, by creating a new PR and link to the old one.

PR #68, opened 2021-11-17
Info flags refactor

vkkoskie:
Reopens #63 which was closed by an accidental merge (now reversed).

I renamed the branch to reflect its intent a little better, but made no content changes beyond some minor typos.

Threads that may still be live/relevant at the time of opening form prior PR:
* [Rough outline of future work](https://github.com/parallaxsecond/rust-cryptoki/pull/63#issuecomment-967228360)
* [Discussion](https://github.com/parallaxsecond/rust-cryptoki/pull/63#discussion_r749875856) on a choice of return type
* ~[Tangent about Date type](https://github.com/parallaxsecond/rust-cryptoki/pull/63#discussion_r750825650) which doesn't affect this PR, but may still be relevant generally.~ EDIT: I've just added a `UtcTime` type which is much simpler than the one I'd originally conceived.


cryptoki/src/flag.rs
```diff
@@ -0,0 +1,319 @@
...
+#[derive(Default)]
+#[repr(transparent)]
+pub(crate) struct CkFlags<T> {
+    value: CK_FLAGS,
+    _type: PhantomData<T>,
+}
+
+// Must hand write Clone and Copy traits because the derived
+// versions assume T to be Copy and constrain the impl accordingly.
+// In general, we'll have non-Copy Ts and want to avoid this.
```
wiktor-k, line 75: Great comment because I was just wondering about this :)

cryptoki/src/flag.rs
```diff
@@ -0,0 +1,319 @@
...
+/// A type representing a 1-bit flag constant or some composition thereof
+#[repr(transparent)]
+pub(crate) struct FlagBit<T> {
+    value: CK_FLAGS,
+    _type: PhantomData<T>,
+}
+
+/// A type representing the one's compliment of a FlagBit<T>
+#[repr(transparent)]
+pub(crate) struct FlagMask<T> {
```
wiktor-k, line 60: This is a very nice implementation of bit fields & masks. This is mostly generic - I was wondering - isn't there a crate the we could reuse for this purpose?

cryptoki/src/slot/slot_info.rs
```diff
@@ -0,0 +1,159 @@
...
+        self.hardware_version
+    }
+
+    /// Version number of the slot's firmware
+    pub fn firmware_version(&self) -> Version {
+        self.firmware_version
+    }
+}
+
+#[doc(hidden)]
```
wiktor-k, line 87: Why `#[doc(hidden)]`?

cryptoki/src/types.rs
```diff
@@ -208,3 +252,110 @@ impl From<CK_VERSION> for Version {
         }
     }
 }
+
+/// A UTC datetime returned by a token's clock if present.
+#[derive(Copy, Clone, Debug)]
+pub struct UtcTime {
+    /// **[Conformance](crate#conformance-notes):**
+    /// Guaranteed to be in range 0..=9999
+    pub year: u16,
```
wiktor-k, line 261: :thinking: Not sure if it's correct to say it's guaranteed to be in range if it's `pub`. I guess what you mean is that the PKCS#11 lib will not return other values but I'm not sure how to approach this. Make this struct immutable and only expose fields via impl functions?

wiktor-k (commented):
This looks nice :+1:

cryptoki/src/flag.rs
vkkoskie: Yes! There is a [bitflags](https://docs.rs/bitflags/latest/bitflags/) crate, which I may actually switch over to before calling this done. The reason I avoided it initially, is I thought these types would be public, and so wanted them bound to their containing struct to avoid any kind of confusion or misuse allowed by the public API. But once I got to the struct itself, I found that they could easily be private. So don't be surprised if I do end up doing exactly that.

cryptoki/src/slot/slot_info.rs
vkkoskie: I think it's best to avoid any direct linkage between this crate and the sys crate, and this is one aspect of that. The reasoning is to avoid implying that a user application should mix the two or need to dip into the ffi for any reason.

What I'd like to do is have this trait be private, but the visibility rules for traits don't allow that. So hiding the docs is a weaker version of that.

cryptoki/src/types.rs
vkkoskie: Yes. I meant the original format for this type is a character string with exactly 4 digits for this field. If it doesn't land in that range, the conversion will fail and generate an error, never constructing the type at all. I'm definitely open to alternatives. Did you happen to see my [comment](5dfddf8) on this? Because I definitely considered several.

wiktor-k: Ha, no I haven't. Now that I've read them I understand the problem better. FTR I'd still use getters just in case the struct will be used as an input one day. We'd avoid the API breakage. One more thing would be additional `to_iso_string` function. I know Display does that but from the API discoverability perspective when someone wants the ISO time they'd be sure `to_iso_string` would do that while for Display they'd have to read the docs and still be unsure if the format won't change.

Still, cool work. I'll be watching this PR for further changes. 👍

cryptoki/src/slot/slot_info.rs
wiktor-k: Nice idea. Maybe we should apply it more broadly. OTOH if someone knows the low-level type they may be surprised the conversion is not listed in the docs :thinking:

@ionut-arm what do you think about it?

vkkoskie: > apply it more broadly

This was my intent. I've started a Discussion page (#69) that might be a better place for topics like this so they don't disappear when the PR closes. This item would fall under the second-to-last bullet in the "as I go" list. No one seems to have commented on (noticed?) that page yet, but it has plenty of stuff that's debatable or subjective.

One of my hopes in putting it up was to sanity check any ideas hat might raise objections well before sinking any real time into them. So if there is disagreement, there might be a good place to move them.

cryptoki/src/types.rs
vkkoskie: > FTR I'd still use getters just in case the struct will be used as an input one day. We'd avoid the API breakage.

I can do this, but how would you feel about making the fields pub and [read-only](https://docs.rs/readonly/latest/readonly/index.html) instead of getters? I'm concerned a bit about callback bloat in ancillary types. This one in particular doesn't seem to have changed or found new use in the 3.x drafts that will eventually come into play, and I don't think _any_ kind of change is being made to the 2.x standard any more.

> One more thing would be additional to_iso_string function. I know Display does that but from the API discoverability perspective when someone wants the ISO time they'd be sure to_iso_string would do that while for Display they'd have to read the docs and still be unsure if the format won't change.

Glad to hear another opinion on "How much ISO (if any) is appropriate?" I like the suggestion of having a named function instead. Would you recommend removing Display altogether then?

vkkoskie:
@ionut-arm @wiktor-k @mjb3279

I believe this is ready for review and/or approval. There may be some unfinished threads (here or #63) but I'm done making new additions.

vkkoskie:
> I'm done making new additions.

To clarify, I'm done on this PR. There are actually some places I resisted the urge to edit beyond the scope of this PR, which is already fairly large. I think this is a good checkpoint to merge on and then start those changes on a fresh branch.

cryptoki/src/types.rs
wiktor-k: > I can do this, but how would you feel about making the fields pub and read-only instead of getters?

This is 👌 too. 

> I'm concerned a bit about callback bloat in ancillary types.

Do you mean the overhead of function calls for simple getters? I guess they should be inlined anyway, either by default or with a [little bit of help](https://nnethercote.github.io/perf-book/inlining.html). 

> Glad to hear another opinion on "How much ISO (if any) is appropriate?" I like the suggestion of having a named function instead. Would you recommend removing Display altogether then?

I don't mind the Display but it seems the docs indicate it's for human consumption not any stable format. I see you've removed it. That's OK too. If someone needs it it can be added later.

wiktor-k:
> To clarify, I'm done on this PR. 

This looks very nice now. Thanks 👍 

> There are actually some places I resisted the urge to edit beyond the scope of this PR, which is already fairly large. I think this is a good checkpoint to merge on and then start those changes on a fresh branch.

I agree. The PR is quite sizeable at this point and already takes a moment to digest. For me personally the end of the year is always particularly hard so sorry in advance for the slow process here but I hope we can manage to get this merged sooner rather than later 😊

ionut-arm:
Hey! I'm starting to review this now, should be done by the end of the day, sorry for the delay! I think the (generic) advice here would be to try and split PRs into smaller chunks. I know that's kinda what the commits should help with, but it's somewhat difficult to review commit-by-commit, for a number of reasons...

ionut-arm (commented):
Ok, went through all of it. Left a few comments, but happy to approve otherwise.

Also, writing some (generic) comments here as I go through the commits as well:
- what’s the worry with stability around `Display`and `Debug`? Might’ve missed it from a different conversation. I don’t think anyone should be using `Debug` and/or `Display` in a way that requires stability, they’re just intended for printing out various values in developer/user friendly ways, it doesn’t come with any guarantees - as in, it’s not supposed to be a proper serialization mechanism.
- to reiterate the point from the previous comment I left - please split out PRs into smaller chunks, it makes them a lot easier to review and get merged. the UtcTime part, for example, could be a separate PR. Having gone through all of it now, this could’ve easily been split into 4-5 separate PRs.

cryptoki/src/slot/mod.rs
```diff
@@ -163,121 +162,356 @@ impl From<CK_SLOT_INFO> for SlotInfo {
...
-#[derive(Debug, Clone, Copy, Default)]
+#[derive(Debug, Clone)]
 pub struct TokenInfo {
-    val: CK_TOKEN_INFO,
+    label: String,           // 32
+    manufacturer_id: String, // 32
+    model: String,           // 16
+    serial_number: String,   // 16
+    flags: CkFlags<Self>,
+    max_session_count: Option<Option<u64>>,
```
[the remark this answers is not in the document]
vkkoskie: Yeah, I don't like either the way I did it or what you're suggesting. But you're the second person to express this preference, so I guess I should take the hint. :) I think I just don't much like creating uni-tasker new types when I know that this crate in particular is going to have loads of them when mechanisms start filling out.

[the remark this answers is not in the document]
vkkoskie: See f18c73e.

cryptoki/src/types.rs
```diff
@@ -251,3 +252,117 @@ impl From<CK_VERSION> for Version {
...
+
+    #[test]
+    fn utc_time_convert_bad() {
+        let invalid: [u8; 16] = [
+            0x41, 0x39, 0x37, 0x30, 0x30, 0x31, 0x30, 0x31, 0x30, 0x30, 0x30, 0x30, 0x30, 0x30,
+            0x30, 0x30,
+        ];
+        let invalid = convert_utc_time(invalid);
+        assert!(invalid.is_none());
+    }
```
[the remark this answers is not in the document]
vkkoskie: See 40810ee.

cryptoki/src/types.rs
```diff
@@ -289,24 +289,17 @@ impl fmt::Display for UtcTime {
...
 
 // UTC time has the format YYYYMMDDhhmmss00 as ASCII digits
-pub(crate) fn convert_utc_time(orig: [u8; 16]) -> Option<UtcTime> {
-    // skip check of reserved padding chars
-    for c in orig[0..14].iter() {
-        if !c.is_ascii_digit() {
-            return None;
-        }
-    }
+pub(crate) fn convert_utc_time(orig: [u8; 16]) -> Result<Option<UtcTime>> {
```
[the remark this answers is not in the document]
vkkoskie: Good catch. The getter on the token can return an Option if the device doesn't have a clock. But that shouldn't have been expressed in this converter function.

See 7a0ad30.

vkkoskie:
>     * what’s the worry with stability around `Display`and `Debug`? Might’ve missed it from a different conversation. I don’t think anyone should be using `Debug` and/or `Display` in a way that requires stability, they’re just intended for printing out various values in developer/user friendly ways, it doesn’t come with any guarantees - as in, it’s not supposed to be a proper serialization mechanism.

Agree, but I try to program with [Hyrum's Law](https://www.hyrumslaw.com) front-of-mind. I wanted to provide something more useful than dumping a raw integer, but I also didn't want to invent a new, mostly arbitrary formatting scheme that would have to be remembered anywhere else that this was necessary. Using the existing tools from the language seemed like the best way to do this, and anything "extra" felt like it would create more problems than it would solve.

>     * to reiterate the point from the previous comment I left - please split out PRs into smaller chunks, it makes them a lot easier to review and get merged. the UtcTime part, for example, could be a separate PR. Having gone through all of it now, this could’ve easily been split into 4-5 separate PRs.

Yeah, sorry about that. I primarily experience git as a CLI tool and tend to use a graph view to understand change sets. So good individual commits can be reviewed in sequence, and the size of an MR doesn't usually matter. I'll try to keep in mind not everyone uses the same tooling.

I think one thing I dislike about many small PRs is the history it (often) creates. To my knowledge, github doesn't yet support a [semi-linear merge strategy](https://github.community/t/feature-request-semi-linear-history-merging-strategy/2325). This is why you see me constantly rebasing within my PRs, and why I don't like to have more than one or two submitted at a time. If they all get merged at once from an assortment of common ancestors, it creates a rat's nest that future-me will hate to have to mentally untangle.

cryptoki/src/types.rs
```diff
@@ -218,6 +218,11 @@ impl std::fmt::Display for Version {
 }
 
 impl Version {
+    /// Construct a new version
+    pub fn new(major: u8, minor: u8) -> Self {
+        Self { major, minor }
+    }
```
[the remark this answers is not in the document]
vkkoskie: Yep. See aae8de0.

ionut-arm:
> So good individual commits can be reviewed in sequence, and the size of an MR doesn't usually matter.

I think the problem with that is how the Github interface for this works (or doesn't work) - you don't get the same controls when reviewing one commit at a time. And another issue (for me, at least) is that, like in this PR, some commits "overwrite" (or build on) changes made by a previous one, so you need to have the whole PR in mind when trying to understand what's going on - that's the kind of situation where the size ends up mattering. Obviously, if every commit is entirely independent of the others then it's not a problem at all.

cryptoki/src/slot/slot_info.rs
ionut-arm: I'm not really sure about this - as in, not sure it's a (big) benefit. Don't think it's something that would bother or distract the average user, and it might actually be helpful if they go ahead an use the FFI for something that's not currently implemented in the abstracted crate, and then want to convert to use the result with the abstracted stuff. But then again, I don't think that would be the average user either. I'm happy with this going either way, though, I guess

ionut-arm (approved):
Based on the changes after my review: 👍🏻

cryptoki/src/slot/slot_info.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// This string is maximally 32 bytes (*not* chars) as UTF-8
+    pub fn manufacturer_id(&self) -> &str {
+        &self.manufacturer_id
+    }
+
+    /// True if a token is in the slot (e.g., a device is in the reader).
+    ///
+    /// **[Conformance](crate#conformance-notes):**
+    /// If this slot does not represent a removable device, a token is *always*
+    /// considered to be present. That is, `slot.removable device() == false`
```
wiktor-k, line 50: Missing `_` in `slot.removable device()`?

cryptoki/src/slot/token_info.rs
```diff
@@ -0,0 +1,553 @@
...
+    /// The token does not provide information about this limit
+    Unavailable,
+    /// The limit is "effectively infinite" and may be treated as such
+    Infinite,
+}
+
+/// Information about a token
+#[derive(Debug, Clone)]
+pub struct TokenInfo {
+    label: String,           // 32
```
wiktor-k, line 53: It'd be nice to expand the comment to say sth like "Up to/exactly 32 characters, space-padded".

cryptoki/src/slot/token_info.rs
```diff
@@ -0,0 +1,553 @@
...
+    }
+}
+
+impl MaybeUnavailable for u64 {
+    fn maybe_unavailable(value: CK_ULONG) -> Option<u64> {
+        if value == CK_UNAVAILABLE_INFORMATION {
+            None
+        } else {
+            // Must have cast for when ulong is 32 bits
+            // Must have lint suppression when ulong is 64 bits
```
wiktor-k, line 92: Very nice comment. Also: it'd be nice if we had CI running 32-bit builds too :)

cryptoki/src/context/mod.rs
```diff
@@ -158,7 +157,7 @@ impl Pkcs11 {
     }
 
     /// Open a new session with no callback set
-    pub fn open_session_no_callback(&self, slot_id: Slot, flags: SessionFlags) -> Result<Session> {
-        session_management::open_session_no_callback(self, slot_id, flags)
+    pub fn open_session_no_callback(&self, slot_id: Slot, read_write: bool) -> Result<Session> {
```
wiktor-k, line 160: :thinking: I'm not really too much into having boolean parameters since it's not obvious what `open_session_no_callback(..., true)` does (an enum with `READ_WRITE` or `READ_ONLY` would be more... readable) but on the other hand it's commonly used and the IDE can offer hints.

Or even export two functions: `open_readonly_session_no_callback` or `open_readwrite_session_no_callback`...

Btw very nice to see you providing explicit reasoning in the commit message for SessionFlags refactor: https://github.com/parallaxsecond/rust-cryptoki/pull/68/commits/67e758928631c68f3ed4db601df8e58dd95ad50a :clap: thanks!

wiktor-k (approved):
Looks great, thanks for being so patient with the review process @vkkoskie! :)

I added some minor things but I think this is already good to merge as is :+1:

vkkoskie:
One more commit to fix the typos mentioned above. Anyone mind if I do one final rebase onto 0.3.0 before it gets merged?

ionut-arm:
> One more commit to fix the typos mentioned above. Anyone mind if I do one final rebase onto 0.3.0 before it gets merged? (Or, if you're merging, mind rebasing when you do?)

We don't have `Rebase and merge` enabled, but could enable it - just need to have a conversation about it for this repo in particular. In the meantime you can try and do it by hand if you don't mind too much. I hope things didn't change around too much!

wiktor-k:
I don't mind it either way. Before working on these Rust crates I would rebase everything religiously but here the simple merge strategy also works well (modulo when CI is green on the top commit and red when merged, hehe).

For the record I'm *for* enabling additional merge options in this repo. It doesn't hurt :)

ionut-arm:
> For the record I'm for enabling additional merge options in this repo. It doesn't hurt :)

I'm _for_ as well. We can discuss in the [community meeting](https://github.com/parallaxsecond/community#community-meeting) starting in a few minutes. @vkkoskie - if you'd like to join you're more than welcome! It's a more generic, Parsec-related meeting, but we do discuss issues related to all the repos we host.

wiktor-k:
Don't want to interrupt you @ionut-arm but if you can please enable and "Rebase and merge" it when you have time. Thanks!

ionut-arm:
> Don't want to interrupt you @ionut-arm but if you can please enable and "Rebase and merge" it when you have time. Thanks!

Yes, was about to confirm that we've agreed to enable that. It should be available now!

Issue #72, opened 2021-12-06
`pkcs11.open_session_no_callback` against Luna Network HSM crashed with SIGSEGV


hansonchar:
It doesn’t seem `rust-cryptoki` can be used to open a pkcs11 session to the Thales eLab, i.e. Luna Network HSM without ending up in `(signal: 11, SIGSEGV: invalid memory reference)`.

## Sample unit test:

```rust

extern crate cryptoki;

use cryptoki::Pkcs11;
use cryptoki::types::slot_token::Slot;
use cryptoki::types::locking::CInitializeArgs;
use cryptoki::types::Flags;
use std::env;

fn open_session() -> (Pkcs11, Slot) {
    let pkcs11 = Pkcs11::new(
        env::var("PKCS11_HSM_MODULE")
            .unwrap_or_else(|_| "/usr/safenet/lunaclient/lib/libCryptoki2_64.so".to_string()),
    )
    .unwrap();

    // initialize the library
    pkcs11.initialize(CInitializeArgs::OsThreads).unwrap();

    // find a slot, get the first one
    let slot = pkcs11.get_slots_with_token().unwrap().remove(0);

    println!("slot: {}", slot.id());

    // set flags
    let mut flags = Flags::new();
    let _ = flags.set_rw_session(true).set_serial_session(true);

    {
        pkcs11.open_session_no_callback(slot, flags).unwrap();
    }

    (pkcs11, slot)
}

#[test]
#[serial]
fn test_open_session() {
    open_session();
}
```

## Test run and failure

```bash
# Centos 7
$ cargo test -- --nocapture
running 1 test
slot: 0
error: test failed, to rerun pass '--bin xks-proxy'

Caused by:
  process didn't exit successfully: `/local/centos/ThalesElab/rust/target/debug/deps/my_project-be90c88b811a4002 --nocapture` (signal: 11, SIGSEGV: invalid memory reference)
```

Note
1. I can open the session in pure C without issue.
2. Switching between `rust` stable vs nightly build doesn't seem to make any difference.

Also, changing the `libloading` dependency to open the `.so` file from:
```rust
Library::open(Some(filename), RTLD_LAZY | RTLD_LOCAL)
```
to:
```rust
Library::open(Some(filename), RTLD_NOW)
```
doesn't seem to make any difference.

ionut-arm:
Hi,

[...]

PR #73, opened 2021-12-08
Fix SIGSEGV upon pkcs11.open_session_no_callbac

hansonchar:
https://github.com/parallaxsecond/rust-cryptoki/issues/72

Signed-off-by: Hanson Char <hanson.char@gmail.com>


hug-dev:
Hey @hansonchar!!

Thanks for the detailed bug report and the PR. As this file is automatically generated from `bindgen`, I guess that a long term solution would be to fix the code-generation side directly, so that it's fixed for everyone? I am afraid that we will overwrite your fix if we re-generate the code.

Is there any issue about that on the `bindgen` repo? Might be worth to discuss with them too. Also @joechrisellis added the `libloading` feature in `bindgen` so he might be able to give pointers!

hug-dev:
I'll let the others share their thoughts though! Maybe it does not hurt to include that temporary as long as the new bindings also contain it!

ionut-arm:
As per my message in the issue - an approach that could help in the future with other problems as well could be to place those `libloading` calls from this PR in a separate method implemented in [`lib.rs`](https://github.com/parallaxsecond/rust-cryptoki/blob/main/cryptoki-sys/src/lib.rs), but allowing the user to decide which flags they want to pass down. Thoughts?

hansonchar:
Great idea.

PR #81, opened 2022-01-18
Add SHAn-RSA-PKCS-PSS mechanisms

ionut-arm:


cryptoki/src/mechanism/rsa.rs
```diff
@@ -135,12 +135,6 @@ pub struct PkcsPssParams {
     pub s_len: Ulong,
 }
 
-impl From<PkcsPssParams> for Mechanism {
-    fn from(pkcs_pss_params: PkcsPssParams) -> Self {
-        Mechanism::RsaPkcsPss(pkcs_pss_params)
-    }
-}
-
```
hug-dev, line 143: Good point!

hug-dev (approved):
:ok_hand:

cryptoki/src/mechanism/mod.rs
```diff
@@ -108,6 +108,20 @@ impl MechanismType {
...
+        val: CKM_SHA256_RSA_PKCS_PSS,
+    };
+    /// SHA384-RSA-PKCS-PSS mechanism
+    pub const SHA384_RSA_PKCS_PSS: MechanismType = MechanismType {
+        val: CKM_SHA384_RSA_PKCS_PSS,
+    };
+    /// SHA512-RSA-PKCS-PSS mechanism
+    pub const SHA512_RSA_PKCS_PSS: MechanismType = MechanismType {
+        val: CKM_SHA512_RSA_PKCS_PSS,
+    };
```
beamer159, line 123: Could you also add `SHA1_RSA_PKCS_PSS` for completeness?

ionut-arm: Done!

cryptoki/src/mechanism/rsa.rs
wiktor-k: This breaks the API... fortunately we're on 0.x and the fix for clients would be that difficult.

wiktor-k (approved):
This looks simple and straightforward :+1:

cryptoki/src/mechanism/rsa.rs
ionut-arm: True, though once you upgrade this conversion simply wouldn't be possible. And I think #68 is a much bigger issue for anyone upgrading :grin:

wiktor-k: Yes, let's use #68 as an opportunity to clean up stuff and having this From would be super-extremely confusing :)

PR #84, opened 2022-02-09
Solve open issues

### LINKED ISSUE #74: Segmentation fault on parsing `Date`
When requesting `StartDate` or `EndDate` attributes on a key we get a segmentation fault. For example, running the following unit test:
```
#[test]
#[serial]
fn aes_key() -> Result<()> {
    let (pkcs11, slot) = init_pins();

    // set flags
    let mut flags = SessionFlags::new();
    let _ = flags.set_rw_session(true).set_serial_session(true);

    // open a session
    let session = pkcs11.open_session_no_callback(slot, flags)?;

    // log in the session
    session.login(UserType::User, Some(USER_PIN))?;

    // get mechanism
    let mechanism = Mechanism::AesKeyGen;

    // pub key template
    let key_template = vec![
        Attribute::Class(ObjectClass::SECRET_KEY),
        Attribute::Token(true),
        Attribute::Sensitive(true),
        Attribute::ValueLen(16.into()),
        Attribute::KeyType(KeyType::AES),
        Attribute::Label(b"testAES".to_vec()),
        Attribute::Private(true),
    ];

    // generate a key pair
    let key = session.generate_key(&mechanism, &key_template)?;

    let attributes_result =
        session.get_attributes(key, &[AttributeType::EndDate, AttributeType::StartDate]);

    match attributes_result {
        Ok(attributes) => println!("working with version: {:?}", attributes),
        Err(e) => println!("error getting attributes: {:?}", e),
    }
    Ok(())
}
```

results (for me) in a SIGSEGV.

### LINKED ISSUE #77: Add `is_initialized()` to `Pkcs11`
I recently created [a PR for the Krill project](https://github.com/NLnetLabs/krill/pull/754) to switch from the `pkcs11` crate to the `cryptoki` crate and in doing so I discovered a missing piece of functionality: the `cryptoki` crate has no equivalent of the `pkcs11` crate [`is_initialized()](https://docs.rs/pkcs11/latest/pkcs11/struct.Ctx.html#method.is_initialized) function.

Of course I can track whether or not I have called initialized myself but that's extra work. Or alternatively add an `fn initialize_if_needed()` or enable `fn initalize()` to tolerate being already initialized, which would prevent the caller from having to check and then call initialize to avoid getting an already initialized error.

### LINKED ISSUE #78: How to test for supported functions?
I recently created [a PR for the Krill project](https://github.com/NLnetLabs/krill/pull/754) to switch from the `pkcs11` crate to the `cryptoki` crate and in doing so I discovered a difference: the `pkcs11` crate requires lots of the pointers returned by `C_GetFunctionList()` to be non-null when the library is loaded, while `cryptoki` delays validation of function pointers until they are invoked.

If I want to verify as early as possible, and without actually invoking the functions, which functions a loaded library supports, how can I do that with `cryptoki`?

ionut-arm:
Fixing a few of the open issues:

* Fixes #74 - changing the way we handle `Date` parsing to account for empty values
* Closes #77 - add a method to check whether the context was initialized
* Closes #78 - add a method to check for support of specific functions

### DIFF
--- a/cryptoki/src/context/general_purpose.rs
+++ b/cryptoki/src/context/general_purpose.rs
@@ -10,7 +10,7 @@ use std::convert::TryFrom;
 
 // See public docs on stub in parent mod.rs
 #[inline(always)]
-pub(super) fn initialize(ctx: &Pkcs11, init_args: CInitializeArgs) -> Result<()> {
+pub(super) fn initialize(ctx: &mut Pkcs11, init_args: CInitializeArgs) -> Result<()> {
     // if no args are specified, library expects NULL
     let mut init_args = CK_C_INITIALIZE_ARGS::from(init_args);
     let init_args_ptr = &mut init_args;
@@ -19,6 +19,9 @@ pub(super) fn initialize(ctx: &Pkcs11, init_args: CInitializeArgs) -> Result<()>
             init_args_ptr as *mut CK_C_INITIALIZE_ARGS as *mut ::std::ffi::c_void,
         ))
         .into_result()
+        .map(|_| {
+            ctx.initialized = true;
+        })
     }
 }
 
--- a/cryptoki/src/context/general_purpose.rs
+++ b/cryptoki/src/context/general_purpose.rs
@@ -5,6 +5,7 @@
 use crate::context::{CInitializeArgs, Info, Pkcs11};
 use crate::error::{Result, Rv};
 use cryptoki_sys::{CK_C_INITIALIZE_ARGS, CK_INFO};
+use paste::paste;
 use std::convert::TryFrom;
 
 // See public docs on stub in parent mod.rs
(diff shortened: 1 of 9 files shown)


cryptoki/src/types.rs
```diff
@@ -53,6 +53,29 @@ impl Date {
         let date = CK_DATE { year, month, day };
         Self { date }
     }
+
+    /// Creates a new, empty `Date` structure
+    ///
+    /// This represents the default value of the attribute (on
+    /// newer implementations of `Cryptoki`).
+    pub fn new_empty() -> Self {
+        let date = CK_DATE {
```
wiktor-k, line 62: It seems (from my local git grep) `CK_DATE` implements Default so maybe this can be shortened to:

```rust
Self { date: Default::default() }
```

?

cryptoki/src/context/general_purpose.rs
```diff
@@ -30,3 +34,162 @@ pub(super) fn get_library_info(ctx: &Pkcs11) -> Result<Info> {
...
+    GenerateRandom,
+    GetFunctionStatus,
+    CancelFunction,
+    WaitForSlotEvent,
+}
+
+#[inline(always)]
+pub(super) fn is_fn_supported(ctx: &Pkcs11, function: Function) -> bool {
+    match function {
+        Function::Initialize => check_fn!(ctx, Initialize),
```
wiktor-k, line 126: Hehe, very clever :+1:

wiktor-k (approved):
Looks ok :+1:

ximon18:
Thanks, I'll take a look tomorrow!

cryptoki/src/types.rs
wiktor-k: I meant the entire `date` field could be constructed using `Default`:

```rust
    pub fn new_empty() -> Self {
        Self { date: Default::default() } // or: CK_DATE::default()
    }
```

or even if `Date` implemented `Default` (it could since there is only one field that allegedly impls `Default`) we could have:

```rust
    pub fn new_empty() -> Self {
        Self::default()
    }
```

Please decide which one of all these options here reads better :) (this also reminds of of the law of reviews that smaller ones get a lot of nitpicking while enormous PRs are always "LGTM" haha).

ionut-arm: Ah, yes, I mindlessly replaced something with `Default::default()` without actually checking what I'm doing. I'll add `Default` derivation to `Date`, since even the comment above this method implies this is the correct value.

ximon18:
LGTM!

cryptoki/src/object.rs
```diff
@@ -884,41 +884,49 @@ impl TryFrom<CK_ATTRIBUTE> for Attribute {
...
-                    let day = String::from_utf8_lossy(Vec::from((*date).day).as_slice())
-                        .trim_end()
-                        .to_string();
-                    Ok(Attribute::StartDate(Date::new_from_str_slice(
-                        year.as_str(),
-                        month.as_str(),
-                        day.as_str(),
-                    )?))
+                if val.is_empty() {
+                    Ok(Attribute::EndDate(Date::new_empty()))
```
anta5010, line 911: Shouldn't it be `Ok(Attribute::StartDate(Date::new_empty()))` ?

ionut-arm: Yes, thank you!!!

wiktor-k: And that's why 2 approvals are definitely a good idea: good catch Anton :clap:

anta5010 (approved):

ximon18:
@ionut-arm : Am I right in thinking that these new capabilities are still not available in a released version of the `cryptoki` crate?

ionut-arm:
> @ionut-arm : Am I right in thinking that these new capabilities are still not available in a released version of the `cryptoki` crate?

Ah, no, sorry, what I meant is that after I release it'll be possible to update, but I haven't published the new version to crates.io. 

And yes, I do have to update the changelog too... 😬

ximon18:
So, @ionut-arm, is there any idea when a new release of cryptoki including these improvements will be made?

ionut-arm:
Hey @ximon18 ! Sure, I can prioritise a release of this, though it will be quite a change to some parts of the API (just to make sure you're aware 😄 ).

Also, I just realised that when I answered the question above I was in zombie mode and thought I was answering the `cryptoki-sys` convo, apologies.

ximon18:
Thanks @ionut-arm. It's not per se that it is urgent, more a chicken and egg situation that there are some new things I requested that I could take advantage of, but not without there first being a release ;-)

ximon18:
> Hey @ximon18 ! Sure, I can prioritise a release of this, though it will be quite a change to some parts of the API (just to make sure you're aware smile ).

@ionut-arm: I just did a test compile against current `main` of `cryptoki` which worked fine, in our application I only had to change the call to `open_session()` (and raised #97 as a possible improvement to consider before you release).

PR #90, opened 2022-04-29
session_management: Add ability to login with raw bytes

Subject38:
Rationale for this addition is identical to https://github.com/mheese/rust-pkcs11/pull/8. Some dongle drivers are insane and allow you to use invalid strings as a PIN. As a result we need a way to pass in a raw slice of bytes.

### DIFF
--- a/cryptoki/src/session/mod.rs
+++ b/cryptoki/src/session/mod.rs
@@ -100,6 +100,22 @@ impl Session {
         session_management::login(self, user_type, pin)
     }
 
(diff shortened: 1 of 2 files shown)


hug-dev (approved):
Looks good, thanks for the PR!

ionut-arm (approved):
Excellent, thank you very much for contributing this!

I had one question - do you have a "list" (or at least some examples) of PKCS11 implementations where this is needed, apart from the Safenet example from the PR you linked? This kind of edge case sounds like something we should be covering in Parsec, and it would be really good to have documentation as to which token implementations might need it.

cc @paulhowardarm

PR #96, opened 2022-08-04
Prepare release of changes to bindings

ionut-arm:
cc @ximon18

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,34 @@
 # Changelog
 
-## [cryptoki-0.4.0](https://github.com/parallaxsecond/rust-cryptoki/tree/cryptoki-0.4.0) (2021-11-22)
+## [cryptoki-sys-0.1.4](https://github.com/parallaxsecond/rust-cryptoki/tree/cryptoki-sys-0.1.4) (2022-08-03)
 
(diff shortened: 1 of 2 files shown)


ionut-arm:
Not sure if I'll be able to release this today, but will get to it tomorrow if not :)

wiktor-k (approved):
LGTM :tada:

ximon18:
Thanks @ionut-arm!

I assume this will be a release of both cryptoki-sys and cryptoki, is that right?

Tomorrow or soon is fine, I just wanted to understand whether it would be soon or later, very welcome news, thankyou!

ionut-arm:
> I assume this will be a release of both cryptoki-sys and cryptoki, is that right?

That's actually not necessary - because I'm bumping the patch number, you can simply do a `cargo update -p cryptoki-sys` and you'll be able to use the new version without an update to `cryptoki`.

ximon18:
Don't you also need to update CHANGELOG.md?

ximon18:
Hi all,

Do you plan to go ahead with the 0.1.4 release of `cryptoki-sys` soon? I ask because we're getting ready to make a new release candidate of our product and ideally would like to include this change but cannot as it is not released yet.

Ximon

ionut-arm:
I was hoping to get another approval, but it should be fine as is. I'll release it now.

ionut-arm:
Done!

ximon18:
Thanks @ionut-arm !

PR #99, opened 2022-08-31
Use psa-crypto with updated zeroize

gowthamsk-arm:
Relates to #602

Signed-off-by: Gowtham Suresh Kumar <gowtham.sureshkumar@arm.com>

### DIFF
--- a/cryptoki/Cargo.toml
+++ b/cryptoki/Cargo.toml
@@ -16,7 +16,7 @@ bitflags = "1.3"
 libloading = "0.7.0"
 log = "0.4.14"
 derivative = "2.2.0"
-psa-crypto = { version = "0.9.0", default-features = false, optional = true }
+psa-crypto = { git = "https://github.com/parallaxsecond/rust-psa-crypto", rev = "22ebb4d8e479e1c0d853398cf8487d6389b1ed6f", default-features = false, optional = true }
 cryptoki-sys = { path = "../cryptoki-sys", version = "0.1.3" }
 paste = "1.0.6"
 


ionut-arm:
The version of `cryptoki` we use in the Parsec service isn't the one from `main`, it's based on the [cryptoki-0.2.x](https://github.com/parallaxsecond/rust-cryptoki/tree/cryptoki-0.2.x) branch.

One thing to remember here is that if the version we'll be releasing of `psa-crypto` is going to be a "point" release (as in, we only increase the patch number), then that release can be included in the service (for example) without having to update the `Cargo.toml`, by just doing a `cargo update`.

Another thing to consider for the future is that `cryptoki` should really not depend on `psa-crypto`. We should implement those conversions somewhere else, preferably in the Parsec service. I think we only did it because you can't have `TryFrom` and `From` implemented outside the crate, but we could just use some functions to convert, like we do for everything else in the service.

gowthamsk-arm:
Thanks for the comment Ionut. :) So in that case should we update the documentation to explicitly run `cargo update` before building the service and drop this PR along with the PR raised for the interfaces repo? 

Also regarding the `TryFrom` and `From` dependency, should I create an issue? :)

ionut-arm:
> Thanks for the comment Ionut. :) So in that case should we update the documentation to explicitly run `cargo update` before building the service and drop this PR along with the PR raised for the interfaces repo?

We should update the documentation, but you will sometimes need a minor/major version update in the dependency, in which case you do need to change the `Cargo.toml` and release the other stuff as well.

> Also regarding the `TryFrom` and `From` dependency, should I create an issue? :)

Yes, please!

wiktor-k:
:thinking: It would be nice to have a proper release of psa-crypto beforehand and then update to numbered version instead of somewhat ugly git dependency... Is there anything blocking psa-crypto from a new release?

(I've read https://github.com/parallaxsecond/parsec/issues/602 but releasing new psa-crypto doesn't seem to be a problem).

gowthamsk-arm:
We will be releasing psa-crypto crate today. This MR will be dropped. :)

gowthamsk-arm:
@ionut-arm I am creating the ticket for removing the conversions `impl TryFrom<psa_crypto::types::algorithm::Algorithm> for Mechanism` and `pub fn from_psa_crypto_hash` from this crate and implementing the conversion functions under Parsec. Is there anything else that should be considered as part of this ticket?

ionut-arm:
> @ionut-arm I am creating the issue for removing the conversions `impl TryFrom<psa_crypto::types::algorithm::Algorithm> for Mechanism` and `pub fn from_psa_crypto_hash` from this crate and implementing the conversion functions under Parsec. Is there anything else that should be considered as part of this issue?

Nah, that should be enough - getting anything PSA-related removed from this crate is the goal

gowthamsk-arm:
Created #100

gowthamsk-arm:
Not required.

PR #101, opened 2022-09-02
Make separate constructors for RO/RW sessions

ionut-arm:
The previous constructor took a boolean argument which made it difficult
to understand what the user code would be doing. That constructor is now
two separate functions, one for RO sessions, one for RW sessions. The
`_no_callback` part of the method name was also removed and added to the
documentation instead.

cc @ximon18

### DIFF
--- a/cryptoki/tests/basic.rs
+++ b/cryptoki/tests/basic.rs
@@ -29,7 +29,7 @@ fn sign_verify() -> Result<()> {
     let (pkcs11, slot) = init_pins();
 
     // open a session
-    let session = pkcs11.open_session_no_callback(slot, true)?;
+    let session = pkcs11.open_rw_session(slot)?;
 
     // log in the session
     session.login(UserType::User, Some(USER_PIN))?;
@@ -77,7 +77,7 @@ fn encrypt_decrypt() -> Result<()> {
     let (pkcs11, slot) = init_pins();
 
     // open a session
-    let session = pkcs11.open_session_no_callback(slot, true)?;
+    let session = pkcs11.open_rw_session(slot)?;
 
     // log in the session
     session.login(UserType::User, Some(USER_PIN))?;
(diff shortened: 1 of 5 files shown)


cryptoki/tests/basic.rs
```diff
@@ -723,3 +723,61 @@ fn aes_key_attributes_test() -> Result<()> {
...
+    // Try out Read-Only session
+    {
+        // open a session
+        let ro_session = pkcs11.open_ro_session(slot)?;
+
+        // log in the session
+        ro_session.login(UserType::User, Some(USER_PIN))?;
+
+        // generate a key pair
+        // This should NOT work using the Read/Write session
```
ximon18, line 754: I think you meant "Read-Only" in this comment rather than "Read/Write".

ximon18:
> The _no_callback part of the method name was also removed and added to the
documentation instead.

Should the same change also be made to `session_management::open_session_no_callback()` for consistency?

ximon18 (approved):

ionut-arm:
> Should the same change also be made to session_management::open_session_no_callback() for consistency?

I was wondering, but I felt more comfortable with a more verbose (but accurate) name for an internal function, though maybe consistency is more important 🤔 Will change

cryptoki/tests/basic.rs
```diff
@@ -723,3 +723,61 @@ fn aes_key_attributes_test() -> Result<()> {
...
+        let e = ro_session.create_object(&template).unwrap_err();
+
+        if let Error::Pkcs11(RvError::SessionReadOnly) = e {
```
mohamedasaker-arm, line 757: Minor: why not use `assert_eq!` ?

ionut-arm: Because I can't derive `PartialEq` on `Error`, as it contains some types that don't implement it 😞 So I'd have to use an `if let` anyway.

ximon18 (approved):

mohamedasaker-arm (approved):
Thanks ✅

wiktor-k (approved):
This looks good indeed :+1: just remember to bump the minor version (since the API change) but this is a nice improvement for readability.

PR #103, opened 2022-09-23
Implement Eq for SessionState

a1ien:
Without this it's impossible to compare SessionState.

Can you also made release after accept this PR please.

### DIFF
--- a/cryptoki/src/session/session_info.rs
+++ b/cryptoki/src/session/session_info.rs
@@ -67,7 +67,7 @@ impl TryFrom<CK_SESSION_INFO> for SessionInfo {
 
 /// The current state of the session which describes access to token and session
 /// obects based on user type and login status
-#[derive(Debug, Copy, Clone)]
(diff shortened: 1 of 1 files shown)


cryptoki/src/session/session_info.rs
```diff
@@ -67,7 +67,7 @@ impl TryFrom<CK_SESSION_INFO> for SessionInfo {
 
 /// The current state of the session which describes access to token and session
 /// obects based on user type and login status
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, Eq)]
```
ionut-arm, line 70: I think you need `PartialEq` as well

ionut-arm (commented):
Thanks for the patch!

ionut-arm:
> Can you also made release after accept this PR please.

Sure, can do a point release, though it'll take a few days, probably.

ionut-arm (approved):

Issue #106, opened 2022-11-18
Expose more fine-grained control over `find_objects`


jhagborgftx:
Currently, the only way to search for objects is through `Session::find_objects`, which returns a vector of all objects.  If there may be a huge number of objects, it's desirable to be able to find only a bounded number of objects.  This is possible with PKCS#11, but the functionality is not exposed by this crate.

Further, it may be desirable to iterate through objects incrementally, i.e. calling `C_FindObjects` multiple times, with some processing in between each call.  Also, the user may care whether the library makes several small calls to `C_FindObjects`, or one large one, especially if the HSM is being accessed over a network.

I believe all of these needs may be met safely by the following API:

```rust
impl Session {
    // ...

    pub fn find_objects_init<'a>(&'a mut self, template: &[Attribute]) -> Result<FindObjects<'a>> {
        // call `C_FindObjectsInit`
    }
}

pub FindObjects<'a> {
    session: '&mut Session // exclusive borrow to ensure at most one object search per session
}

impl<'a> FindObjects<'a> {
    pub fn find_next(&self, object_count: usize) -> Result<Option<Vec<ObjectHandle>>> {
        // call `C_FindObjects` once, with a buffer of size `object_count`.
        // return Ok(Some(...)) if we found objects, or Ok(None) if the search is over
    }
}

impl Drop for FindObjects { ... } // call `C_FindObjectsFinal` on drop
```

Although `find_next` *could* return just `Result<Vec<ObjectHandle>>`, I think returning an `Option` is more ergonomic, since
 - It's consistent with iterators and streams, which return `None` when finished
 - It reminds the user they need to check for termination
 - It allows for looping over all objects with `while let` notation

I believe the current implementation of `Session::find_objects`, as well as any other versions we might want to make (e.g. `Session::find_objects_bounded(template, object_count)`), can be implemented in terms of this one.

I will make a PR with something like this soon, but first I would like to collect some feedback about the API.

jhagborgftx:
One downside to the API above the mutable borrow means it is not possible to interleave calls to `find_next` with calls to other session operations.  I think it's possible to recover that by adding a method `session(&self) -> &Session` to `FindObjects`.  Since it only returns an & reference, it should be possible to do any other session operation *except* starting a new object search.

Also, I just realized that it would be necessary to change `find_objects` to take a mutable reference as well, to prevent someone from calling `find_objects` while another object search is ongoing.  That would make this a breaking change.

[...]

PR #109, opened 2022-11-30
Incremental object search

jhagborgftx:
This PR implements #106.  It also adds tests for both new and old-style object searches, and changes `find_objects` to use this new API internally.

As it is, `find_next(0)` returns `Some(vec![])`, which is I *think* the least surprising thing to do?  It shouldn't return `None`, since there may be more objects remaining.  I think it would also make sense if it panicked.  Let me know what you think.

### DIFF
--- a/cryptoki/tests/basic.rs
+++ b/cryptoki/tests/basic.rs
@@ -912,3 +912,141 @@ fn find_objects_many() -> Result<()> {
     assert_eq!(expected_labels, labels);
     Ok(())
 }
+
+// Find all objects incrementally (i.e. empty template)
+#[test]
+#[serial]
+fn find_objects_incrementally_all() -> Result<()> {
+    let (pkcs11, slot) = init_pins();
+    let mut session = pkcs11.open_rw_session(slot)?;
+    session.login(UserType::User, Some(USER_PIN))?;
+
+    // Generate some sample objects
+    let expected_labels = vec!["bar", "baz", "foo"];
+    generate_sample_objects(&session, &expected_labels)?;
+
+    // Find all objects
+    let mut search = session.find_objects_init(&[])?;
+    let objects = search.find_next(10)?;
+    assert!(search.find_next(10)?.is_empty());
+    drop(search);
+
+    // Check that we get the same objects back
+    let labels = object_labels_sorted(&session, objects)?;
+    assert_eq!(expected_labels, labels);
+
+    Ok(())
+}
+
+// Find objects incrementally when none match.
+#[test]
+#[serial]
+fn find_objects_incrementally_none() -> Result<()> {
+    let (pkcs11, slot) = init_pins();
+    let mut session = pkcs11.open_rw_session(slot)?;
+    session.login(UserType::User, Some(USER_PIN))?;
+
+    // Generate a sample object labeled "foo"
+    generate_sample_objects(&session, ["foo"])?;
+
+    // Search for objects labeled "bar"
+    let mut search = session.find_objects_init(&[Attribute::Label(b"bar".to_vec())])?;
+    assert!(search.find_next(10)?.is_empty());
+    Ok(())
+}
+
+// Find objects incrementally in a single batch.
+#[test]
+#[serial]
+fn find_objects_incrementally_single_batch() -> Result<()> {
+    let (pkcs11, slot) = init_pins();
+    let mut session = pkcs11.open_rw_session(slot)?;
+    session.login(UserType::User, Some(USER_PIN))?;
+
+    // generate some sample AES keys to match
+    let expected_labels = vec!["bar", "baz", "foo"];
+    generate_sample_objects(&session, &expected_labels)?;
+
+    // generate a key that shouldn't match (DES3 vs AES)
+    session.generate_key(
+        &Mechanism::Des3KeyGen,
+        &[Attribute::Label(b"quux".to_vec())],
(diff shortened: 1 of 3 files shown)


wiktor-k (dismissed):
LGTM. It kind of bothers me that `None` and `Some(vec![])` are basically the same state but using `Option` has the advantage of nice `while let Some(...)` loops on the client side :thinking: 

Passing 0 to the function is indeed unfortunate. On one hand it's good when the library never panics, on the other some projects (like `rust-openssl`) panic left and right on bad programmer-supplied data :man_shrugging:

jhagborgftx:
On Thu, 2022-12-01 at 00:48 -0800, Wiktor Kwapisiewicz wrote:
> LGTM. It kind of bothers me that None and Some(vec![]) are basically
> the same state but using Option has the advantage of nice while let
> Some(...) loops on the client side 🤔

I agree that it's not so nice.  I thought about it a bit more, and I
think we can just return a vector, and the user can use slice patterns
to get a nice-ish syntax in most cases:

```rust
while let ref chunk @ [_, ..] = search.find_next(SIZE)?[..] {
    // now chunk is a &[ObjectHandle], and is non-empty
}
```

However, if the user needs to take the vector and own it, they're left
with either:

```rust
loop {
    let chunk = search.find_next(SIZE)?;
    if
chunk.is_empty() { break; }
    // use chunk here...
}
```

or 

```rust
let mut chunk = vec![];
while { chunk = search.find_next(SIZE)?; !chunk.is_empty() } {
    // use chunk here...
}
```

With all this in mind, I don't think the strangeness of Option<Vec<_>>
is really worth it.

As for find_objects(0), I think we can just return vec![], in order to
be consistent with how `read` works.

jhagborgftx:
Updated to return just a vector, rather than an option of vector.

wiktor-k:
Thanks for the updates! I did take some time to add the iterator API on top of it: https://github.com/wiktor-k/rust-cryptoki/commit/1c00f377a10e45260375fcd811ab566209498d00

I'm still thinking about the API ease of use from the perspective of the API clients :thinking: 

For the time being maybe let's wait until @ionut-arm voices his opinion (since he's the most senior here along with @hug-dev)? Sorry that this takes a little bit of time but it's the end of year and everyone is super busy...

cryptoki/tests/basic.rs
```diff
@@ -781,3 +783,270 @@ fn ro_rw_session_test() -> Result<()> {
...
+
+    // Generate some sample objects
+    let expected_labels = vec!["bar", "baz", "foo"];
+    generate_sample_objects(&session, &expected_labels)?;
+
+    // Find all objects
+    let mut search = session.find_objects_init(&[])?;
+    let objects = search.find_next(10)?;
+    assert!(search.find_next(10)?.is_empty());
+    drop(search);
```
ionut-arm, line 932: Hmmmmm... Would it be possible to add a method to `FindObjects`, something like `finalize`, that just takes ownership of the value and lets it drop? It seems suboptimal to have to drop the object manually to achieve this, and an empty method isn't too bad.

jhagborgftx:
On Tue, 2022-12-06 at 02:57 -0800, Ionuț Mihalcea wrote:
> > +// Search 10 elements at a time
> +const MAX_OBJECT_COUNT: usize = 10;
> Any reason for this choice? Just curious

That wasn't my addition.  Look at the old implementation for
`find_objects`.

jhagborgftx:
On Tue, 2022-12-06 at 02:57 -0800, Ionuț Mihalcea wrote:
> Hmmmmm... Would it be possible to add a method to FindObjects,
> something like finalize, that just takes ownership of the value and
> lets it drop? It seems suboptimal to have to drop the object manually
> to achieve this, and an empty method isn't too bad.

Sure, then we can also get the benefit of a place to return errors from
C_FindObjectsFinal.  I'd call it `final` to be consistent with `init`.

Either way, I don't think manual dropping is that strange.  The docs
for MutexGuard from std have an example.  The same could also be
acheived by just adding some curly braces around the search.

ionut-arm:
> Either way, I don't think manual dropping is that strange.  The docs
for MutexGuard from std have an example.  The same could also be
acheived by just adding some curly braces around the search.

Sure, we've had cases before where dropping something was the way to go, but since we know this is a potential use-case, it seems quite straight-forward to add something to do it "natively". But I get your point!

One question I wanted to raise is: how much time do you have/are you willing to put into resolving comments on these PRs? Because if you only want to get them in as is, we could review the changes here, raise issues for what's left (like this new method), and fix those afterwards in separate PRs. Asking only because I don't want to pressure you into putting in more effort.

jhagborgftx:
On Thu, 2022-12-08 at 02:43 -0800, Ionuț Mihalcea wrote:
> One question I wanted to raise is: how much time do you have/are you
> willing to put into resolving comments on these PRs? Because if you
> only want to get them in as is, we could review the changes here,
> raise issues for what's left (like this new method), and fix those
> afterwards in separate PRs. Asking only because I don't want to
> pressure you into putting in more effort.

I'd like to resolve everything before we get them in.  I've had other
things on my plate the past few days, but I can definately get these
updated within a day or two.

hug-dev:
#223 fixed the upstream issue, so this could probably be closed if you agree!

wiktor-k:
Closing, as this has already been implemented. Thanks for your time! 👋

PR #110, opened 2022-12-02
Make RSA OAEP and ECDH1 safe using lifetime parameters

### LINKED ISSUE #107: RSA OAEP interface is unsound
The following code compiles without error or warning:

```rust
use cryptoki::context::{CInitializeArgs, Pkcs11};
use cryptoki::mechanism::{Mechanism, MechanismType};
use cryptoki::mechanism::rsa::{PkcsMgfType, PkcsOaepParams, PkcsOaepSourceType};
use cryptoki::object::Attribute;
use cryptoki::session::UserType;
use std::error::Error;

const PKCS11_LIB_PATH: &str = "/usr/lib/softhsm/libsofthsm2.so";
const SLOT_ID: u64 = 1359578051; // from softhsm2-util
const PIN: &str = "testpin";

fn main() -> Result<(), Box<dyn Error>> {
    let mut context = Pkcs11::new(PKCS11_LIB_PATH)?;
    context.initialize(CInitializeArgs::OsThreads)?;
    let session = context.open_rw_session(SLOT_ID.try_into()?)?;
    session.login(UserType::User, Some(PIN))?;

    let pub_key_template = [Attribute::ModulusBits(2048.into())];
    let (pubkey, _privkey) = session.generate_key_pair(&Mechanism::RsaPkcsKeyPairGen,
                                                       &pub_key_template, &[])?;

    let encrypt_mechanism = Mechanism::RsaPkcsOaep(PkcsOaepParams {
        hash_alg: MechanismType::SHA1,
        mgf: PkcsMgfType::MGF1_SHA1,
        source: PkcsOaepSourceType::DATA_SPECIFIED,
        source_data: 0xBADC0DE as _, // uh oh!
        source_data_len: 1.into(),
    });

    session.encrypt(&encrypt_mechanism, pubkey, b"Hello, world!")?;

    Ok(())
}
```

This is almost certainly UB (although in this example, libsofthsm returns CKR_BAD_ARGUMENTS, only because it does not support non-null `source_data`).  Rust code should not expose an interface that may dereference a user-supplied raw pointer, without marking that interface `unsafe`.

I don't care much about *this* parameter, but the same problem exists for other, not-yet-supported mechanisms whose parameters contain pointers (notably, AES-GCM and AES-CCM).  I *do* care about safe interfaces for those, and it would make sense to treat them all consistently.

# Possible solutions

No matter what, we can't let the user supply arbitrary pointers, without making them call an `unsafe` function.  The internals of `PkcsOaepParams` should be private, with one or more safe constructors (and possibly an unsafe constructor taking a raw pointer).

## Don't support `source_data` at all

I don't see any tests for this feature.  libsofthsm2 doesn't support it.  Is anyone using this thing anyway?  Maybe we just force it to be null.

On the other hand, we'll need to figure out the same thing for AES-GCM and AES-CCM, where we *do* care about non-null parameters.

## Make `PkcsOaepParams` own its source data

This is ergonomic, although inefficient if `source_data` is large and must be shared.  The spec allows `source_data` to be huge (as large as the hash function supports), but again, is anyone doing that?

## Add a lifetime parameter to `Mechanism` and `PkcsOaepParams`

This allows for more flexibility, although it clutters the interface of anything that uses a `Mechanism`.  Luckily, any variant which does not use the lifetime parameter can be inferred `'static`.

This is what I personally would like most, but I understand it's an invasive change.

## Provide an `unsafe` constructor taking a pointer and length

This isn't great if it's the *only* way, but it does provide a useful backdoor if we go with forcing ownership.

jhagborgftx:
This PR adds a lifetime parameter to `Mechanism`.  This allows the user to pass references to slices in places where the C interface to PKCS11 expects pointers and lengths.  This fixes #107, and makes ECDH1 safe as well.

The new interfaces have the same level of extensibility as the old interface, in the sense that we can add new KDFs or encoding parameter sources without it being a breaking change, but the user cannot add their own.

### DIFF
--- a/cryptoki/src/mechanism/elliptic_curve.rs
+++ b/cryptoki/src/mechanism/elliptic_curve.rs
@@ -1,12 +1,10 @@
 //! ECDH mechanism types
 
-use crate::error::{Error, Result};
 use crate::types::Ulong;
 use cryptoki_sys::*;
-use log::error;
-use std::convert::TryFrom;
-use std::ffi::c_void;
-use std::ops::Deref;
+use std::convert::TryInto;
+use std::marker::PhantomData;
+use std::ptr;
 
 /// ECDH derivation parameters.
 ///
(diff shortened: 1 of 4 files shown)


ionut-arm (commented):
Thanks for the patch! It mostly looks good to me, had a couple of comments below, will continue the review later on (though I think I've covered it all)

jhagborgftx:
On Mon, 2022-12-05 at 03:09 -0800, Ionuț Mihalcea wrote:
> +    pub fn new(kdf: EcKdf<'a>, public_data: &'a [u8]) -> Self {
> +        Self {
> +            kdf: kdf.kdf_type,
> +            shared_data_len: kdf
> +                .shared_data
> +                .map_or(0, <[u8]>::len)
> +                .try_into()
> +                .expect("usize can not fit in CK_ULONG"),
> Hmm, could we return an error here instead of panicking? Same for the
> other uses of expect in this patch.

Ah, I forgot LLP64 platforms exist!

This was to be consistent with two other places in this crate where we
cast usize to Ulong: in the From<&Mechanism> instance for CK_MECHANISM,
and in the From<&Attribute> instance for CK_ATTRIBUTE.  In the latter
case, the user can certainly trigger a panic, e.g. by trying to create
a 2^32+1 byte long CKO_DATA object on 64-bit Windows.

I agree this sort of panicking is bad, but it's also annoying to have
to write boilerplate around something that's infallible on every other
OS/architecture combination.

Perhaps we should introduce a type like Pkcs11Slice<'a, T>, whose
length is gauranteed to fit into a CK_ULONG.  There could be an
infallible conversion from &[T] that is conditionally compiled, and a
fallible version which is always available.  Same for Pkcs11Vec<T>,
where Vec<T> is used today.  Let me know what you think.

> Perhaps expanding this a bit to make it clear that it's a
> constructor? And similarly for data_specified

Sure thing.

ionut-arm:
> This was to be consistent with two other places in this crate where we
cast usize to Ulong: in the From<&Mechanism> instance for CK_MECHANISM,
and in the From<&Attribute> instance for CK_ATTRIBUTE.  In the latter
case, the user can certainly trigger a panic, e.g. by trying to create
a 2^32+1 byte long CKO_DATA object on 64-bit Windows.
> 
> I agree this sort of panicking is bad, but it's also annoying to have
to write boilerplate around something that's infallible on every other
OS/architecture combination.

Yeah, no worries, let's leave it as is, too much of a niche case - if we get comments about it we can change everywhere.

ionut-arm (approved):
👍🏻

wiktor-k (approved):

wiktor-k:
Since this is approved I'll merge it to `main`.

PR #111, opened 2022-12-05
Add CBC mode block ciphers

jhagborgftx:
This PR adds support for CBC-mode AES, DES, and DES3, both with and without PKCS#7 padding.  All of these mechanisms take their IVs as owned, fixed-sized byte arrays.  This is consistent with the general approach I've outlined in #107, where the parameter is owned by `Mechanism`, but any pointers in that parameter are replaced with references.

### DIFF
--- a/cryptoki/src/mechanism/mod.rs
+++ b/cryptoki/src/mechanism/mod.rs
@@ -32,6 +32,22 @@ impl MechanismType {
     pub const AES_KEY_GEN: MechanismType = MechanismType {
         val: CKM_AES_KEY_GEN,
     };
+    /// AES-CBC mechanism
+    ///
+    /// For encryption, the message length must be a multiple of the block
+    /// size.  For wrapping, the mechanism encrypts the value of the key,
+    /// padded on the trailing end with up to block size minus one null bytes.
+    /// For unwrapping, the result is truncated according to the key type and
+    /// the length provided by the template.
+    pub const AES_CBC: MechanismType = MechanismType { val: CKM_AES_CBC };
+    /// AES-CBC with PKCS#7 padding mechanism
+    ///
+    /// The plaintext may be any size.  The PKCS#7 padding allows the length of
+    /// the plaintext to be recovered from the ciphertext.  Therefore no length
+    /// should be provided when unwrapping keys with this mechanism.
+    pub const AES_CBC_PAD: MechanismType = MechanismType {
+        val: CKM_AES_CBC_PAD,
+    };
     /// AES-ECB mechanism
     pub const AES_ECB: MechanismType = MechanismType { val: CKM_AES_ECB };
     /// AES key wrap mechanism.  This mechanism can only wrap a key or encrypt a block of data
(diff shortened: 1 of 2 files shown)


cryptoki/tests/basic.rs
```diff
@@ -781,3 +781,64 @@ fn ro_rw_session_test() -> Result<()> {
...
+        Attribute::Class(ObjectClass::SECRET_KEY),
+        Attribute::KeyType(KeyType::AES),
+        Attribute::Value(key),
+        Attribute::Encrypt(true),
+    ];
+    let key_handle = session.create_object(&template)?;
+    let mechanism = Mechanism::AesCbc(iv);
+    let cipher = session.encrypt(&mechanism, key_handle, &plain)?;
+    assert_eq!(expected_cipher[..], cipher[..]);
+    Ok(())
```
wiktor-k, line 812: Looks OK but I've been wondering if you could stuff in `decrypt` too to see if the decrypted plaintext matches the input?

cryptoki/src/mechanism/mod.rs
```diff
@@ -816,6 +914,20 @@ impl From<&Mechanism> for CK_MECHANISM {
     }
 }
 
+// Make a CK_MECHANISM from mechanism type and parameter
+fn make_mechanism<T>(mechanism: CK_MECHANISM_TYPE, param: &T) -> CK_MECHANISM {
```
wiktor-k, line 918: This is nice :+1:

cryptoki/tests/basic.rs
ionut-arm: I'm ok either way on this, tbh, don't know if we need to test both directions since I assume we're testing the conversion of the mechanism here.

ionut-arm (approved):
As much as it pains me to see these supported, it's probably important to do it... :]

jhagborgftx:
On Wed, 2022-12-07 at 02:28 -0800, Ionuț Mihalcea wrote:
> @ionut-arm approved this pull request.
> 
> As much as it pains me to see these supported, it's probably
> important to do it... :]
> 
> In cryptoki/src/mechanism/mod.rs:
> 
> > +        // SAFETY: Although the type signature says *mut, none of
> the
> +        // mechanisms we support involve mutating the parameter, so
> +        // this cast is OK.
> Any suggestions for how we can keep this true and in our minds for
> future reviews? 🤔

I don't know of any mechanisms that modify their parameter block, but
I'm not sure if that's a general rule.  The `const` keyword doesn't
seem to appear once in the PKCS#11 spec...

In any case, the closest I can find is this note in UnwrapKey: 

https://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc323024159

> Some mechanisms may modify, or attempt to modify. the contents of the
> pMechanism structure at the same time that the key is unwrapped.

But I have no idea which mechanisms it's referring to. :/

jhagborgftx:
I think AES-CTR may be one such example.  Crypto operations increment
the counter inside the parameter block.  This is annoying, but I think
to properly support that mode we would need operations to take mutable
references to mechanisms (or do something with interior mutability, but
I think that would add more confusion than anything).

ionut-arm:
I'm happy with the PR on my side - I'll leave it to @wiktor-k to decide what to do with the testing!

cryptoki/tests/basic.rs
wiktor-k: Yeah, sorry, maybe I got carried over since the test already checks for expected ciphertext so let's not inflate the test code artificially here.

wiktor-k:
In general I'm also happy. @jhagborgftx if you could fix a couple of stylistic nits it would be perfect :+1: but it's not blocking so I'm approving it anyway ;)

wiktor-k:
I'll merge it not to leave approved PRs open. Thanks for your work @jhagborgftx !

PR #112, opened 2022-12-06
Add AEAD block cipher modes

jhagborgftx:
This PR adds support for AES-GCM and AES-CCM.  Both take parameter structs with pointers, and this is represented in Rust as references.  This PR builds on top of #110, and also will likely need tweaks to rebase onto #111, once that is merged.  Please look at individual commits for relevant changes.  I will rebase once those two PR's can land.

I needed to modify the p11-kit header file to add support for CCM.  Unfortunately, CCM is not supported by SoftHSM, so I only added tests for GCM.  I haven't had a chance to test CCM against other (proprietary) implementations yet, but really all there is to test is that we are packing the parameters into the structure correctly and not goofing up the lifetimes.


Bragolgirith:
Both blocker PRs seem to now have been merged 😇

wiktor-k:
Okay, I've spent a couple of minutes trying to reconcile this branch with `main` and came to the conclusion that we should cherry pick GCM from this PR and drop CCM.

GCM is already present in the header files and supported by SoftHSM and (at least from my experience) more widely used. So this would give us most of the benefits with less code to maintain.

What do you think about it @Bragolgirith ?

Bragolgirith:
Not sure how much my opinion weighs as a non-contributor, but I agree on all accounts. 

GCM is already a very juicy feature that will make a lot of folks (including me!) happy, and CCM (with the headers and tests) can be addressed in either this or a separate PR.

wiktor-k:
Superseded by #144 which only implements GCM.

PR #113, opened 2022-12-15
Add MechanismType::CKM_EDDSA & Mechanism::Eddsa

legounix:


ionut-arm:
Hey, thanks for the patch but I think this won't work (for now), because of something we should rectify... 

The current interface only works with v2.40 of the PKCS11 spec, which contains [no EdDSA mechanism](https://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html). I believe EdDSA was only added to [v3](https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html).

legounix:
> Hey, thanks for the patch but I think this won't work (for now), because of something we should rectify...
> 
> The current interface only works with v2.40 of the PKCS11 spec, which contains [no EdDSA mechanism](https://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html). I believe EdDSA was only added to [v3](https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html).

Yes updating the pkcs11 spec would nice! 
I could generate/sign/verify stuff with this and ed25519 with softhsm. I can add a test for this later on. So it worked with softhsm, or something worked :D I still don't feel like i understood pkcs11, currently doing some baby steps and diving into that whole field so take this with a grain of salt.

What makes me wonder, if pkcs11 2.40 spec didnt contain the eddsa mechanism, why is the constant listed in https://github.com/parallaxsecond/rust-cryptoki/blob/main/cryptoki-sys/pkcs11.h#L889 ?

ionut-arm:
> Yes updating the pkcs11 spec would nice!

I'll have a look at that, but it feels like it would be a fairly big task - mind you, I've not actually done a diff between the two, so can't say for certain. But because it's a bump in the major number I assume some breaking changes would've occurred, which we'd have to mitigate somehow. If it's only about adding some mechanisms, that would be great, but I wouldn't think that's it.

> What makes me wonder, if pkcs11 2.40 spec didnt contain the eddsa mechanism, why is the constant listed in https://github.com/parallaxsecond/rust-cryptoki/blob/main/cryptoki-sys/pkcs11.h#L889 ?

Indeed, no clue, I've checked the SoftHSM header file too, and it has the same constant defined, even though v2.40 of the spec doesn't contain any mention to EdDSA... I'm wondering if I'm the one confused here, or if everyone is, because I've seen other places where this "discrepancy" appears.

DeppLearning:
I can confirm the usefulness of this patch for SoftHSM and Thales Luna Cloud HSM and would love to see it merged. While originally not part of the PKCS#11 v2.40 standard, it looks like some vendors have nevertheless included it in their v2.40 implementations, [see](https://thalesdocs.com/dpod/services/luna_cloud_hsm/extern/client_guides/Content/sdk/luna_cloud_mechanisms/CKM_EDDSA.htm)?

wiktor-k:
I agree that this sounds like a relatively high value addition considering the costs. What do you think @ionut-arm ?

I'm even volunteering to rebase it and fix conflicts in case @legounix doesn't want to. (The lack of DCO is annoying though :/)

wiktor-k:
Superseded by #140.

PR #118, opened 2023-01-13
Split the implementation of session and pkcs11

baloo:
I know this is uncommon, and I don't know if that's recent or not, but Rust allows for an implementation of an object to happen in submodules.

I believe this was the intention with the various inline implementations so I'm submitting this alternative.


baloo:
I'd suggest reviewing this one ignoring whitespaces:
https://github.com/parallaxsecond/rust-cryptoki/pull/118/files?w=1

baloo:
(no hard feeling if you believe this is too much noise and you prefer to close it)

wiktor-k:
I'll need more time to think about it. I like the places where it removes `pub(super)` and leaves functions private. I think it makes the code more clear.

I've seen that in some places `pub(super)` (effectively private outside of crate) has been turned into regular `pub` (public). Is that on purpose?

baloo:
I should have been more explicit but a picture is worth a thousand words right? Let's see that goes with a code sample.

What you were doing was essentially:
```rust
struct Session;

impl Session {
    /// Initialize the normal user's pin for a token
    pub fn init_pin(&self, pin: &str) -> Result<()> {
        slot_token_management::init_pin(self, pin)
    }
    
    /// Log a session in.
    pub fn login(&self, user_type: UserType, pin: Option<&str>) -> Result<()> {
        session_management::login(self, user_type, pin)
    }
}
mod slot_token_management {
    // See public docs on stub in parent mod.rs
    #[inline(always)]
    pub(super) fn init_pin(session: &Session, pin: &str) -> Result<()> {
        todo!();
    }
}
mod session_management {
    // See public docs on stub in parent mod.rs
    #[inline(always)]
    pub(super) fn login(session: &Session, user_type: UserType, pin: Option<&str>) -> Result<()> {
        todo!();
    }
}
```

What I'm trying to explain here is that this is equivalent to doing:
```rust
struct Session;

mod slot_token_management {
    impl Session {
        /// Initialize the normal user's pin for a token
        pub fn init_pin(&self, pin: &str) -> Result<()> {
            todo!();
        }
    }
}
mod session_management {
    impl Session {
        /// Log a session in.
        pub fn login(&self, user_type: UserType, pin: Option<&str>) -> Result<()> {
            todo!();
        }
    }
}
```

baloo:
> I've seen that in some places `pub(super)` (effectively private outside of crate) has been turned into regular `pub` (public). Is that on purpose?

which one exactly went from `pub(super)` to `pub`?

cryptoki/src/session/slot_token_management.rs
```diff
@@ -6,30 +6,31 @@ use crate::error::{Result, Rv};
...
-            session.handle(),
-            old_pin.as_ptr() as *mut u8,
-            old_pin.len().try_into()?,
-            new_pin.as_ptr() as *mut u8,
-            new_pin.len().try_into()?,
-        ))
-        .into_result()
+    /// Changes the PIN of either the currently logged in user or of the `CKU_USER` if no user is
+    /// logged in.
+    pub fn set_pin(&self, old_pin: &str, new_pin: &str) -> Result<()> {
```
wiktor-k, line 24: >      which one exactly went from pub(super) to pub?

If my diff-eyes don't deceive me this is `pub` now but was `pub(super) fn set_pin` previously...?

baloo: https://github.com/parallaxsecond/rust-cryptoki/blob/9934be7902a3a5b143d33240fdc8c250fa1243ce/cryptoki/src/session/mod.rs#L82

baloo: (again, the method implementation stays in the submodule, but the external interface moved as I tried to explain here:
https://github.com/parallaxsecond/rust-cryptoki/pull/118#issuecomment-1382122686)

wiktor-k: Ah, right :facepalm: good point!

baloo:
What is your overall opinion about this?
This is a pretty large rewrite/reorg and also pretty brittle.

ionut-arm:
So, if I recall this correctly, the reason why we went with functions in modules, that get called from the appropriate method in the main module file, is because if you separate methods across modules your documentation also ends up grouped following the same lines. See some of the discussions in #55 about it.

I think as long as the documentation doesn't look too bad, I'd normally prefer your approach as well - we've used it in Parsec too. Unfortunately my bandwidth is quite limited at the moment, and the work I've been meaning to put into docs and tests here is on hold for some time... :(

baloo:
To my knowledge, the only difference between the previous implementation is indeed the documentation. The functions will be grouped per `impl Session` block and ordered per module ordering (I can tweak that to keep the previous order, just have to play with the import order in the `session/mod.rs`)
Before:
20230123_08h59m49s_grim
After:
20230123_09h03m35s_grim

(notice the multiple lines of `impl Session` in the after)

ionut-arm (approved):
I'm inclined to accept this. Hopefully the docs won't look too bad, though I think this pattern exists in more prominent parts of the Rust ecosystem.

wiktor-k (approved):
I'm also OK with this change. From my PoV it seems to make stuff simpler. The CI needs to be fixed though (rebasing this change on top of `main` should suffice).

baloo:
it's already on top of main, I think the new failures are due to a new release of rust. I'll submit the changes.

baloo:
see https://github.com/parallaxsecond/rust-cryptoki/pull/127

wiktor-k:
Great. Let's go! 🚀

PR #120, opened 2023-01-13
Simplify tests using TestResult

wiktor-k:
This is a follow-up to an old PR #39 and uses [packaged-as-a-library `TestResult` type](https://crates.io/crates/testresult) instead of an ad-hoc struct. The library provides better diagnostics due to small tweaks on how the error is captured but is otherwise identical with what the code does currently.

Actually I packaged the type into a crate and then later it was [polished by other people](https://github.com/wiktor-k/testresult/pull/3).

### DIFF
--- a/cryptoki/Cargo.toml
+++ b/cryptoki/Cargo.toml
@@ -24,6 +24,7 @@ paste = "1.0.6"
 num-traits = "0.2.14"
 hex = "0.4.3"
 serial_test = "0.5.1"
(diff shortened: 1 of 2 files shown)


ionut-arm (approved):
Nice, thank you!

paulhowardarm (approved):
Looks good, thanks. New dependency introduced, but looks very useful and doesn't pull in anything else indirectly.

PR #114, opened 2022-12-23
Update to remove const_err

marcsvll:
This PR only removed `const_err` since it will be a hard error in future: https://github.com/rust-lang/rust/issues/71800

Signed-off-by: Marcus de Lima <marcus.lima@azion.com>

PR #122, opened 2023-01-24
Update to remove const_err lint

wiktor-k:
This PR is exactly the same as #114 but rebased on top of `main` so that it builds successfully.


paulhowardarm (approved):
LGTM - thanks.

ionut-arm (approved):
🚀

Issue #121, opened 2023-01-20
`#[hsm_test]` attribute/macro


baloo:
Hi there!

I ended up creating an `#[hsm_test]` macro to be able to write:
```rust
#[cfg(test)]
mod tests {
    use cryptoki::{context::Pkcs11, slot::Slot};
    use softhsm_test::hsm_test;

    #[hsm_test]
    fn test_foo(pkcs11: Pkcs11, slot: Slot) {
        assert!(false, "oh no!");
    }
}
```

This will run the tests in a dedicated subprocess, set up a dedicated softhsm configuration in a test dir, update environment variables to then run the test.

Do you have any interest for such a contribution? (this will be at least one extra crate because one is a `proc-macro`)

wiktor-k:
> Do you have any interest for such a contribution? 

Good question. By "contribution" do you mean rewriting tests to use your macro? (This would have an advantage that there wouldn't be a time-dependency between them) Or do you mean having that macro as part of the crate?

In first case we could make the extra dependency a dev-dependency. In the second case if could be hidden behind a feature... Just a couple of thoughts :)

baloo:
> Good question. By "contribution" do you mean rewriting tests to use your macro? (This would have an advantage that there wouldn't be a time-dependency between them) Or do you mean having that macro as part of the crate?

I guess both :) The original question was about shifting ownership (and maintenance burden) here. But it would make sense to use it too.

We can definitely get everything behind a `test-harness` feature flags, although I'm not sure how to get CI run here if it's necessary to run our own tests.

wiktor-k:
I would gate it with something like `cfg(any(test, feature = "macros"))`. This way it would be available for tests, and also exposed if someone wants that using `macros` feature.

This way both scenarios are implementable.

I'll bring @ionut-arm here for opinion :)

baloo:
ha right!

ionut-arm:
Sorry, just catching up now...

> I would gate it with something like `cfg(any(test, feature = "macros"))`. This way it would be available for tests, and also exposed if someone wants that using `macros` feature.

Yeah, as pointed out in the PR, the `test` predicate is not set during integration tests... :] What we did in the past was to create a new feature called something like `testing` which we only use/set for these integration tests..

Otherwise I like the idea and the approach - commenting here instead of the PR. Thanks for putting in the effort! 🙏🏻

wiktor-k:
[...]

PR #123, opened 2023-01-27
Implement a `#[hsm_test]` macro

baloo:
This is an implementation of the macro I was talking about in #121.

This is still a work in progress. Unlike what we discussed in the issue, the `#[cfg(any(test, feature = "macros"))]` trick doesn't work as the `test` predicate is not turned on during integration tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,2 @@
 [workspace]
-members = ["cryptoki", "cryptoki-sys"]
+members = ["cryptoki", "cryptoki-derive", "cryptoki-sys"]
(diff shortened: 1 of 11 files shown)


wiktor-k:
In general I like the change. I'll just need a moment to properly read it so please bear with me 🙏

baloo:
Nothing urgent!
I think I can clean up the dependency graph if I introduced a `cryptoki-tests` (mostly dummy) and a `cryptoki-tests-derive` (where the macro lives) crate.

baloo:
won't work for non-unix targets like current solution

PR #124, opened 2023-02-01
Implement PartialEq/Eq for Attribute

Bobo1239:
I'm (mis)using this crate to simplify creating a basic PKCS#11 implementation. For that being able to compare `Attribute`s is helpful.

### DIFF
--- a/cryptoki/src/object.rs
+++ b/cryptoki/src/object.rs
@@ -403,7 +403,7 @@ impl TryFrom<CK_ATTRIBUTE_TYPE> for AttributeType {
     }
 }
 
-#[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


wiktor-k (approved):
The commit message mentions Attribute but the change also adds PartialEq/Eq for Date. I'm okay with that but ideally it would either be two commits or one commit with message that mentions both.

As for the change, it looks very much okay and raising the UX of the API :) Thanks! :+1:

Bobo1239:
Sure, added a mention to the commit message.

wiktor-k:
Looks great for me, thanks! 👏

ionut-arm (approved):
Thanks!! 👍🏻

PR #125, opened 2023-02-01
Fix CI (clippy warnings)

Bobo1239:
Only needed to adjust for the new uninlined_format_args lint.


wiktor-k (approved):
Yep, looks more readable too. Thanks! 👍

gowthamsk-arm (approved):

PR #127, opened 2023-02-13
fixup clippy errors

baloo:
Rust 1.67 brings new clippy suggestions, this causes CI errors.


ionut-arm (approved):
Thanks, LGTM

We'll need to pin down an MSRV - but that's not something to worry about here

wiktor-k (approved):
LGTM 👍 thanks!

baloo:
looks like it was already fixed ... and I'm just a dummy with git :D 
`git fetch` will fetch from my remote and not origin when I'm on a branched with my fork set as upstream. Made that mistake before, and made it again. 
Anyway, sorry for the noise.

Issue #126, opened 2023-02-11
Session Pool Management


sbihel:
Hi, this issue is just to say I wrote a session pool management crate based on `r2d2`: https://github.com/spruceid/r2d2-cryptoki.

I believe it's the right way of handling concurrent requests for service HSMs like cloud-hosted ones, but I thought maybe you'd have an opinion.

On a side note, I added a Zeroize wrapper for pins to ensure users understand how sensitive they are, and unless you think it's unnecessary I will create a PR as it would belong in this crate.

ionut-arm:
Hey! Thanks for the heads-up, the `r2d2` crate does look pretty nice. I'm considering adding some links, perhaps on the README, to crates that have useful "3rd party" functionality. 🤔 

In terms of a PR for zeroizing PINs - yes, please, that would be appreciated.

hug-dev:
Sounds like a good idea! Added some info about r2d2 in the README in https://github.com/parallaxsecond/rust-cryptoki/pull/242

A question, is there any security implications in always having a session open, ready to use?

PR #128, opened 2023-02-16
Add new Pin type with secrecy/zeroize wrapper for

sbihel:
Related to #126

In the end, I don't know if it's such a good idea. The code is only handling `&str` so it is up to you to decide if you think this breaking change is worth making users aware that they should be careful with their pins.


wiktor-k:
I think it looks mostly OK. I'm a liiiitle bit worried about name confusion with [`std::pin::Pin`](https://doc.rust-lang.org/std/pin/struct.Pin.html) struct. At least it confused me initially before I saw the alias :sweat_smile:

sbihel:
That's fair, and I mistakenly imported `std::pin::Pin` a few times... Here are a few options:
1. `PIN` but it's not really idiomatic;
2. `UserPin` but users might be confused and look for an `SOPin`;
3. `Pkcs11Pin`;
4. `AuthPin`;
5. `TokenPin`;
6. `LoginPin`; or
7. `SecretPin`.

I would vote for 2 but I don't have a strong preference.

ionut-arm:
Hey, sorry for the long delay.

I think I'd go for something like 4, 5, or 6 from your list, but I don't have any strong feelings between them. The reason I'd avoid 2 is exactly what you mentioned, that it's technically incorrect and perhaps confusing, if it does cover both SO and User PINs.

wiktor-k:
Yep. Maybe let's go with `AuthPin` and that's it. (otherwise the bikeshedding can take forever :D )

ionut-arm (approved):
That was fast! Thanks for the patch, I think at least one of your commits is missing a sign-off, though.

sbihel:
Sorry about that, fixed

wiktor-k (approved):
LGTM 👍

PR #132, opened 2023-03-15
Add binding for single-part digest function

ellerh:


cryptoki/src/session/digesting.rs
```diff
@@ -0,0 +1,54 @@
+// Copyright 2021 Contributors to the Parsec project.
```
ionut-arm, line 1:
```suggestion
// Copyright 2023 Contributors to the Parsec project.
```

ionut-arm (approved):
Hi, thanks for the patch, looks good to me! Only one change requested, and I think you need to provide a sign-off for the commit.

adamjohnstonarm:
I would suggest squashing these commits when merging to keep the copyright date clean.

wiktor-k (approved):
Indeed. Looks good 👍

ionut-arm:
@hcldan - please see #160. Unfortunately, the many changes that have occurred since 0.4.1 means we have to publish a 0.5.0 version now.

Issue #142, opened 2023-05-18
CKA_PUBLIC_KEY_INFO getting TypeInvalid


hcldan:
I hope you can give me a pointer.   I'm using a smartcard that has some certs on it for smime.

I can use the ActivClient smartcard tool on windows to poke around at the details, and I've been using this library fairly successfully pointing to their driver so far... until this.

I can see when I inspect the certs in the ActivClient program that they have a SubjectKeyIdentifier
![image](https://github.com/parallaxsecond/rust-cryptoki/assets/648319/7eb85699-6a4a-4ab3-bd0c-39c5914ac85f)

When I try to get the attribute from any of the objects returned, it always says TypeInvalid...
Is this a driver problem?  Am I misunderstanding something here?

hcldan:
hmm... the dll says it's for pkcs11 2.20... i guess that attribute isn't in that spec.

wiktor-k:
I'm not sure if you can query these attributes directly. IIUC you need to first get the X.509 certificate object from the card (for example using this object class: https://docs.rs/cryptoki/latest/cryptoki/object/struct.ObjectClass.html#associatedconstant.CERTIFICATE) and then parse the X.509 cert to get the field you want (the last part is outside of this library's scope).

Does it answer your question or am I completely wrong? :sweat_smile:

hcldan:
@wiktor-k That's interesting.   I am going to try to go that route, as my options are rather limited.

FWIW, it looks like in pkcs11 2.40, those attributes are defined and *should* work.  But I found the driver I have only supports pkcs11 2.20 (which does not define those attributes)

hcldan:
@wiktor-k I have the object_handle for the cert... how do I get the bytes?

wiktor-k:
I think you used [`Attribute::Value`](https://docs.rs/cryptoki/latest/cryptoki/object/enum.Attribute.html#variant.Value) right? (In general it's good to paste your solution for the next person that has the same problem :sweat_smile: )

hcldan:
@wiktor-k Yes, I was actually hoping to wrap things up with my task and contribute a few examples.

Is that something that would be welcome?   Is there a good place to put them?

hcldan:
@wiktor-k Sent a pr.  I'm not sure the doctests will work well seeing as you need a library to init the card.

hug-dev:
I assume that your questions were answered and I see that the PR is merged so closing for now. Feel free to re-open if there remains issues :)

PR #143, opened 2023-05-22
Add an example to `find_objects`

hcldan:
Adds some doc mentioned in #142

### DIFF
--- a/cryptoki/src/session/object_management.rs
+++ b/cryptoki/src/session/object_management.rs
@@ -14,6 +14,42 @@ const MAX_OBJECT_COUNT: usize = 10;
 
 impl Session {
     /// Search for session objects matching a template
+    ///
+    /// # Arguments
+    /// * `template` - A [Attribute] of search parameters that will be used
+    ///                 to find objects.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use cryptoki::session::Session;
+    /// use cryptoki::context::Pkcs11;
+    /// use cryptoki::object::{Attribute, AttributeType, CertificateType, ObjectClass, ObjectHandle};
+    ///
+    /// let mut client = Pkcs11::new(
+    ///     std::env::var("PKCS11_SOFTHSM2_MODULE")
+    ///        .unwrap_or_else(|_| "/usr/local/lib/softhsm/libsofthsm2.so".to_string()),
+    /// ).expect("Failed to load library");
+    /// client.initialize(cryptoki::context::CInitializeArgs::OsThreads).expect("Failed to init session");
+    ///
+    /// // Use the first slot
+    /// let slots = client.get_all_slots().expect("Failed to list slots on card");
+    /// let slot = slots.first().expect("Failed to find 1st slot");
+    /// let session = client.open_ro_session(*slot).expect("Failed to open read-only session");
+    ///
+    /// // Get handles to all of the x509 certificates on the card
+    /// let search = vec![Attribute::Class(ObjectClass::CERTIFICATE), Attribute::CertificateType(CertificateType::X_509)];
+    /// for handle in session.find_objects(&search).expect("Failed to get certificates from card") {
+    ///   // each cert: get the "value" which will be the raw certificate data
+    ///   for value in session.get_attributes(handle, &[AttributeType::Value]).expect("Failed to search for object atrributes") {
+    ///      match value {
+    ///       // Log the data
+    ///      Attribute::Value(value) => println!("Certificate value: {value:?}"),
+    ///      _ => (),
+    ///      }
+    ///   }
+    /// }
+    /// ```
     pub fn find_objects(&self, template: &[Attribute]) -> Result<Vec<ObjectHandle>> {
         let mut template: Vec<CK_ATTRIBUTE> = template.iter().map(|attr| attr.into()).collect();
 


wiktor-k:
:thinking: it looks like it fails due to bad formatting.

If you could try something like this locally:

```
cargo fmt
git commit -a --amend --no-edit --signoff
git push --force
```

It should be all that's needed.

Thanks for your time! :wave:

hcldan:
I assumed that --signoff was implied seeing as it was my commit I amended and I use commit signing

cryptoki/src/session/object_management.rs
```diff
@@ -14,6 +14,42 @@ const MAX_OBJECT_COUNT: usize = 10;
...
+    ///                 to find objects.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use cryptoki::session::Session;
+    /// use cryptoki::context::Pkcs11;
+    /// use cryptoki::object::{Attribute, AttributeType, CertificateType, ObjectClass, ObjectHandle};
+    ///
+    /// let mut client = Pkcs11::new(env!("PKCS11_LIBRARY_LOCATION")).expect("Failed to load library");
```
wiktor-k, line 29:
```
error: environment variable `PKCS11_LIBRARY_LOCATION` not defined at compile time
 --> src/session/object_management.rs:29:30
  |
8 | let mut client = Pkcs11::new(env!("PKCS11_LIBRARY_LOCATION")).expect("Failed to load library");
  |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: Use `std::env::var("PKCS11_LIBRARY_LOCATION")` to read the variable at run time
```

wiktor-k: Just in case it's helpful a similar code is in https://github.com/parallaxsecond/rust-cryptoki/blob/72673c8cfb2285ab86f485845f1ec439111a8b6a/cryptoki/tests/common.rs#L15

hcldan:
yes :) 
that helps

cryptoki/src/session/object_management.rs
```diff
@@ -14,6 +14,45 @@ const MAX_OBJECT_COUNT: usize = 10;
...
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use cryptoki::session::Session;
+    /// use cryptoki::context::Pkcs11;
+    /// use cryptoki::object::{Attribute, AttributeType, CertificateType, ObjectClass, ObjectHandle};
+    ///
+    /// let mut client = Pkcs11::new(
+    ///     env::var("PKCS11_SOFTHSM2_MODULE")
```
wiktor-k, line 30: We're almost there!

```
 9 |     env::var("PKCS11_SOFTHSM2_MODULE")
  |     ^^^ use of undeclared crate or module `env`
```

(I guess prefixing `env` with `std::` would be enough in this case).

Not sure if you're getting notifications from GitHub builds but if you click the "details" under failed job item you'd see the exact error detail.

Thanks for your time and persistence :) ! :wave:

hcldan:
stupid mistake...  doing too many things at once.

cryptoki/src/session/object_management.rs
hcldan: I am getting notifications, though I'm not having luck seeing details on mobile devices.

wiktor-k:
> stupid mistake... doing too many things at once.

Haha, no worries, I can totally relate!

As for the `let` error that happened in last build maybe you can try `cargo test --doc` locally to see if it passes and if for some reason you can't do that (eg SoftHSM missing) you could add [`no_run`](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes) attribute to the codeblock to make cargo only compile the test, not run it. This way you won't have to have any SoftHSM installed.

(For the record I'm running the test suite like that: `PKCS11_SOFTHSM2_MODULE=/usr/lib/softhsm/libsofthsm2.so cargo test` since I've got SoftHSM in another location).

cryptoki/src/session/object_management.rs
```diff
@@ -14,6 +14,42 @@ const MAX_OBJECT_COUNT: usize = 10;
...
+    /// let mut client = Pkcs11::new(
+    ///     std::env::var("PKCS11_SOFTHSM2_MODULE")
+    ///        .unwrap_or_else(|_| "/usr/local/lib/softhsm/libsofthsm2.so".to_string()),
+    /// ).expect("Failed to load library");
+    /// client.initialize(cryptoki::context::CInitializeArgs::OsThreads).expect("Failed to init session");
+    ///
+    /// // Use the first slot
+    /// let slots = client.get_all_slots().expect("Failed to list slots on card");
+    /// let slot = slots.first().expect("Failed to find 1st slot");
+    /// let session = client.open_ro_session(*slot).expect("Failed to open read-only session");
```
ionut-arm, line 38: I wonder if we need to make this visible as part of this example, as setting up a session is common to the vast majority of operations. Instead we could maybe point in a comment to a different place where this is shown.

ionut-arm (approved):
Hey! Thanks for the example, I've got one question but happy to go either way.

cryptoki/src/session/object_management.rs
```diff
@@ -14,6 +14,42 @@ const MAX_OBJECT_COUNT: usize = 10;
...
+    ///
+    /// // Get handles to all of the x509 certificates on the card
+    /// let search = vec![Attribute::Class(ObjectClass::CERTIFICATE), Attribute::CertificateType(CertificateType::X_509)];
+    /// for handle in session.find_objects(&search).expect("Failed to get certificates from card") {
+    ///   // each cert: get the "value" which will be the raw certificate data
+    ///   for value in session.get_attributes(handle, &[AttributeType::Value]).expect("Failed to search for object atrributes") {
+    ///      match value {
```
wiktor-k, line 45: I think this match with one arm can be more succinctly expressed using something like: `if let Attribute::Value(value) = value { ...`

cryptoki/src/session/object_management.rs
wiktor-k: Hmm yeah I agree that since most of the library is basically without examples it would be good to copy them to other places (say [`Pkcs11::open_ro_session`](https://docs.rs/cryptoki/latest/cryptoki/context/struct.Pkcs11.html#method.open_ro_session)) and hide them here (using `#` prefix in code examples) and maybe link to it.

This is more work so I'm not sure if you'd like to do it. As it is the example is already good I think.

hcldan:
I've been unable to continue on this for a while, sorry.   I do hope to get to it.

wiktor-k:
If you don't mind I'll rebase your PR, fix the remaining issues and merge (preserving the authorship - of course). :wave:

hcldan:
If you don't mind, that would be great.  Dan DumontE-mail:  ***@***.***  ----- Original message ----- From: "Wiktor Kwapisiewicz" ***@***.***> To: "parallaxsecond/rust-cryptoki" ***@***.***> Cc: "Dan Dumont" ***@***.***>, "Author" ***@***.***> Subject: Re: [parallaxsecond/rust-cryptoki] Add an example to `find_objects` (PR #143)Date: Fri, Aug 30, 2024 8:01 AM [CAUTION: This email is from outside the organization. Unless you trust the sender, don't click links or open attachments as it may be a phishing email, which can steal your information and compromise your computer.]  If you don't mind I'll rebase your PR, fix the remaining issues and merge (preserving the authorship - of course). 👋—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

PR #144, opened 2023-05-23
Add AES-GCM mechanism

wiktor-k:
Hi :wave: 

I've extracted GCM bits from #112 as they can be tested in SoftHSM and require minimal changes.

@jhagborgftx is still credited as the author.

One of the tests (with empty AAD) fails due to what appears to be [a bug in SoftHSM](https://github.com/opendnssec/SoftHSMv2/issues/605) so I've added `#[ignore]` to it (it's otherwise good IMO).

### DIFF
--- a/cryptoki/src/mechanism/aead.rs
+++ b/cryptoki/src/mechanism/aead.rs
@@ -1,0 +1,91 @@
+// Copyright 2023 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+//! AEAD block cipher mechanism types
+
+use crate::types::Ulong;
+use cryptoki_sys::*;
+use std::convert::TryInto;
+use std::marker::PhantomData;
+use std::slice;
+
+/// Parameters for AES-GCM.
+#[derive(Debug, Clone, Copy)]
+#[repr(transparent)]
+pub struct GcmParams<'a> {
+    inner: CK_GCM_PARAMS,
+    _marker: PhantomData<&'a [u8]>,
+}
+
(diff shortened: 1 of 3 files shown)


cryptoki/src/mechanism/aead.rs
```diff
@@ -0,0 +1,89 @@
+//! AEAD block cipher mechanism types
```
ionut-arm, line 1: Could you please add a copyright here?

ionut-arm (approved):
Thank you for porting the change! 🙏🏻 Got only one minor comment

cryptoki/src/mechanism/aead.rs
wiktor-k: D'oh, right! Thanks for spotting that!

cryptoki/src/mechanism/aead.rs
```diff
@@ -0,0 +1,91 @@
+// Copyright 2021 Contributors to the Parsec project.
```
ionut-arm, line 1: Is this the correct year? Was the file here before?

wiktor-k: :thinking: nope, actually I just copied it from the other file not being sure what's the procedure :sweat_smile: 

These date ranges are always problematic and I've seen [other projects remove them altogether](https://github.com/curl/curl/pull/10205). :shrug: 

(For this PR I fixed what you reported to keep this consistent).

ionut-arm (approved):
:+1:

PR #147, opened 2023-06-04
Add function name to errors and logs

ionut-arm:
Adding the name of the function that lead to a backend error to the log messages it generates and to the error returned to the client.

cc @ellerh

### DIFF
--- a/cryptoki/src/context/general_purpose.rs
+++ b/cryptoki/src/context/general_purpose.rs
@@ -7,6 +7,7 @@ use crate::error::{Result, Rv};
 use cryptoki_sys::{CK_C_INITIALIZE_ARGS, CK_INFO};
 use paste::paste;
 use std::convert::TryFrom;
+use std::fmt::Display;
 
 // See public docs on stub in parent mod.rs
 #[inline(always)]
@@ -18,7 +19,7 @@ pub(super) fn initialize(ctx: &Pkcs11, init_args: CInitializeArgs) -> Result<()>
         Rv::from(get_pkcs11!(ctx, C_Initialize)(
             init_args_ptr as *mut CK_C_INITIALIZE_ARGS as *mut ::std::ffi::c_void,
         ))
-        .into_result()
+        .into_result(Function::Initialize)
     }
 }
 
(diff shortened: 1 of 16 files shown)


cryptoki/src/error/mod.rs
```diff
@@ -131,11 +133,11 @@ impl From<std::convert::Infallible> for Error {
     }
 }
 
-impl From<RvError> for Error {
-    fn from(rv_error: RvError) -> Self {
-        Error::Pkcs11(rv_error)
-    }
-}
+// impl From<RvError> for Error {
```
wiktor-k, line 136: If it's not used I guess it can be removed.

cryptoki/src/context/general_purpose.rs
```diff
@@ -120,6 +121,81 @@ pub enum Function {
     WaitForSlotEvent,
 }
 
+impl Display for Function {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
```
wiktor-k, line 126: :thinking: quite a lot of code... is there a reason you didn't just use `write!(f, "Function::{:?}", self)`? (Maybe you didn't want to depend on the Debug trait's derive that just outputs variant's name?)

cryptoki/src/error/mod.rs
```diff
@@ -18,7 +20,7 @@ pub enum Error {
     LibraryLoading(libloading::Error),
 
     /// All PKCS#11 functions that return non-zero translate to this error.
-    Pkcs11(RvError),
+    Pkcs11(RvError, Function),
```
wiktor-k, line 23: This looks like an API break...? :thinking: (nothing against it just noting for version bump).

wiktor-k (approved):
LGTM :+1:

cryptoki/src/error/mod.rs
ionut-arm: Yeah, sorry, commented out when I was fixing things, forgot to delete it 😁

cryptoki/src/error/mod.rs
ionut-arm: Hmm, indeed. I was aware, but now thinking about it, maybe I can hide the `Function` in `RvError` instead. Will need to investigate

wiktor-k:
@ionut-arm  it seems this is largely OK to merge (after fixing conflicts). Is there anything else you'd like to add to it?

ionut-arm:
> @ionut-arm it seems this is largely OK to merge (after fixing conflicts). Is there anything else you'd like to add to it?

Ok, I've incorporated your feedback, I don't think I can easily avoid breaking the interface since we rely on enums for errors..

wiktor-k (approved):
Okay, I think this looks good :+1: I'm a bit worried about extending `Error` enum but if you bump version on release it should be fine.

Thanks for this change! I think it will make debugging quite a bit easier!

gowthamsk-arm (approved):
Apologies for the delay. LGTM :)

PR #149, opened 2023-06-07
Include AES_CBC_ENCRYPT_DATA mechanism

chrisFrodo:
Hi,
I noticed AES_CBC_ENCRYPT_DATA and it's param structure AES_CBC_ENCRYPT_DATA_PARAMS were not included in the Mechanism enum.
Using other implementation codebase, i managed to include it.
I tested it using SoftHSM2.


wiktor-k (commented):
Not sure if SoftHSMv2 supports this mechanism but if it does it'd be good if we had tests for it. Otherwise this looks good 👍

wiktor-k:
Apparently this is already implemented through #178. Closing...

PR #152, opened 2023-06-17
"fix" for clone-then-initialize problem

arjennienhuis:
This PR fixes the problem with clone and initialize.

It adds more locking while PKCS11 is supposed to be thread safe. For me that doesn't feel right. I would deprecate or remove `is_initialized()`:

```rust
    #[deprecated]
    pub fn is_initialized(&self) -> bool {
        get_library_info(self).is_ok()
    }
```

### DIFF
--- a/cryptoki/src/context/mod.rs
+++ b/cryptoki/src/context/mod.rs
@@ -31,6 +31,7 @@ use std::mem;
 use std::path::Path;
 use std::ptr;
 use std::sync::Arc;
+use std::sync::RwLock;
 
 // Implementation of Pkcs11 class that can be enclosed in a single Arc
 pub(crate) struct Pkcs11Impl {
@@ -76,7 +77,7 @@ impl Drop for Pkcs11Impl {
 #[derive(Clone, Debug)]
 pub struct Pkcs11 {
     pub(crate) impl_: Arc<Pkcs11Impl>,
-    initialized: bool,
+    initialized: Arc<RwLock<bool>>,
 }
 
 impl Pkcs11 {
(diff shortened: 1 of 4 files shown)


ionut-arm (approved):
LGTM, thanks! 👍🏻

cryptoki/src/context/mod.rs
```diff
@@ -99,23 +100,23 @@ impl Pkcs11 {
...
     }
 
     /// Initialize the PKCS11 library
-    pub fn initialize(&mut self, init_args: CInitializeArgs) -> Result<()> {
-        if !self.initialized {
-            initialize(self, init_args)
-        } else {
-            Err(Error::AlreadyInitialized)
+    pub fn initialize(&self, init_args: CInitializeArgs) -> Result<()> {
+        let mut init_lock = self.initialized.as_ref().write().unwrap();
```
wiktor-k, line 110: This is super minor but I'd nonetheless prefer expect here with the reason `lock not to be poisoned` or something like that. I know it's rare but if initialize below panics it'd poison the lock and I'd rather see descriptive error than a generic one in the stacktrace.

cryptoki/tests/basic.rs
```diff
@@ -707,6 +703,35 @@ fn is_initialized_test() {
     }
 }
 
+#[test]
+#[serial]
+#[allow(deprecated)]
```
wiktor-k, line 708: Are these suppressions needed? I see nothing deprecated added and the clone warning should also not appear (just wanted to double check).

cryptoki/src/context/mod.rs
```diff
@@ -99,23 +100,23 @@ impl Pkcs11 {
...
+        if *init_lock {
+            Err(Error::AlreadyInitialized)?
         }
+        initialize(self, init_args).map(|_| *init_lock = true)
     }
 
     /// Check whether the PKCS11 library has been initialized
     pub fn is_initialized(&self) -> bool {
-        self.initialized
+        *self.initialized.as_ref().read().unwrap()
```
wiktor-k, line 119: Likewise for unwrap/expect.

wiktor-k (commented):
Looks good 👍

cryptoki/tests/basic.rs
arjennienhuis: The `deprecated` check can indeed be removed. The clone warning does trigger:

```
error: redundant clone
   --> cryptoki/tests/basic.rs:714:23
    |
714 |     let clone = pkcs11.clone();
    |                       ^^^^^^^^ help: remove this
    |
note: cloned value is neither consumed nor mutated
```

I only clone in that test to show the effect of cloning. Not to do anything useful with the clone.

cryptoki/src/context/mod.rs
arjennienhuis: done.

cryptoki/src/context/mod.rs
arjennienhuis: done.

wiktor-k (approved):
Looks good now 👌

PR #161, opened 2023-08-14
elliptic_curve: Fix broken references

wiktor-k:
I wanted to use some link for `ANSI X9.62` but didn't find anything useful to link to (I'm all ears if someone has a good idea) but for the time being I'm removing the broken markup.

One more thing is that the code references stuff in `cryptoki_sys` (like raw structure names) I wonder if it's a good idea (and technically possible) to link to them in the docs (the disadvantage is "leaking" raw types but I'm not sure it's a big deal in practice).

Edit: btw I'm surprised that `cargo rustdoc -- -D rustdoc::broken_intra_doc_links` doesn't complain :( it seems it only detects broken references if they don't contain a space.

### DIFF
--- a/cryptoki/src/mechanism/elliptic_curve.rs
+++ b/cryptoki/src/mechanism/elliptic_curve.rs
@@ -14,7 +14,7 @@ use std::ptr;
 /// ANSI X9.63, where each party contributes one key pair all using
 /// the same EC domain parameters.
 ///
-/// This structure wraps a CK_ECDH1_DERIVE_PARAMS structure.
+/// This structure wraps a `CK_ECDH1_DERIVE_PARAMS` structure.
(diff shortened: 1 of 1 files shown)


ionut-arm (approved):
I'm happy with the changes!

In terms of the ANSI standard, an equivalent thing that we could directly reference is SECG "SEC 1: Elliptic Curve Cryptography" - they seem to be referenced together in a lot of places (for example, see the ECDSA definition in the [PSA Cryptography API](https://arm-software.github.io/psa-api/crypto/1.1/IHI0086-PSA_Certified_Crypto_API-1.1.2.pdf)).

paulhowardarm (approved):
LGTM - thanks!

Issue #163, opened 2023-09-06
PKCS OAEP padding always returns: Pkcs11(ArgumentsBad)


probablynachi:
This OAEP code  complies but then throws the Pkcs11(ArgumentsBad) error for any parameters set. 
I am using softhsm2.
Here is the code
```rust
use cryptoki::context::{CInitializeArgs, Pkcs11};
use cryptoki::mechanism::{Mechanism, MechanismType};
use cryptoki::mechanism::rsa::{PkcsMgfType, PkcsOaepParams, PkcsOaepSource};
use cryptoki::object::Attribute;
use cryptoki::session::UserType;
use std::error::Error;
use cryptoki::types::AuthPin;
const PKCS11_LIB_PATH: &str = "PATH";
const SLOT_ID: u64 = SLOT;
fn main() -> Result<(), Box<dyn Error>> {
    let context = Pkcs11::new(PKCS11_LIB_PATH)?;
    context.initialize(CInitializeArgs::OsThreads)?;
    let session = context.open_rw_session(SLOT_ID.try_into()?)?;
    session.login(UserType::User, Some(&AuthPin::new("PIN".into()))).expect("Invalid PIN");
    let pub_key_template = [Attribute::ModulusBits(2048.into())];
    let (publickey, _privkey) = session.generate_key_pair(&Mechanism::RsaPkcsKeyPairGen,
                                                       &pub_key_template, &[])?;
    let oaep=PkcsOaepParams::new(MechanismType::SHA1 ,PkcsMgfType::MGF1_SHA1, PkcsOaepSource::empty());
    let encrypt_mechanism: Mechanism = Mechanism::RsaPkcsOaep(oaep);

    session.encrypt(&encrypt_mechanism, publickey, b"Test")?;
    Ok(())
}
```
Which results in the error: **Error: Pkcs11(ArgumentsBad).**
I tried the same code without using OAEP and it works as well, So i presume it is an issue with OAEP Padding.
Thank you

wiktor-k:
:thinking: It seems it should be supported in SoftHSM:

```sh
$ pkcs11-tool --modul /usr/lib64/pkcs11/libsofthsm2.so -M
Using slot 0 with a present token (0x63baf6de)
Supported mechanisms:
  AES-CBC, keySize={16,32}, encrypt, decrypt, wrap
...
  RSA-PKCS-KEY-PAIR-GEN, keySize={512,16384}, generate_key_pair
[...]

PR #165, opened 2023-09-12
Fix RSA OKCS OAEP mechanism

wiktor-k:
Handling of empty data specified was broken.

As specified in "2.1.7 PKCS #1 RSA OAEP mechanism parameters" the `CKZ_DATA_SPECIFIED` source:

> Array of CK_BYTE containing the value of the encoding parameter.
> If the parameter is empty, pSourceData must be NULL and ulSourceDataLen
> must be zero.

Adjusts by returning `std::ptr::null()` when the source data has not been specified (`PkcsOaepSource::empty()` was used) and adds a tests case for this.

One more test case that is ignored is added to handle the case when the data specified is non-empty. Unfortunately this test fails for unknown reasons (I haven't been able to find if SoftHSM supports it or not).

Fixes: https://github.com/parallaxsecond/rust-cryptoki/issues/163
Fixes: https://github.com/parallaxsecond/rust-cryptoki/pull/164


probablynachi:
Thank you so much for your quick response and fix!

wiktor-k:
> Thank you so much for your quick response and fix!

No problem. I surveyed existing alternative implementations and started changing stuff here and there and finally your test case succeeded. Thanks for submitting it as this was the critical factor that allowed me to quickly check various things.

(Note that it may take some time until it's reviewed and a new version of the crate is released to crates.io)

See you later! :wave:

hug-dev (approved):

cryptoki/tests/basic.rs
```diff
@@ -1028,6 +1029,57 @@ fn aes_gcm_with_aad() -> TestResult {
...
+    let session = pkcs11.open_rw_session(slot)?;
+    session.login(UserType::User, Some(&AuthPin::new(USER_PIN.into())))?;
+
+    let pub_key_template = [Attribute::ModulusBits(2048.into())];
+    let (pubkey, privkey) =
+        session.generate_key_pair(&Mechanism::RsaPkcsKeyPairGen, &pub_key_template, &[])?;
+    let oaep = PkcsOaepParams::new(
+        MechanismType::SHA1,
+        PkcsMgfType::MGF1_SHA1,
+        PkcsOaepSource::data_specified(&[1, 2, 3, 4, 5, 6, 7, 8]),
```
ionut-arm, line 1071: We've had a similar problem in Parsec with the software TPM, where the salt/label would not be accepted unless it ends with a `0` byte. Maybe you can try and see if that helps?

ionut-arm (approved):
Thank you for the fix!

cryptoki/tests/basic.rs
wiktor-k: Hmm... I did some testing and browsed softhsm's source code and apparently they don't support any source data that is not null :/ (I'm on mobile now and will send links later).

Not sure if in that case it's better to remove the test case or leave it so that at least it's compile checked...?

ionut-arm: I think we can leave it for now! Good to know that we at least have it, and why it doesn't work on CI.

wiktor-k: Oh, it seems I forgot to update this PR with my findings. :sweat_smile: 

Here's a line of code in softhsm that checks `pSourceData` and returns `CKR_ARGUMENTS_BAD` if it is set (!= `NULL`):

https://github.com/opendnssec/SoftHSMv2/blob/f4661af6680e187f220921c0a91d17d71a680345/src/lib/SoftHSM.cpp#L12850

So, currently this is untestable with softhsm. Apparently it *is* supported by yubico library (ykcs11): https://github.com/Yubico/yubico-piv-tool/blob/a72b2054de4b8ee0bf534edadd4de64d0ded92a9/ykcs11/mechanisms.c#L983 I'll add testing that on a yubico to my ever-growing list of tasks to do :)

Thanks for the review and merging :wave:

PR #178, opened 2023-10-10
Adding 2 new mechanisms: generic key generation and key derivation via encryption

Nk185:
Added CKM_AES_CBC_ENCRYPT_DATA;
Added CKM_GENERIC_SECRET_KEY_GEN;
Added prettier `MechanismInfo` fmt.

### DIFF
--- a/cryptoki/src/mechanism/ekdf.rs
+++ b/cryptoki/src/mechanism/ekdf.rs
@@ -1,0 +1,55 @@
+// Copyright 2021 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+//! Mechanisms of key derivation by data encryption
+//! See: https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203514
+
+use std::{convert::TryInto, marker::PhantomData, slice};
+
+/// AES CBC derivation parameters.
+///
+/// The mechanisms will function by performing the encryption over the data provided using the base
+/// key. The resulting cipher text shall be used to create the key value of the resulting key.
+///
+/// This structure wraps a `CK_AES_CBC_ENCRYPT_DATA_PARAMS` structure.
+#[derive(Debug, Clone, Copy)]
+#[repr(transparent)]
+pub struct AesCbcDeriveParams<'a> {
+    inner: cryptoki_sys::CK_AES_CBC_ENCRYPT_DATA_PARAMS,
+
+    /// Marker type to ensure we don't outlive the data
+    _marker: PhantomData<&'a [u8]>,
+}
+
+impl<'a> AesCbcDeriveParams<'a> {
+    /// Construct parameters for key derivation via encryption (EKDF).
+    ///
+    /// # Arguments
+    ///
+    /// * `iv` - The initialization vector
+    ///
+    /// * `data` - Data that will be encryption with the base key to obtain
+    /// the new key from the resulted cypher.
+    pub fn new(iv: [u8; 16], data: &'a [u8]) -> Self {
+        Self {
+            inner: cryptoki_sys::CK_AES_CBC_ENCRYPT_DATA_PARAMS {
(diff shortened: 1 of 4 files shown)


Nk185:
@wiktor-k 
The two mechanisms are surely supported in v2.6.1:
* The generic secret: [here](https://github.com/opendnssec/SoftHSMv2/blob/2.6.1/src/lib/SoftHSM.cpp#L7320)
* The derivation via encryption: [here](https://github.com/opendnssec/SoftHSMv2/blob/2.6.1/src/lib/SoftHSM.cpp#L7128)

Sure, I will add tests over weekends, maybe earlier... How would you like those tests for the EKDF? Should it be a simple derive that makes sure a new symmetric key is created? Or should I also check that an expected key was generated?

wiktor-k:
> Should it be a simple derive that makes sure a new symmetric key is created? Or should I also check that an expected key was generated?

Either is fine for me and I guess we have both variants in existing tests 😅

Thanks for taking care of it!

Nk185:
@wiktor-k 
Well, apparently, I found some time earlier than I thought 😅 
Tests added :)

cryptoki/src/mechanism/mechanism_info.rs
```diff
@@ -219,6 +218,26 @@ impl MechanismInfo {
     }
 }
 
+impl std::fmt::Display for MechanismInfo {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let flags = format!("{0:#?}", self.flags);
+        let key_size_info = match self.min_key_size == 0 && self.max_key_size == 0 {
```
wiktor-k, line 224: I think this could be shortened to just `if` instead of `match` with `true`/`false` arms.

cryptoki/src/mechanism/mechanism_info.rs
```diff
@@ -219,6 +218,26 @@ impl MechanismInfo {
...
+                    format!(", min_key_size={}", self.min_key_size)
+                } else {
+                    format!(
+                        ", min_key_size={}, max_key_size={}",
+                        self.min_key_size, self.max_key_size
+                    )
+                }
+            }
+        };
+        write!(f, "{}{}", flags, key_size_info)
```
wiktor-k, line 237: :thinking: I think it's a bit more complex than necessary: first `flags` are always written first so instead of `let flags = format!("{0:#?}", self.flags);` just write them directly: `write!(f, "{:#?}", self.flags);` and then you can have the additional constraint and the function should be roughly:

```rust
write!(f, "{:#?}", self.flags);
if !(self.min_key_size == 0 && self.max_key_size == 0) {

  write!(f, ", min_key_size={}", self.min_key_size);

  if self.max_size != 0 {
    write!(f, ", max_key_size={}", self.max_key_size);
  }
}
```

With no additional allocations (format! and String allocates)... phew... I hope it makes sense :sweat_smile: 

Since this is a bit involved I'm wondering if a simple test would also be in order :thinking:

wiktor-k (approved):
Okay Nikita, I'm super happy with the PR, it looks high-quality :+1: 

I've added a couple of small "improvements" but I'll approve it already.

Thanks for your time and contributing to cryptoki crate! :bow:

cryptoki/src/mechanism/mechanism_info.rs
Nk185: updated :)

wiktor-k (approved):
LGTM :+1: thanks!

cryptoki/src/mechanism/ekdf.rs
```diff
@@ -0,0 +1,55 @@
+// Copyright 2021 Contributors to the Parsec project.
```
ionut-arm, line 1:
```suggestion
// Copyright 2023 Contributors to the Parsec project.
```

ionut-arm (approved):
Thanks, looks good to me! :)

wiktor-k (approved):

PR #196, opened 2024-01-02
Fix clippy warnings introduced with rust 1.75

baloo:
This would raise errors in CI after the bump to rustc 1.75


hug-dev (approved):

gowthamsk-arm:
#197 has the same changes. Didn't notice this PR. :(

PR #200, opened 2024-02-21
Add support for `CARGO_TARGET_DIR` in `regenerate_bindings.sh`

wiktor-k:
The script to regenerate bindings hardcoded `../target` path. This did not work if the user customized the target directory location through the `CARGO_TARGET_DIR` environment variable.

Fix the issue by first reading the variable and default to `../target` if it is unset.


hug-dev (approved):
:bow:

ionut-arm (approved):
Thanks!!

PR #202, opened 2024-03-07
Fix errors linked to psa-crypto-conversions feature

gowthamsk-arm:
The CI script so far was not testing the psa-crypto-conversions feature and the code bounded by this feature was not updated with changes to the structures in the crate resulting in errors. This patch fixes these errors.

Signed-off-by: Gowtham Suresh Kumar <gowtham.sureshkumar@arm.com>

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -9,6 +9,8 @@ set -euf -o pipefail
 
 RUST_BACKTRACE=1 cargo build
 
+RUST_BACKTRACE=1 cargo build --all-features
+
 RUST_BACKTRACE=1 cargo build --target arm-unknown-linux-gnueabi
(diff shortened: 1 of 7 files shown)


ci.sh
```diff
@@ -33,3 +33,5 @@ if cargo clippy --version; then
 fi
 
 RUST_BACKTRACE=1 cargo test
+
+RUST_BACKTRACE=1 cargo build --all-features
```
tgonzalezorlandoarm, line 37: Should we put it higher up with the rest of the build tests?

tgonzalezorlandoarm (changes requested):
small change

wiktor-k (approved):
LGTM 👍 Thanks!

tgonzalezorlandoarm (approved):

PR #203, opened 2024-03-08
20240308 mechanisms

Firstyear:
While testing this library with the tpm2_pkcs11 provided I noticed a number of mechanisms that were unable to be displayed.


wiktor-k (approved):
:thinking: Well, generally when adding new mechanisms we ask for tests cases showing that they work against SoftHSM but the majority of this change is adding stuff that already exists and for the rest I'm not sure there's much value in testing HMAC values...

Please squash your commits and add a sign-off and for me it's good to go. :+1:

Firstyear:
Should be there now, I think it just needed a rebase on main

hug-dev (approved):

PR #213, opened 2024-08-09
Define CKD_SHA256_KDF transformation

freedge:
Define CKD_SHA256_KDF transformation to be used with CKM_ECDH1_DERIVE.

Some HSM with FIPS restriction will refuse to derive keys with CKD_NULL. CKD_SHA256_KDF will do fine though.

Unfortunately this is not implemented on softHSM
(https://github.com/opendnssec/SoftHSMv2/pull/599) so I provide no test. This was tested fine against Thales DPOD.

### DIFF
--- a/cryptoki/src/mechanism/elliptic_curve.rs
+++ b/cryptoki/src/mechanism/elliptic_curve.rs
@@ -93,6 +93,17 @@ impl<'a> EcKdf<'a> {
         }
     }
 
+    /// The sha256 transformation as defined in the x9 standard. The
+    /// derived key is produced by concatenating hashes of the shared
+    /// value followed by 00000001, 00000002, etc. until we find
+    /// enough bytes to fill the `CKA_VALUE_LEN` of the derived key.
+    pub fn sha256_x9() -> Self {
+        Self {
+            kdf_type: CKD_SHA256_KDF,
+            shared_data: None,
+        }
+    }
+
     // The intention here is to be able to support other methods with
     // shared data, without it being a breaking change, by just adding
     // additional constructors here.


cryptoki/src/mechanism/elliptic_curve.rs
```diff
@@ -93,6 +93,17 @@ impl<'a> EcKdf<'a> {
         }
     }
 
+    /// The sha256 transformation as defined in the x9 standard. The
+    /// derived key is produced by concatenating hashes of the shared
+    /// value followed by 00000001, 00000002, etc. until we find
+    /// enough bytes to feel the CKA_VALUE_LEN of the derived key.
```
wiktor-k, line 99:
```suggestion
    /// enough bytes to fill the `CKA_VALUE_LEN` of the derived key.
```

wiktor-k:
I think it looks good. 👍 

Unfortunately due to a new Rust version the lints started to pop up.

We could fix them in a similar way as in the tpm repo. What do you think @ionut-arm ?

wiktor-k (dismissed):

keldonin:
@freedge if that helps, the CI bot seems to work now, after the merge of PR #218. You might want to sync up your branch on your repo with the upstream one, to incorporate the latest fixes and get through these issues. I experienced the same and that solved the Execute CI script failures.

cryptoki/src/mechanism/elliptic_curve.rs
```diff
@@ -93,6 +93,17 @@ impl<'a> EcKdf<'a> {
         }
     }
 
+    /// The sha256 transformation as defined in the x9 standard. The
```
ionut-arm, line 96:
```suggestion
    /// The key derivation function based on sha256 as defined in the ANSI X9.63 standard. The
```
I'd recommend this change to make the reference clearer. If you'd like to put in a link to the standard that'd be even nicer 😉

cryptoki/src/mechanism/elliptic_curve.rs
```diff
@@ -93,6 +93,17 @@ impl<'a> EcKdf<'a> {
...
+    /// derived key is produced by concatenating hashes of the shared
+    /// value followed by 00000001, 00000002, etc. until we find
+    /// enough bytes to fill the `CKA_VALUE_LEN` of the derived key.
+    pub fn sha256_x9() -> Self {
```
ionut-arm, line 100:
```suggestion
    pub fn sha256() -> Self {
```
I think SHA256 is actually specific enough here, since the type makes it clear it's a KDF.

ionut-arm (commented):
Thanks for the patch!

hug-dev:
If you rebase and address @ionut-arm 's comments here we can easily get it merged :) !

freedge:
let's close that for the moment as I don't have time or interest to work on it.
thanks for your project though!

hug-dev:
Understood! Was a shame to not merge it after all your effort so pushed the same + the fix [here](https://github.com/parallaxsecond/rust-cryptoki/pull/239), hopefully we can get it merged soon and close the corresponding issue :)

PR #214, opened 2024-08-27
Session.find_objects(): implement fewer calls to C_FindObjects()

keldonin:
This PR contains code that optimizes `Session.find_objects()`. It essentially quit the loop as soon as less than `MAX_OBJECT_COUNT` have been returned by `C_FindObjects()`. 

It typically results in calling one less time the underlying API `C_FindObjects()`.

Testing code included.

### DIFF
--- a/cryptoki/src/session/object_management.rs
+++ b/cryptoki/src/session/object_management.rs
@@ -31,26 +31,7 @@ impl Session {
         let mut object_count = 0;
         let mut objects = Vec::new();
 
-        unsafe {
-            Rv::from(get_pkcs11!(self.client(), C_FindObjects)(
-                self.handle(),
-                object_handles.as_mut_ptr() as CK_OBJECT_HANDLE_PTR,
-                MAX_OBJECT_COUNT.try_into()?,
-                &mut object_count,
-            ))
-            .into_result(Function::FindObjects)?;
-        }
-
-        while object_count > 0 {
-            objects.extend_from_slice(&object_handles[..object_count.try_into()?]);
-
-            // If the returned slice is not full, there are no pending object handles to be returned.
-            // In which case, exit loop.
-            // This avoids, in many situations, an unnecessary API call with 0 object handles returned.
-            if (object_count as usize) < MAX_OBJECT_COUNT {
-                break;
-            }
-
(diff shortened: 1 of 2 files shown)


cryptoki/src/session/object_management.rs
```diff
@@ -44,6 +44,13 @@ impl Session {
         while object_count > 0 {
```
hug-dev, line 44: could it work change this instead to:
```rust
while object_count == MAX_OBJECT_COUNT {
```
?

hug-dev (dismissed):

cryptoki/src/session/object_management.rs
keldonin: Not really, as it would exit the loop directly and not extend the resulting array with the last result from `C_FindObjects()`. It would result in a truncated, incomplete result.

I have changed the loop and removed one unsafe `C_FindObjects` section, see if you find it suitable. Works for me (tests are passing as expected).

keldonin: I think you were correct, actually. Please consider the new commit anyway, It should work as well and reduces code duplication.

cryptoki/src/session/object_management.rs
```diff
@@ -31,19 +31,7 @@ impl Session {
...
-                MAX_OBJECT_COUNT.try_into()?,
-                &mut object_count,
-            ))
-            .into_result(Function::FindObjects)?;
-        }
-
-        while object_count > 0 {
-            objects.extend_from_slice(&object_handles[..object_count.try_into()?]);
-
+        loop {
```
hug-dev, line 34: Good idea to combine both calls in the same loop! I am though generally wary of `loop` though, that in some unexpected data path (now or in a future change) we might end up in an infinite loop.. Is there a way to do the same but having a `while` condition instead? For example by initializing `object_count` to something else than 0?

keldonin: adjusted.

hug-dev (approved):

wiktor-k (approved):
Yup, this looks similar to the C examples I've seen in the wild.

Thanks for adding the tests too! :+1:

PR #215, opened 2024-08-28
Implement Session.copy_object()

keldonin:
This PR implements `C_CopyObject()`. It takes a template as an optional argument (i.e. `Option<&[Attribute]>`).

The PR comes along with rustdoc and dedicated test case.

### DIFF
--- a/cryptoki/src/session/object_management.rs
+++ b/cryptoki/src/session/object_management.rs
@@ -97,12 +97,14 @@ impl Session {
     }
 
     /// Copy an object
+    ///
     /// A template can be provided to change some attributes of the new object, when allowed.
     ///
     /// # Arguments
     ///
     /// * `object` - The [ObjectHandle] used to reference the object to copy
-    /// * `template` - a reference to a slice of attributes
+    /// * `template` - new values for any attributes of the object that can ordinarily be modified
+    ///   check out [PKCS#11 documentation](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/cs01/pkcs11-spec-v3.1-cs01.html#_Toc111203284) for details
     ///
     /// # Returns
     ///
@@ -124,7 +126,7 @@ impl Session {
                 template.len().try_into()?,
                 &mut object_handle as CK_OBJECT_HANDLE_PTR,
             ))
-            .into_result(Function::CreateObject)?;
+            .into_result(Function::CopyObject)?;
         }
 
         Ok(ObjectHandle::new(object_handle))
--- a/cryptoki/src/session/object_management.rs
+++ b/cryptoki/src/session/object_management.rs
@@ -97,24 +97,19 @@ impl Session {
     }
 
     /// Copy an object
-    /// A optional template can be provided to change some attributes of the new object, when allowed.
+    /// A template can be provided to change some attributes of the new object, when allowed.
     ///
     /// # Arguments
     ///
     /// * `object` - The [ObjectHandle] used to reference the object to copy
-    /// * `template` - an optional reference to a slice of attributes, in the form of `Some(&[Attribute])`, 
-    ///    or set to `None`` if not needed.
+    /// * `template` - a reference to a slice of attributes
     ///
     /// # Returns
     ///
     /// This function will return a new [ObjectHandle] that references the newly created object.
     ///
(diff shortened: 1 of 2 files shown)


cryptoki/src/session/object_management.rs
```diff
@@ -96,6 +96,41 @@ impl Session {
...
+    ///
+    /// # Returns
+    ///
+    /// This function will return a new [ObjectHandle] that references the newly created object.
+    ///
+    pub fn copy_object(&self, object: ObjectHandle, template: Option<&[Attribute]>) -> Result<ObjectHandle> {
+        let mut template: Vec<CK_ATTRIBUTE> = match template {
+            Some(template) => template.iter().map(|attr| attr.into()).collect(),
+            None => Vec::new(),
+        };
```
hug-dev, line 116: Maybe you can do what we do in `find_objects` and not use the `Option`:
```rust
let mut template: Vec<CK_ATTRIBUTE> = template.iter().map(|attr| attr.into()).collect();
```

if it works :detective: !

keldonin: I found using Option<> to be more idiomatic. OTOH there is maybe little reason for not providing a template, when doing a copy. I'll adapt the PR if this is preferred.

keldonin: Sure thing. Let me adapt the code.

hug-dev (approved):
Thank you!

keldonin:
The CI script fails for the reason described in #216, can this be looked at? I guess it will block all PRs going forward.

Thank you!

hug-dev (approved):

cryptoki/src/session/object_management.rs
```diff
@@ -96,6 +96,40 @@ impl Session {
         }
     }
 
+    /// Copy an object
+    /// A template can be provided to change some attributes of the new object, when allowed.
```
wiktor-k, line 100: Please insert a new line here:

```suggestion
    ///
    /// A template can be provided to change some attributes of the new object, when allowed.
```

cryptoki/src/session/object_management.rs
```diff
@@ -96,6 +96,40 @@ impl Session {
...
+
+        unsafe {
+            Rv::from(get_pkcs11!(self.client(), C_CopyObject)(
+                self.handle(),
+                object.handle(),
+                template.as_mut_ptr(),
+                template.len().try_into()?,
+                &mut object_handle as CK_OBJECT_HANDLE_PTR,
+            ))
+            .into_result(Function::CreateObject)?;
```
wiktor-k, line 127: I think we need `CopyObject` here:

```suggestion
            .into_result(Function::CopyObject)?;
```

(this may need further changes in `Function` to introduce new variant if it's not there)

cryptoki/src/session/object_management.rs
```diff
@@ -96,6 +96,40 @@ impl Session {
         }
     }
 
+    /// Copy an object
+    /// A template can be provided to change some attributes of the new object, when allowed.
+    ///
+    /// # Arguments
+    ///
+    /// * `object` - The [ObjectHandle] used to reference the object to copy
+    /// * `template` - a reference to a slice of attributes
```
wiktor-k, line 105: The docs need to be improved here as it's not clear what the template is for. Random docs I found on the net clarify that:

```suggestion
    /// * `template` - new values for any attributes of the object that can ordinarily be modified
```

From: https://www.cryptsoft.com/pkcs11doc/v220/pkcs11__all_8h.html#aC_CopyObject

(Feel free to improve upon my suggestion!)

wiktor-k (commented):
It's quite good :+1: . A couple of nitpicks :)

cryptoki/src/session/object_management.rs
keldonin: good catch!

cryptoki/src/session/object_management.rs
keldonin: I made an update containing your suggestion. I also added a ref to the PKCS#11 spec. That looks good for me when generating the documentation, check if that suits your taste.

wiktor-k (approved):
Yep. LGTM 👍 

I just noticed that the template info is explained in the second sentence of the function doc 🤦‍♂️

😅

wiktor-k (approved):
LGTM 👍 Thanks!

hug-dev (approved):

PR #220, opened 2024-09-03
Support vendor extensions for CK_USER_TYPE

larper2axis:
Some HSM vendors have more user types than the standard CKU_SO and CKU_USER. For example a Thales Luna uses the value 0x80000001 to represent the unprivileged Crypto User.


wiktor-k (dismissed):
Looks pretty straightforward 👍

hug-dev (approved):

PR #223, opened 2024-09-04
implements session object handle iterator, with caching

keldonin:
Hello, 

This PR is an alternate candidate for implementing #106.

It implements the Iterator trait for object handlers, given a search template. The iterator is able to fetch values ahead, and cache the results. The cache size can be specified at iterator build time.

## design choices

- The iterator will return an `Option<Result<ObjectHandle>>`. This is because calls to `C_FindObjects()` may fail, and the caller may want to manage such condition. It is similar to what the  [`std::io::Lines`](https://doc.rust-lang.org/std/io/struct.Lines.html#method.next) iterator implements.
- I didn't replace the `find_objects()` method. It seems to me it serves a different goal,  ease of use. On the other hand, the iterator allows the caller a more fine-grained control over how object handles are retrieved from the token, and allows for streaming. Both can coexist, IMO.
- as with the recent `find_objects` change, every effort is made to avoid unnecessary calls to `C_FindObjects()`. 
- The iterator implements the `Drop` trait, to call `C_FindObjectsFinal()`.
- Thanks to the design choices made for the Session object, the iterator features no locking mechanism for its state variables. It assumes that the iterator object will never be accessed concurrently from different threads.
-  No code was added to prevent creating two iterators: It relies upon the underlying token to prevent running two searches simultaneously.
- Because the `next()` method of an iterator cannot return a `Result`, I had to create a new macro from `get_pkcs11!()` that does not carry the syntactic sugar `?`.
- The implementation tries to address the unlikely case when `C_FindObjects` and `C_FindObjectsFinal` are not implemented. But... perhaps the library should refuse to work with such library, and detect that at library loading time? 

For your review,

### DIFF
--- a/cryptoki/src/context/mod.rs
+++ b/cryptoki/src/context/mod.rs
@@ -13,6 +13,14 @@ macro_rules! get_pkcs11 {
     };
 }
 
+/// Same as get_pkcs11! but does not attempt to apply '?' syntactic sugar.
+/// Suitable only if the caller can't return a Result.
+macro_rules! get_pkcs11_func {
+    ($pkcs11:expr, $func_name:ident) => {
+        ($pkcs11.impl_.function_list.$func_name)
+    };
+}
+
 mod general_purpose;
 mod info;
 mod locking;
--- a/cryptoki/src/context/mod.rs
+++ b/cryptoki/src/context/mod.rs
@@ -3,13 +3,10 @@
 //! Pkcs11 context and initialization types
 
 /// Directly get the PKCS #11 operation from the context structure and check for null pointers.
+/// Note that this macro depends on the get_pkcs11_func! macro.
 macro_rules! get_pkcs11 {
     ($pkcs11:expr, $func_name:ident) => {
-        ($pkcs11
-            .impl_
-            .function_list
-            .$func_name
-            .ok_or(crate::error::Error::NullFunctionPointer)?)
+        (get_pkcs11_func!($pkcs11, $func_name).ok_or(crate::error::Error::NullFunctionPointer)?)
     };
 }
 
--- a/cryptoki/src/session/object_management.rs
+++ b/cryptoki/src/session/object_management.rs
@@ -12,7 +12,8 @@ use std::convert::TryInto;
 use std::num::NonZeroUsize;
 
 // Search 10 elements at a time
-const MAX_OBJECT_COUNT: usize = 10;
+// Safety: the value provided (10) must be non-zero
+const MAX_OBJECT_COUNT: NonZeroUsize = unsafe { NonZeroUsize::new_unchecked(10) };
 
 /// Iterator over object handles, in an active session.
 ///
@@ -37,16 +38,16 @@ const MAX_OBJECT_COUNT: usize = 10;
 /// use std::env;
 ///
 /// # fn main() -> testresult::TestResult {
-/// let pkcs11 = Pkcs11::new(
-///     env::var("PKCS11_SOFTHSM2_MODULE")
-///         .unwrap_or_else(|_| "/usr/local/lib/libsofthsm2.so".to_string()),
-/// )?;
-///
-/// pkcs11.initialize(CInitializeArgs::OsThreads)?;
-/// let slot = pkcs11.get_slots_with_token()?.remove(0);
-///
-/// let session = pkcs11.open_ro_session(slot).unwrap();
-/// session.login(UserType::User, Some(&AuthPin::new("fedcba".into())))?;
+/// # let pkcs11 = Pkcs11::new(
+/// #    env::var("PKCS11_SOFTHSM2_MODULE")
+/// #        .unwrap_or_else(|_| "/usr/local/lib/libsofthsm2.so".to_string()),
+/// # )?;
+/// #
+/// # pkcs11.initialize(CInitializeArgs::OsThreads)?;
+/// # let slot = pkcs11.get_slots_with_token()?.remove(0);
+/// #
+/// # let session = pkcs11.open_ro_session(slot).unwrap();
+/// # session.login(UserType::User, Some(&AuthPin::new("fedcba".into())))?;
 ///
 /// let token_object = vec![Attribute::Token(true)];
 /// let wanted_attr = vec![AttributeType::Label];
@@ -84,6 +85,28 @@ pub struct ObjectHandleIterator<'a> {
 }
 
 impl<'a> ObjectHandleIterator<'a> {
+    /// Create a new iterator over object handles.
+    ///
+    /// # Arguments
+    ///
+    /// * `session` - The session to iterate over
+    /// * `template` - The template to match objects against
+    /// * `cache_size` - The number of objects to cache (type is [`NonZeroUsize`])
+    ///
+    /// # Returns
+    ///
+    /// This function will return a [`Result<ObjectHandleIterator>`] that can be used to iterate over the objects
+    /// matching the template. The cache size corresponds to the size of the array provided to `C_FindObjects()`.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the call to `C_FindObjectsInit` fails.
+    ///
+    /// # Note
+    ///
+    /// The iterator `new()` method will call `C_FindObjectsInit`. It means that until the iterator is dropped,
+    /// creating another iterator will result in an error (typically `RvError::OperationActive` ).
+    ///
     fn new(
         session: &'a Session,
         mut template: Vec<CK_ATTRIBUTE>,
@@ -171,11 +194,15 @@ impl<'a> Iterator for ObjectHandleIterator<'a> {
 
 impl Drop for ObjectHandleIterator<'_> {
     fn drop(&mut self) {
-        // bark but pass if C_FindObjectsFinal() is not implemented
         if let Some(f) = get_pkcs11_func!(self.session.client(), C_FindObjectsFinal) {
+            // swallow the return value, as we can't do anything about it,
+            // but log the error
+            if let Rv::Error(error) = Rv::from(unsafe { f(self.session.handle()) }) {
+                log::error!("C_FindObjectsFinal() failed with error: {:?}", error);
+            }
+        } else {
+            // bark but pass if C_FindObjectsFinal() is not implemented
             log::error!("C_FindObjectsFinal() is not implemented on this library");
-            // swallow the return value, as we can't do anything about it
-            let _ = unsafe { f(self.session.handle()) };
         }
     }
 }
@@ -198,7 +225,7 @@ impl Session {
     /// * [`Session::iter_objects_with_cache_size`] for a way to specify the cache size
     #[inline(always)]
     pub fn iter_objects(&self, template: &[Attribute]) -> Result<ObjectHandleIterator> {
-        self.iter_objects_with_cache_size(template, NonZeroUsize::new(MAX_OBJECT_COUNT).unwrap())
+        self.iter_objects_with_cache_size(template, MAX_OBJECT_COUNT)
     }
 
     /// Iterate over session objects matching a template, with cache size
@@ -279,8 +306,7 @@ impl Session {
     ///
     #[inline(always)]
     pub fn find_objects(&self, template: &[Attribute]) -> Result<Vec<ObjectHandle>> {
-        self.iter_objects(template)?
-            .collect::<Result<Vec<ObjectHandle>>>()
+        self.iter_objects(template)?.collect()
     }
 
     /// Create a new object
--- a/cryptoki/src/session/object_management.rs
+++ b/cryptoki/src/session/object_management.rs
@@ -9,6 +9,7 @@ use crate::session::Session;
 use cryptoki_sys::*;
 use std::collections::HashMap;
 use std::convert::TryInto;
+use std::num::NonZeroUsize;
 
 // Search 10 elements at a time
 const MAX_OBJECT_COUNT: usize = 10;
@@ -86,12 +87,8 @@ impl<'a> ObjectHandleIterator<'a> {
     fn new(
         session: &'a Session,
         mut template: Vec<CK_ATTRIBUTE>,
-        cache_size: usize,
+        cache_size: NonZeroUsize,
     ) -> Result<Self> {
-        if cache_size == 0 {
-            return Err(Error::InvalidValue);
-        }
-
         unsafe {
             Rv::from(get_pkcs11!(session.client(), C_FindObjectsInit)(
                 session.handle(),
@@ -101,11 +98,11 @@ impl<'a> ObjectHandleIterator<'a> {
             .into_result(Function::FindObjectsInit)?;
         }
 
-        let cache: Vec<CK_OBJECT_HANDLE> = vec![0; cache_size];
+        let cache: Vec<CK_OBJECT_HANDLE> = vec![0; cache_size.get()];
         Ok(ObjectHandleIterator {
             session,
-            object_count: cache_size,
-            index: cache_size,
+            object_count: cache_size.get(),
+            index: cache_size.get(),
             cache,
         })
     }
@@ -187,6 +184,7 @@ impl Session {
     /// Iterate over session objects matching a template.
     ///
     /// # Arguments
+    ///
     /// * `template` - The template to match objects against
     ///
     /// # Returns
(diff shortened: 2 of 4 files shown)


cryptoki/tests/basic.rs
```diff
@@ -364,6 +366,100 @@ fn session_find_objects() {
     assert_eq!(found_keys.len(), 9);
 }
 
+#[test]
+#[serial]
+fn session_objecthandle_iterator() {
```
wiktor-k, line 371:
```suggestion
fn session_objecthandle_iterator() -> testresult::TestResult {
```

(and `Ok(())` at the end)

And then (almost) all `unwrap`s can just be converted to `?`.

cryptoki/src/session/object_management.rs
```diff
@@ -13,7 +13,217 @@ use std::convert::TryInto;
...
+    ///
+    /// # See also
+    /// * [`ObjectHandleIterator`] for more information on how to use the iterator
+    /// * [`Session::iter_objects`] for a simpler way to iterate over objects
+    pub fn iter_objects_with_cache_size(
+        &self,
+        template: &[Attribute],
+        cache_size: usize,
+    ) -> Result<ObjectHandleIterator> {
+        let template: Vec<CK_ATTRIBUTE> = template.iter().map(|attr| attr.into()).collect();
```
wiktor-k, line 223:
```suggestion
        let template: Vec<CK_ATTRIBUTE> = template.iter().map(Into::into).collect();
```

I think this would be shorter :thinking:

wiktor-k (approved):
Okay, I think it looks very nice :+1: 

I've added a couple of personal, tiny nitpicks.

I also wonder if it'd be a good idea to add `Closes: https://github.com/parallaxsecond/rust-cryptoki/pull/109` in the commit message trailer (since it appears to be the same functionality but with a separate method).

One further improvement would be re-implementing `find_objects` in terms of your `iter_objects` (it should be as simple as just calling `iter_objects` and then `collect` on the result :thinking: )

Thanks for your time! :wave:

wiktor-k:
> No code was added to prevent creating two iterators: It relies upon the underlying token to prevent running two searches simultaneously.

Correct me if I'm wrong but this should be as simple as making `iter_objects_with_cache_size` take a `mut` ref to Session. This way as long as the `ObjectHandleIterator` exists (which would hold a `mut` ref to Session no other iterators could be created (since `mut` means "exclusive"). :thinking:

keldonin:
> > No code was added to prevent creating two iterators: It relies upon the underlying token to prevent running two searches simultaneously.
> 
> Correct me if I'm wrong but this should be as simple as making `iter_objects_with_cache_size` take a `mut` ref to Session. This way as long as the `ObjectHandleIterator` exists (which would hold a `mut` ref to Session no other iterators could be created (since `mut` means "exclusive"). 🤔

That would work. However, if I combine this with a rewrite of `find_objects()` as suggested above, it will propagate the need for a mutable ref, leading to a breaking change for that API: 

```
pub fn find_objects(&mut self, template: &[Attribute]) -> Result<Vec<ObjectHandle>>`
```

Choose your curse 😉

keldonin:
Actually there is maybe a good reason for not taking a mutable ref:

```
    for obj in session.iter_objects(&token_object)? {
        let obj = obj?; 

        let attributes = session.get_attributes(obj, &wanted_attr)?;
        ...
```

In this pattern, attributes are retrieved as objects are fetched. It results in interleaved calls to `C_FindObjects()` and `C_GetAttributeValue()`, which is legit. In fact, crypto operations are also legit during a search. implementing session with a mutable ref would prevent this, ans session.get_attributes() is also needing a ref to the session.


If preventing concurrent searches by the library itself is desirable, it can be easily implemented using a flag in session, inside a `Cell` for interior mutability. 

Any thought?

wiktor-k:
Yeah, all good points. Maybe we could just leave the "no multiple open searches" invariant in docs and that's it. Making it too complex to cover a rare API misuse may not be such a great idea after all :thinking: 

(I've got a couple of complex ideas how to "solve" it but I worry that the cure will be worse than the disease :sweat_smile: )

cryptoki/src/session/object_management.rs
```diff
@@ -13,7 +13,217 @@ use std::convert::TryInto;
...
+/// use cryptoki::context::CInitializeArgs;
+/// use cryptoki::context::Pkcs11;
+/// use cryptoki::error::Error;
+/// use cryptoki::object::Attribute;
+/// use cryptoki::object::AttributeType;
+/// use cryptoki::session::UserType;
+/// use cryptoki::types::AuthPin;
+/// use std::env;
+///
+/// fn test() -> Result<(), Error> {
```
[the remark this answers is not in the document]
keldonin: I'm sorry, but I'm unclear on this one. What is the change about? and what about the trailing `main()` at the end of the example?

cryptoki/src/session/object_management.rs
```diff
@@ -13,7 +13,217 @@ use std::convert::TryInto;
...
+    object_count: usize,
+    index: usize,
+    cache: Vec<CK_OBJECT_HANDLE>,
+}
+
+impl<'a> ObjectHandleIterator<'a> {
+    fn new(
+        session: &'a Session,
+        mut template: Vec<CK_ATTRIBUTE>,
+        cache_size: usize,
```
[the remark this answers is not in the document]
keldonin: Not sure, TBH. I think this will grow verbosity. I couldn't spot any obvious benefit, unless I missed something?

cryptoki/tests/basic.rs
keldonin: That's a good one!

cryptoki/src/session/object_management.rs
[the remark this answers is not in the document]
keldonin: OK I get it now.

keldonin:
> I also wonder if it'd be a good idea to add `Closes: https://github.com/parallaxsecond/rust-cryptoki/pull/109` in the commit message trailer (since it appears to be the same functionality but with a separate method).

When all the reviews are finished, I can rebase the branch and incorporate the comment.

> One further improvement would be re-implementing `find_objects` in terms of your `iter_objects` (it should be as simple as just calling `iter_objects` and then `collect` on the result 🤔 )
> 

That's been incorporated.

cryptoki/src/session/object_management.rs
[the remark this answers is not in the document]
keldonin: Will push a commit for this.

cryptoki/src/session/object_management.rs
```diff
@@ -3,66 +3,249 @@
...
-        unsafe {
-            Rv::from(get_pkcs11!(self.client(), C_FindObjectsFinal)(
-                self.handle(),
-            ))
-            .into_result(Function::FindObjectsFinal)?;
+impl Drop for ObjectHandleIterator<'_> {
+    fn drop(&mut self) {
+        // bark but pass if C_FindObjectsFinal() is not implemented
+        if let Some(f) = get_pkcs11_func!(self.session.client(), C_FindObjectsFinal) {
+            log::error!("C_FindObjectsFinal() is not implemented on this library");
```
ionut-arm, line 176: I'm not sure why this log is here, isn't this branch the one where `C_FindObjectsFinal` exists?

ionut-arm (commented):
Approach looks good to me, happy to go ahead with it, just a few minor comments

cryptoki/src/session/object_management.rs
```diff
@@ -3,66 +3,249 @@
...
+            // swallow the return value, as we can't do anything about it
+            let _ = unsafe { f(self.session.handle()) };
```
wiktor-k, line 178:
```suggestion
            if let Rv::Error(error) = Rv::from(unsafe { f(self.session.handle()) }) {
                log::error!(...);
            }
```

I meant something like this, sorry I wasn't specific :)

cryptoki/src/session/object_management.rs
```diff
@@ -3,66 +3,249 @@
...
+    ///
+    /// It is a convenience function that will call [`Session::iter_objects`] and collect the results.
+    ///
+    /// # See also
+    ///
+    /// * [`Session::iter_objects`] for a way to specify the cache size
+    #[inline(always)]
+    pub fn find_objects(&self, template: &[Attribute]) -> Result<Vec<ObjectHandle>> {
+        self.iter_objects(template)?
+            .collect::<Result<Vec<ObjectHandle>>>()
```
wiktor-k, line 248: I wonder if this would suffice:

```suggestion
            .collect()
```

(since it's in a return position and the type of the result and this is identical)

cryptoki/src/session/object_management.rs
```diff
@@ -3,66 +3,249 @@
...
+use crate::error::{Error, Result, Rv, RvError};
 use crate::object::{Attribute, AttributeInfo, AttributeType, ObjectHandle};
 use crate::session::Session;
 use cryptoki_sys::*;
 use std::collections::HashMap;
 use std::convert::TryInto;
+use std::num::NonZeroUsize;
 
 // Search 10 elements at a time
 const MAX_OBJECT_COUNT: usize = 10;
```
wiktor-k, line 15: AFAIK `NonZeroUsize::new` is const and could be directly initialized here:

```suggestion
const MAX_OBJECT_COUNT: NonZeroUsize = NonZeroUsize::new(10).expect("non-zero value to be passed");
```

wiktor-k (commented):
I think it's the last round of feedback and in general it looks very nice already :ok_hand:

cryptoki/src/session/object_management.rs
```diff
@@ -3,66 +3,249 @@
...
-        let mut template: Vec<CK_ATTRIBUTE> = template.iter().map(|attr| attr.into()).collect();
+/// Iterator over object handles, in an active session.
+///
+/// Used to iterate over the object handles returned by underlying calls to `C_FindObjects`.
+/// The iterator is created by calling the `iter_objects` and `iter_objects_with_cache_size` methods on a `Session` object.
+///
+/// # Note
+///
+/// The iterator `new()` method will call `C_FindObjectsInit`. It means that until the iterator is dropped,
+/// creating another iterator will result in an error (typically `RvError::OperationActive` ).
```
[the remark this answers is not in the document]
keldonin: Sure I can add it. However the `new()` method is not public, it will not appear in the generated doc.

cryptoki/src/session/object_management.rs
keldonin: good catch.

cryptoki/src/session/object_management.rs
keldonin: `expect()` is not stable yet to be assigned to a `const`. In this instance, I believe it is OK to use `new_unchecked()` and document the safety condition in the code.

cryptoki/src/session/object_management.rs
keldonin: resolved - see other related comment.

wiktor-k (approved):
LGTM :+1: thanks for taking the time to address all remarks :)

keldonin:
@wiktor-k: 

NP, feedback is useful and relevant!

I don't know what is your policy for merging changes, but since you made this request:

> I also wonder if it'd be a good idea to add Closes: https://github.com/parallaxsecond/rust-cryptoki/pull/109 in the commit message trailer (since it appears to be the same functionality but with a separate method).

I suggested I would rebase this PR, but that would make another round of approvals, and you would loose visibility on the changes before merging. Maybe is it simpler for you to squash merge and add this comment to the commit message?

cryptoki/src/context/mod.rs
```diff
@@ -13,6 +13,14 @@ macro_rules! get_pkcs11 {
     };
 }
 
+/// Same as get_pkcs11! but does not attempt to apply '?' syntactic sugar.
+/// Suitable only if the caller can't return a Result.
+macro_rules! get_pkcs11_func {
```
hug-dev, line 18: Would it work to modify `get_pkcs11` macro to use this one + the `ok_or`? Meaning to avoid code duplication :)

cryptoki/tests/basic.rs
```diff
@@ -349,19 +350,111 @@ fn session_find_objects() {
...
+    (1..=11).for_each(|i| {
+        let key_template = vec![
+            Attribute::Token(true),
+            Attribute::Encrypt(true),
+            Attribute::Label(format!("key_{}", i).as_bytes().to_vec()),
+            Attribute::Id("12345678".as_bytes().to_vec()), // reusing the same CKA_ID
+        ];
+
+        // generate a secret key
+        let _key = session.generate_key(&Mechanism::Des3KeyGen, &key_template);
```
hug-dev, line 390: I noticed that you removed the `unwrap` here, is it ok?

cryptoki/src/session/object_management.rs
```diff
@@ -3,24 +3,277 @@
...
+            } else {
+                // reset counters and proceed to the next section
+                self.index = 0;
+
+                if self.object_count < self.cache.len() {
+                    // if self.object_count is less than the cache size,
+                    // it means our last call to C_FindObjects returned less than the cache size
+                    // At this point, we have exhausted all objects in the cache
+                    // and we can safely break the loop and return None
+                    self.object_count = 0;
```
hug-dev, line 161: Since this is done in both branches, maybe this line can be done before and so we just need the `if` condition with the `break`?

cryptoki/src/session/object_management.rs
```diff
@@ -50,54 +303,10 @@ impl Session {
...
-            Rv::from(get_pkcs11!(self.client(), C_FindObjectsFinal)(
-                self.handle(),
-            ))
-            .into_result(Function::FindObjectsFinal)?;
-        }
-
-        let objects = objects.into_iter().map(ObjectHandle::new).collect();
-
-        Ok(objects)
+        self.iter_objects(template)?.collect()
```
hug-dev, line 309: Nice to re-use what you did here!

hug-dev (approved):
Nitpicking comments you can feel free to ignore if they are not worth it! Otherwise trusting Wiktor's review :pray:

cryptoki/src/session/object_management.rs
wiktor-k: Haha I had the same thought! But then the if condition depends on the variable that'd be overwritten (so the old value would've to be stored in a temporary variable)... :)

cryptoki/tests/basic.rs
wiktor-k: Yeah, let's use `?` here. The warning is suppressed because of the variable assignment but better to check IMO.

wiktor-k:
> Otherwise trusting Wiktor's review 🙏

Haha, please don't do that 😅 I try to be diligent here but it's still an unpaid side-project so my time is limited.

I still think that given our collective effort the code continously gets better and better (and with tests future refactoring is always possible).

👋

cryptoki/src/session/object_management.rs
keldonin: I couldn't give a better answer 😅

cryptoki/src/session/object_management.rs
keldonin: Rewriting `find_objects` as collection of the iterator was an idea suggested by @wiktor-k. And yes, the code here was part of a previous PR, right? I guess some of it belongs to the iterator code now.

cryptoki/src/context/mod.rs
keldonin: Seems reasonable, I will give it a stab. Not much XP with Rust macro authoring, however. I'll have a look at it this week-end.

cryptoki/tests/basic.rs
keldonin: @hug-dev good catch. I will need to refactor a bit, as the closure don't like handling `?`.

cryptoki/src/context/mod.rs
keldonin: here is it - easier than anticipated it seems.

wiktor-k (approved):

cryptoki/src/session/object_management.rs
hug-dev: ahh got it!! Programming is not as easy as it seems :eyeglasses:

hug-dev (approved):

keldonin:
Thank you for merging this one!

Out of curiosity, are you planning to make a new tagged release at some point in time?

hug-dev:
Nothing planned for now I think, it's mostly demand-based! We also try to group release with other dependencies in `parallaxsecond`. If you need one for your projects, we can start to see when will work!

hug-dev:
On another topic @keldonin, would you like to be a maintainer of this crate? Since you have shown interest in it in the past few weeks with very nice improvements I was wondering if you would like to help us maintaining it?
There is no special asks or responsibility apart from the power to be able to approve and merge PRs!

keldonin:
> Nothing planned for now I think, it's mostly demand-based! We also try to group release with other dependencies in `parallaxsecond`. If you need one for your projects, we can start to see when will work!

Hi @hug-dev. Nothing pressing for now, just asking. Thank you for the clarification!

keldonin:
> On another topic @keldonin, would you like to be a maintainer of this crate? Since you have shown interest in it in the past few weeks with very nice improvements I was wondering if you would like to help us maintaining it? There is no special asks or responsibility apart from the power to be able to approve and merge PRs!

I would be honoured.

Time-wise, something I can perform during my free time (probably like you? 🤔) - that I share with other activities - there is life out there 😂. 

So, glad to give support here, to the extent of my availability.

hug-dev:
That's great, thank you!
There is absolutely no ask from our side, put into it as much time as you want! It's mostly so that we can share the merging power so there is less chance that a PR stays too long in limbo state/unreviewed :smiley:

hug-dev:
It's done now! If you want, you can join the `#parsec` Slack channel on the CNCF Slack workspace :) More info [here](https://github.com/parallaxsecond/community?tab=readme-ov-file#community-channel)

PR #224, opened 2024-09-07
flip order of CI commands, to check fmt and clippy first

keldonin:
Attempt at making faster builds, by simply checking `fmt` and `clippy` upfront. I found it quite frustrating, several times, to go through all the builds to finally end up with a `clippy` error.

Hopefully preventing some carbon dioxide emission as well.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -7,6 +7,18 @@
 
 set -euf -o pipefail
 
+pushd cryptoki-sys
+RUST_BACKTRACE=1 cargo build --features generate-bindings
+popd
+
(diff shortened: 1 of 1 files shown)


wiktor-k (approved):
Agreed. In my own repos I also structure tests so that the faster ones execute first (https://github.com/wiktor-k/service-binding/blob/main/.justfile#L3).

Thanks. I think it's relatively small change 👍

ionut-arm (approved):
Makes sense, thanks!

PR #226, opened 2024-09-20
WIP: supports mutable IV in GcmParams, close #225

### LINKED ISSUE #225: Use GcmParams with AWS CloudHSM will cause undefined behavior
AWS [doc](https://docs.aws.amazon.com/cloudhsm/latest/userguide/pkcs11-mechanisms.html#pkcs11-mech-annotations) says 
> When performing AES-GCM encryption, the HSM does not accept initialization vector (IV) data from the application. You must use an IV that it generates. The 12-byte IV provided by the HSM **is written into the memory reference pointed to by the pIV element** of the `CK_GCM_PARAMS` parameters structure that you supply.

and in code
https://github.com/parallaxsecond/rust-cryptoki/blob/024976fc892b96ad72fbc4af38cf9449c42034c4/cryptoki/src/mechanism/aead.rs#L39
https://github.com/parallaxsecond/rust-cryptoki/blob/024976fc892b96ad72fbc4af38cf9449c42034c4/cryptoki/src/mechanism/aead.rs#L58

It violate the Rust aliasing model, cases undefined behavior, and it's impossible to extract IV content from AWS CloudHSM SDK.

## Possible solutions

1. simplely modify the function signature, from `&'a [u8]` to `&'a mut [u8]`, and it also affect the variance of `GcmParams<'a>` , from `PhantomData<&'a [u8]>` to `PhantomData<&'a mut [u8]>`
2. add a new type `GcmParamsMut` and mark the `GcmParams::new` as unsafe (or safe is ok?).

zkonge:
so that some PKCS11 implementation (like AWS CloudHSM) could write random IV into it.


wiktor-k (approved):
I'm not using this scheme personally but the underlying issue seems real and the trade-offs in this solution reasonable.

Wdyt @hug-dev @ionut-arm @keldonin et al?

zkonge:
Although the trait constraint for `Mechanism` is enough for the alias rule:
```
Mechanism: !Clone+!Send
```

I open another branch https://github.com/parallaxsecond/rust-cryptoki/commit/43c8863750065224f56b29a3ab2a16b51272bdbe that make all operations use the owned `Mechanism`.

It's not necessary, but I think it can give us (and the compiler) more restrictions on when we can get things wrong.

any ideas about that?

wiktor-k:
> Any ideas about that? And is there anyone prefer passing the mutable reference instead of owned data?

Wait, I thought mutable ref would allow the PKCS11 implementation to set these fields that are to be read by the caller, which would be impossible when using owned data? (unless the owned object is being returned from the function too).

Btw, I think we can resume work on this given recent changes in `main`.

Sorry for the delays @zkonge :bow:

fiercelybearded:
Hello, I'm curious if there's any update on this one? I know this PR is out of date with the current release, 0.9.0, but I was able to make the same changes to an updated branch and I have success with AES GCM on AWS CloudHSM. I've verified that the mutable reference allows the IV to be provided back to the caller. This is with the V5 cloudhsm pkcs11 client.

The changes don't seem to affect anything with SoftHSM tests, and my Luna SA works with it as well.

wiktor-k:
Hmm... good question. We regularly have "contributor missing-in-action" situations here but it looks like all checks passed and the PR is merge-able so we could take it as is.

@hug-dev what do you think about merging it as is?

hug-dev:
I think it's fine since we agreed in https://github.com/parallaxsecond/rust-cryptoki/issues/225 that the approach of passing `&mut Mechanism` should be done everywhere.

Although I would have liked a bit more comments on the reason of the mutability I think this is fine! We should complete this at some point moving all operations to `&mut Mechanism` as described in the issue.

Better to merge this as is if more people can benefit from it :)

hug-dev (approved):

wiktor-k:
Yep, totally agreed on all points. Thanks for your time, Hugues :bow:

PR #227, opened 2024-10-11
feat: add HKDF mechanisms

Direktor799:
Add HKDF mechanisms to `cryptoki` and `cryptoki-sys`. The mechanisms were defined in [PKCS#11 v3.0](https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061597).

I found that SoftHSM doesn't support such mechanisms so I couldn't write any unit test. But I did run some `CKM_HKDF_DERIVE` tests against our own HSM and it works.

### DIFF
--- a/cryptoki/src/mechanism/hkdf.rs
+++ b/cryptoki/src/mechanism/hkdf.rs
@@ -1,3 +1,5 @@
+// Copyright 2021 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
 //! Mechanisms of hash-based key derive function (HKDF)
 //! See: <https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061597>
 
@@ -9,8 +11,8 @@ use crate::object::ObjectHandle;
 
 use super::MechanismType;
 
-#[derive(Debug, Clone, Copy)]
 /// The salt for the extract stage.
+#[derive(Debug, Clone, Copy)]
 pub enum HkdfSalt<'a> {
     /// CKF_HKDF_SALT_NULL no salt is supplied.
     Null,
@@ -62,16 +64,17 @@ impl<'a> HkdfParams<'a> {
                     HkdfSalt::Data(_) => CKF_HKDF_SALT_DATA,
                     HkdfSalt::Key(_) => CKF_HKDF_SALT_KEY,
                 },
-                pSalt: match salt {
-                    HkdfSalt::Data(data) => data.as_ptr() as *mut _,
-                    _ => null_mut(),
+                pSalt: if let HkdfSalt::Data(data) = salt {
+                    data.as_ptr() as *mut _
+                } else {
+                    null_mut()
                 },
-                ulSaltLen: match salt {
-                    HkdfSalt::Data(data) => data
-                        .len()
+                ulSaltLen: if let HkdfSalt::Data(data) = salt {
+                    data.len()
                         .try_into()
-                        .expect("salt length does not fit in CK_ULONG"),
-                    _ => 0,
+                        .expect("salt length does not fit in CK_ULONG")
+                } else {
+                    0
                 },
                 hSaltKey: match salt {
                     HkdfSalt::Key(key) => key.handle(),
--- a/cryptoki/src/mechanism/hkdf.rs
+++ b/cryptoki/src/mechanism/hkdf.rs
@@ -1,0 +1,116 @@
+//! Mechanisms of hash-based key derive function (HKDF)
+//! See: <https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061597>
+
+use std::{convert::TryInto, marker::PhantomData, ptr::null_mut, slice};
+
+use cryptoki_sys::{CKF_HKDF_SALT_DATA, CKF_HKDF_SALT_KEY, CKF_HKDF_SALT_NULL};
+
+use crate::object::ObjectHandle;
+
+use super::MechanismType;
+
+#[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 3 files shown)


hug-dev:
Thank you!! As a prerequisite to this PR maybe we should agree and go on with https://github.com/parallaxsecond/rust-cryptoki/pull/156 so that we use `v3.0` and then merge your changes based on that?

wiktor-k:
> Thank you!! As a prerequisite to this PR maybe we should agree and go on with #156 so that we use `v3.0` and then merge your changes based on that?

That's a good idea. This PR unnecessarily couples the HKDF and the 3.0 change that maybe should be there in the first place. Does #156 also solve the "unofficial OASIS header" issue reported at https://github.com/parallaxsecond/rust-cryptoki/pull/66? If so maybe we need some acks from other maintainers that the licensing looks good in #156 and then rebase this on top of master (when #156 is done).

@hug-dev would you ask in `#parsec` if everyone is OK with this approach? :pray:

Direktor799:
> Thank you!! As a prerequisite to this PR maybe we should agree and go on with #156 so that we use `v3.0` and then merge your changes based on that?

That will be nice, thank you!

hug-dev:
> @hug-dev would you ask in #parsec if everyone is OK with this approach? 🙏

Yes! Also pinged on the related PR to use another free source of headers.

cryptoki/src/mechanism/hkdf.rs
```diff
@@ -0,0 +1,116 @@
...
+use std::{convert::TryInto, marker::PhantomData, ptr::null_mut, slice};
+
+use cryptoki_sys::{CKF_HKDF_SALT_DATA, CKF_HKDF_SALT_KEY, CKF_HKDF_SALT_NULL};
+
+use crate::object::ObjectHandle;
+
+use super::MechanismType;
+
+#[derive(Debug, Clone, Copy)]
+/// The salt for the extract stage.
```
wiktor-k, line 13: Could you move the doc comment above the `#[derive]` for consistency?

cryptoki/src/mechanism/hkdf.rs
```diff
@@ -0,0 +1,116 @@
...
+    ///
+    /// * `salt` - The salt for the extract stage.
+    ///
+    /// * `info` - The info string for the expand stage.
+    pub fn new(
+        extract: bool,
+        expand: bool,
+        prf_hash_mechanism: MechanismType,
+        salt: HkdfSalt,
+        info: &'a [u8],
```
wiktor-k, line 53: Hmm... `info` is required even when `expand` is set to `false`? Is `info` always set when `expand` is `true` and empty otherwise?

If so the two could be collapsed into an `expand_info: Option<&[u8]>` param :thinking: 

(in general bool parameters are annoying since it's not visible from the call-site which is which)

cryptoki/src/mechanism/hkdf.rs
```diff
@@ -0,0 +1,116 @@
...
+                    _ => null_mut(),
+                },
+                ulSaltLen: match salt {
+                    HkdfSalt::Data(data) => data
+                        .len()
+                        .try_into()
+                        .expect("salt length does not fit in CK_ULONG"),
+                    _ => 0,
+                },
+                hSaltKey: match salt {
```
wiktor-k, line 76: I'd convert all these two-arm matches to `if let`s... but this is purely a stylistic choice :sweat_smile:

cryptoki/src/mechanism/hkdf.rs
```diff
@@ -0,0 +1,116 @@
+//! Mechanisms of hash-based key derive function (HKDF)
```
wiktor-k, line 1: It'd be good to insert a copyright header here, just like so: https://github.com/parallaxsecond/rust-cryptoki/blob/main/cryptoki/src/types.rs#L1-L2

wiktor-k (dismissed):
In general this looks good, just a couple of nits.

Btw it seems softhsm [does not implement HKDF](https://github.com/opendnssec/SoftHSMv2/issues/189#issuecomment-187915000) which is sad :(

wiktor-k (approved):
LGTM :+1: thanks!

cryptoki/src/mechanism/hkdf.rs
hug-dev: @wiktor-k do we use to put 2021 everywhere or the year the file is added instead?

hug-dev (approved):
Thanks! ONe question about copyright year that can be solved later

hug-dev:
@Direktor799 You might need to rebase!

wiktor-k:
> do we use to put 2021 everywhere or the year the file is added instead?

I think the date of the file addition was used in previous instances. The entire copyright years issues is a minefield: https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/

In [my other project](https://gitlab.archlinux.org/archlinux/signstar/-/blob/main/REUSE.toml?ref_type=heads) we've migrated to [reuse](https://reuse.software/) which automatically checks licensing info. Previously we used headers, just like here, but then we aggregated everything in one file and it's good. (I guess this is a question if we want to improve copyright headers in the long run and if so how).

cryptoki/src/mechanism/hkdf.rs
hug-dev: Let's use 2024 here instead there, until we find a  better solution :)

cryptoki/src/mechanism/hkdf.rs
```diff
@@ -0,0 +1,121 @@
+// Copyright 2021 Contributors to the Parsec project.
```
wiktor-k, line 1:
```suggestion
// Copyright 2024 Contributors to the Parsec project.
```

wiktor-k (approved):
LGTM :+1: thanks!

hug-dev (approved):

PR #228, opened 2024-10-18
Feat/pkcs11 3.0

Direktor799:
Replace `pkcs11.h` with https://github.com/latchset/pkcs11-headers/blob/main/public-domain/3.0/pkcs11.h, which was mentioned in #156.

- Use latchset PKCS#11 3.0 header
- Manually specify `CK_UNAVAILABLE_INFORMATION` due to a bug(i believe) in `bindgen`
- Add 3.0 macro definations
- Enable `PKCS11_DEPRECATED` for backward compatibility with 2.4
- Deprecate `EC_NAMEDCURVE`, replace it with `EC_OID`


cryptoki-sys/src/lib.rs
```diff
@@ -138,22 +138,6 @@ include!(concat!(
...
-pub const CKA_DERIVE_TEMPLATE: CK_ATTRIBUTE_TYPE = CKF_ARRAY_ATTRIBUTE | 0x00000213;
-pub const CKA_ALLOWED_MECHANISMS: CK_ATTRIBUTE_TYPE = CKF_ARRAY_ATTRIBUTE | 0x00000600;
-pub const CK_UNAVAILABLE_INFORMATION: CK_ULONG = !0;
-pub const CKF_EXTENSION: CK_FLAGS = 0x80000000;
-pub const CKK_EC_MONTGOMERY: CK_KEY_TYPE = 0x00000041;
-pub const CKR_VENDOR_DEFINED: CK_RV = 0x80000000;
-pub const CKR_CURVE_NOT_SUPPORTED: CK_RV = 0x00000140;
-pub const CKM_VENDOR_DEFINED: CK_MECHANISM_TYPE = 0x80000000;
+// bindgen generates u64::MAX value for ~0UL macro defination, it's not valid on 32bit ulong platforms.
+// This is a workaround for that.
```
hug-dev: Isn't it something we could configure when generating the bindings with bindgen? Or if that's a bindgen bug, is there a public issue we could link maybe :)?

cryptoki-sys/pkcs11.h
hug-dev, line 1: Could you maybe apply the same structure than in #156 and put this header inside a `vendor` folder where we also would have a `README.md` with a little bit more context on where the headers are coming from? Would be easier to update then!

cryptoki/src/mechanism/mechanism_info.rs
```diff
@@ -193,10 +194,21 @@ impl MechanismInfo {
     /// **[Conformance](crate#conformance-notes):**
     /// *At least* one of [`ec_from_parameters`](Self::ec_from_parameters) and
     /// [`ec_from_named_curve`](Self::ec_from_named_curve) must be `true`
+    #[deprecated = "use `ec_from_oid` instead"]
```
hug-dev, line 197: Nice touch!

hug-dev (commented):
Thanks a lot! Excellent as well if this is compatible with 2.4. A few general comments but I am okay otherwise with it!

cryptoki-sys/pkcs11.h
Direktor799: > Could you maybe apply the same structure than in #156 and put this header inside a `vendor` folder where we also would have a `README.md` with a little bit more context on where the headers are coming from? Would be easier to update then!

updated

hug-dev (approved):
Feel free to resolve the last comment, if it's not something you were planning to investigate. I think it's fine to go as it is, and I can open an issue for this specific item to be looked at later :)

Direktor799:
@hug-dev also what do you suggest to do with the misspelling in the new header..

hug-dev:
Thanks for research! I think it's fine as it is, this PR will be the history in case someone wants to look this up.

> @hug-dev also what do you suggest to do with the misspelling in the new header..

I think we can add the folder in the ignored [here](https://github.com/parallaxsecond/rust-cryptoki/blob/024976fc892b96ad72fbc4af38cf9449c42034c4/.codespellrc#L2) to make it work!

PR #230, opened 2024-10-25
Add capability to load symbols from current executable

EliseChouleur:
If PKCS#11 symbols are available inside the current executable as it can be necessary for iOS development.

### DIFF
--- a/cryptoki/src/context/mod.rs
+++ b/cryptoki/src/context/mod.rs
@@ -85,15 +85,32 @@ pub struct Pkcs11 {
     initialized: Arc<RwLock<bool>>,
 }
 
+#[derive(Debug)]
+/// Type of library to load in the instantiation of a new Pkcs11 context.
+pub enum LibLoadingType<P: AsRef<Path>> {
+    /// Load current executable, the PKCS11 implementation is contained in the current executable
+    OpenSelf,
+    /// Open dynamic library specify in input
+    Open(P),
+}
+
 impl Pkcs11 {
     /// Instantiate a new context from the path of a PKCS11 dynamic library implementation.
-    pub fn new<P>(filename: P) -> Result<Self>
+    pub fn new<P>(filename: LibLoadingType<P>) -> Result<Self>
     where
         P: AsRef<Path>,
     {
         unsafe {
-            let pkcs11_lib =
-                cryptoki_sys::Pkcs11::new(filename.as_ref()).map_err(Error::LibraryLoading)?;
+            let pkcs11_lib = match filename {
+                LibLoadingType::OpenSelf => {
+                    #[cfg(not(windows))]
+                    let this_lib = libloading::os::unix::Library::this();
+                    #[cfg(windows)]
+                    let this_lib = libloading::os::windows::Library::this();
+                    cryptoki_sys::Pkcs11::from_library(this_lib)?
+                }
+                LibLoadingType::Open(filename) => cryptoki_sys::Pkcs11::new(filename.as_ref()).map_err(Error::LibraryLoading)?
+            };
             let mut list = mem::MaybeUninit::uninit();
 
             Rv::from(pkcs11_lib.C_GetFunctionList(list.as_mut_ptr()))
--- a/cryptoki/src/context/mod.rs
+++ b/cryptoki/src/context/mod.rs
@@ -85,47 +85,46 @@ pub struct Pkcs11 {
     initialized: Arc<RwLock<bool>>,
 }
 
-#[derive(Debug)]
-/// Type of library to load in the instantiation of a new Pkcs11 context.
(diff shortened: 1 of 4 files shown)


hug-dev (commented):
Thanks for the PR!

I am curious, in which context would the library be contained in the executable? In case it has been statically linked :thinking:?

Also wondering if instead of changing the interface and introducing the new `LibLoadingType` we could create a new constructor, like `new_from_self` which would do the same :)

wiktor-k:
I agree with @hug-dev that this can be simplified. Btw, I'm on mobile, isn't the libloading dep missing from cargo.toml?

The idea behind the change is good though!

EliseChouleur:
> Thanks for the PR!
> 
> I am curious, in which context would the library be contained in the executable? In case it has been statically linked 🤔?
> 
> Also wondering if instead of changing the interface and introducing the new `LibLoadingType` we could create a new constructor, like `new_from_self` which would do the same :) What do you think?

Yes when it's statically linked, which is mandatory on iOS when following the AppStore rules :)
I was hesitating between different approch when searching on how to add this feature and the duplication of the new function was one of them, so I've changed it to that.

EliseChouleur:
> I agree with @hug-dev that this can be simplified. Btw, I'm on mobile, isn't the libloading dep missing from cargo.toml?
> 
> The idea behind the change is good though!

We were already using this on iOS and Android with the old rust-pkcs11 crate (but never merged on upstream), so it's working well. And I've tested it on a real device ;)

cryptoki/src/context/mod.rs
```diff
@@ -94,21 +94,37 @@ impl Pkcs11 {
...
+        }
+    }
+
+    /// Instantiate a new context from current executable, the PKCS11 implementation is contained in the current executable
+    pub fn new_from_self() -> Result<Self> {
+        unsafe {
+            #[cfg(not(windows))]
+            let this_lib = libloading::os::unix::Library::this();
+            #[cfg(windows)]
+            let this_lib = libloading::os::windows::Library::this();
```
wiktor-k, line 107: Apparently the Windows variant returns a Result: https://docs.rs/libloading/latest/libloading/os/windows/struct.Library.html#method.this

```suggestion
            let this_lib = libloading::os::windows::Library::this()?;
```

wiktor-k:
This is shaping really nicely 👌 

I won't be able to look at it deeply before Monday, though :(

wiktor-k (approved):
LGTM :+1: thanks!

cryptoki/src/context/mod.rs
```diff
@@ -94,21 +94,37 @@ impl Pkcs11 {
...
+            #[cfg(windows)]
+            let this_lib = libloading::os::windows::Library::this()?;
+            let pkcs11_lib = cryptoki_sys::Pkcs11::from_library(this_lib)?;
+            Self::_new(pkcs11_lib)
+        }
+    }
 
-            Rv::from(pkcs11_lib.C_GetFunctionList(list.as_mut_ptr()))
-                .into_result(Function::GetFunctionList)?;
+    unsafe fn _new(pkcs11_lib: cryptoki_sys::Pkcs11) -> Result<Self> {
```
hug-dev, line 113: Is it idiomatic in Rust to have function names starting with `_` like this? Wondering if it could be a normal name instead, since the function is not public. But no big deal! Can be done later :)

hug-dev (approved):

cryptoki/src/context/mod.rs
EliseChouleur: No preference on my side, it's just something we commonly do at my work :)

EliseChouleur:
Could you tell me what's happening next ? Or more precisely, what could be the ETA of an official release with this version ? :)
Thanks a lot !

cryptoki/src/context/mod.rs
hug-dev: Ok let's go with it! If someone has strong opinions they can change in the future :)

hug-dev:
> what could be the ETA of an official release with this version 

We have no schedule set! It's all based on demand. Since the last release was in June and we got quite a lot of contributions recently, I guess we could make a new version if you wanted!

wiktor-k:
> I guess we could make a new version if you wanted!

I think it's a good idea. As you've said we have quite a lot of good contributions piled-up in `main` that people may wait for :+1:

hug-dev:
We can maybe announce in the [community channel](https://github.com/parallaxsecond/community?tab=readme-ov-file#community-channel) that a new release will be done in like two weeks, so that people have time to push in more changes before the release?

wiktor-k:
> We can maybe announce in the [community channel](https://github.com/parallaxsecond/community?tab=readme-ov-file#community-channel) that a new release will be done in like two weeks, so that people have time to push in more changes before the release?

Yes, excellent idea! :)

PR #231, opened 2024-10-30
Move crate documentation to README and add code example

wiktor-k:
The old root module documentation referred to the unused "new" module. This PR cleans up the docs a bit and adds code example which gets executed as part of the unit tests.


cryptoki/README.md
```diff
@@ -7,4 +7,77 @@
 
 This is the high-level, Rust idiomatic wrapper crate for PKCS #11.
 
+The items in this crate only expose idiomatic and safe Rust types and
+functions to interface with the PKCS11 API. All the PKCS11 items might
+not be implemented but everything that is implemented is safe.
+
+## Example
+
+The following example initializes an empty token and generates a new RSA key.
```
ionut-arm, line 16: Any chance you could also add some comments throughout the code to broadly describe each section?

ionut-arm (dismissed):
💯

cryptoki/README.md
```diff
@@ -7,4 +7,77 @@
 
 This is the high-level, Rust idiomatic wrapper crate for PKCS #11.
 
+The items in this crate only expose idiomatic and safe Rust types and
+functions to interface with the PKCS11 API. All the PKCS11 items might
+not be implemented but everything that is implemented is safe.
+
+## Example
```
hug-dev, line 14: The example is very nice! Wouldn't it be nice to copy it to a new `cryptoki/examples/rsa.rs` file (like we do in `tss-esapi`) so that the example is checked for compilation every time `cargo test` is run? That way we are sure it at least compiles as the crate evolves.

Here we could link to the respective example instead :)

hug-dev (commented):
Thanks a lot! Added a suggestion to have an `examples` folder

cryptoki/README.md
wiktor-k: > so that the example is checked for compilation every time cargo test is run? That way we are sure it at least compiles as the crate evolves.

I'm not sure if it's completely clear but this example in the readme is *also* checked every time `cargo test` runs (due to the `doc include str` thing in `lib.rs`).

(Adding more examples, or even the same for redundancy, is :ok_hand: with me)

hug-dev: ahhhh I did not know that, this is very neat!

hug-dev (approved):

cryptoki/README.md
wiktor-k: Haha, yes, I like it because it's both illustrative *and* checked. I try to do that in every crate I contribute to. Just today [I fixed a README.md example](https://github.com/matusf/z-base-32/pull/60) that was slightly broken because it didn't use this style :)

cryptoki/README.md
wiktor-k: Please see if this looks better. Frankly, I'm kind of bad with coming up with comments so if something's not good I'd be really glad if you added a couple of suggestions :bow:

hug-dev (approved):

cryptoki/README.md
ionut-arm: Awesome stuff, thank you!!

ionut-arm (approved):
💯

Issue #105, opened 2022-10-25
Vendored mechanisms


Zorvalt:
Hello,

I am working with a PKCS11 interface requiring the use of vendored mechanisms. I did not find a way to do so with this crate and was wondering if this could be added. Beside this issue (#54)  which focuses on errors, it seems to me the subject was not discussed, at least publicly.

Do you plan implementing this feature or would you accept a PR for it?

In case you are open to the idea, I already looked at the code and it seems to me this boils down to provide a mean to extend the `Mechanism` enum. Because the rust-cryptoki library cannot know in advance, I could not think of a simple way to ensure the library user constructs correct PKCS11 mechanisms.

I think a simple solution would be to offer an additional `VendoredMechanism` in the `Mechanism` enum which would expose the inners of what makes a `CK_MECHANISM` : a `MechanismType` and some arbitrary bytes as parameters.

In a working PoC, I implemented this:
```rust
// In src/mechanism/vendored.rs
pub struct VendoredMechanism {
    pub mech_type: MechanismType,
    pub params: Vec<CK_BYTE>,
}

// In src/mechanism/mod.rs
pub enum Mechanism {
    ...
    VendoredMechanism(vendored::VendoredMechanism),
}

impl Mechanism {
    /// Get the type of a mechanism
    pub fn mechanism_type(&self) -> MechanismType {
        ...
        Mechanism::VendoredMechanism(mech) => mech.mech_type,
    }
}

impl From<&Mechanism> for CK_MECHANISM {
    fn from(mech: &Mechanism) -> Self {
        let mechanism = mech.mechanism_type().into();
        match mech {
            Mechanism::VendoredMechanism(mech) => CK_MECHANISM {
                mechanism,
                pParameter: mech.params.as_slice() as *const _ as *mut c_void,
                ulParameterLen: mech.params.len() as u64,
            },
        }
    }
}
```

Which can be used as so:
```rust
pub const SOME_VENDORED_MECH: MechanismType = MechanismType {
    val: SOME_VENDORED_MECH_VALUE,
};

pub struct SomeVendoredMech {
    pub field1: u8,
    pub field2: u8,
}

impl From<&SomeVendoredMech> for VendoredMechanism {
    fn from(mech: &SomeVendoredMech) -> Self {
        VendoredMechanism {
            mech_type: SOME_VENDORED_MECH,
            params: vec![
                mech.field1,
                mech.field2,
            ],
        }
    }
}
```

[...]

PR #232, opened 2024-11-07
feat: support vendor defined mechanisms

Direktor799:
Add support for vendor defined mechanisms.

In order to avoid issues mentioned in #105, the parameters are pre-serialized in `VendorDefinedMechanism::new`, so we don't need to introduce generic to `Mechanism` or pay extra cost for dynamic dispatching.
It's also guaranteed that the new mechanism value is greater or equal to `CKM_VENDOR_DEFINED`.

It seems to be working fine with AWS CloudHSM's custom `CKM_SP800_108_COUNTER_KDF`.

### DIFF
--- a/cryptoki/src/mechanism/mod.rs
+++ b/cryptoki/src/mechanism/mod.rs
@@ -8,6 +8,7 @@ pub mod elliptic_curve;
 pub mod hkdf;
 mod mechanism_info;
 pub mod rsa;
+pub mod vendor_defined;
 
 use crate::error::Error;
 use cryptoki_sys::*;
@@ -18,6 +19,7 @@ use std::fmt::Formatter;
 use std::mem::size_of;
 use std::ops::Deref;
 use std::ptr::null_mut;
+use vendor_defined::VendorDefinedMechanism;
 
 use crate::mechanism::rsa::PkcsOaepParams;
 pub use mechanism_info::MechanismInfo;
@@ -297,6 +299,28 @@ impl MechanismType {
     };
     /// HKDF-DATA mechanism
     pub const HKDF_DATA: MechanismType = MechanismType { val: CKM_HKDF_DATA };
+
+    /// Create vendor defined mechanism
+    ///
+    /// # Arguments
+    ///
+    /// * `adding` - The adding based on `CKM_VENDOR_DEFINED`
+    ///
+    /// Usually vendors defines custom mechanism like this:
+    /// ```c
+    /// #define CKM_SOME_CUSTOM_MECH (CKM_VENDOR_DEFINED | 0x00000001UL)
+    /// ```
+    ///
+    /// It maps to
+    /// ```rust
+    /// pub const CKM_SOME_CUSTOM_MECH: MechanismType =
+    ///     MechanismType::new_vendor_defined(0x00000001);
+    /// ```
+    pub const fn new_vendor_defined(adding: u64) -> MechanismType {
+        MechanismType {
+            val: CKM_VENDOR_DEFINED | adding,
+        }
+    }
 
     pub(crate) fn stringify(mech: CK_MECHANISM_TYPE) -> String {
         match mech {
@@ -937,6 +961,9 @@ pub enum Mechanism<'a> {
     HkdfDerive(hkdf::HkdfParams<'a>),
     /// HKDF-DATA mechanism
     HkdfData(hkdf::HkdfParams<'a>),
+
+    /// Vendor defined mechanism
+    VendorDefined(VendorDefinedMechanism<'a>),
 }
 
 impl Mechanism<'_> {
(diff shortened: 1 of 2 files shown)


cryptoki/src/mechanism/mod.rs
```diff
@@ -298,6 +300,30 @@ impl MechanismType {
...
+    /// ```
+    ///
+    /// It maps to
+    /// ```rust
+    /// use cryptoki::mechanism::MechanismType;
+    ///
+    /// pub const CKM_SOME_CUSTOM_MECH: MechanismType =
+    ///     MechanismType::new_vendor_defined(0x00000001);
+    /// ```
+    pub const fn new_vendor_defined(adding: CK_ULONG) -> MechanismType {
```
hug-dev, line 321: Should we shortcut it and directly allow passing the mechanism type value (like directly `CKM_SOME_CUSTOM_MECH` in your example). The function could though have a small safety check and check that the mechanism type is higher or equal than `CKM_VENDOR_DEFINED`.

Also, what do you think of that function being `unsafe`? Since we are short circuiting the abstraction layers built in this crate?

cryptoki/src/mechanism/vendor_defined.rs
```diff
@@ -0,0 +1,36 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct VendorDefinedMechanism<'a> {
+    pub(crate) inner: CK_MECHANISM,
+    /// Marker type to ensure we don't outlive the data
+    _marker: PhantomData<&'a [u8]>,
+}
+
+impl<'a> VendorDefinedMechanism<'a> {
+    /// Create a new vendor defined mechanism.
+    pub fn new<T>(mechanism_type: MechanismType, params: Option<&'a T>) -> Self {
```
hug-dev, line 23: Similarly here, should we make that `unsafe` and leave the responsability to the caller to make sure it's safe?

hug-dev (commented):
Good idea to do it this way! Only questions about unsafety

cryptoki/src/mechanism/mod.rs
Direktor799: Hi, thanks for reviewing.

The reason for using `adding` here is to avoid returning a `Result` or `Option` that can not be unwraped in constants. If we want to change this while keeping it as a const function, we will have to use `panic!`.

Regarding the unsafety, my understanding is there are 2 sides.
1. User side: We are (kind of?) sure the user's input is valid by lifetime.
2. Vendor side: I think the current implementation already assumes vendors are trustworthy (by not marking functions as unsafe). So maybe we should continue trusting their library will return reasonable results without breaking safety rules?

hug-dev: > keeping it as a const function

Is it something which is needed to make this implementation work? Maybe a normal function is also ok?

I agree with your two points but was thinking what could happen if a completely wrong `adding` was given to the `new_vendor_defined` function? Then we would have a instance of `MechanismType` which might lead to errors without having need to have an `unsafe` call. But maybe as long as it's errors and not panics/seg faults it's fine?

Direktor799: Const function is not necessary but definitely more user friendly. But I'm ok with directly passing the value, it also makes sense to me.

Yes, I think it's fine, it should return `CKR_MECHANISM_INVALID` and `CKR_MECHANISM_PARAM_INVALID`.

hug-dev: > Yes, I think it's fine, it should return CKR_MECHANISM_INVALID and CKR_MECHANISM_PARAM_INVALID.

Nice, let's go with that then!

Direktor799: Sure. One last thing is do we consider re-export `CKM_VENDOR_DEFINED` from `cryptoki-sys`?

cryptoki/src/mechanism/vendor_defined.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright 2024 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+//! Mechanism types are defined with the objects and mechanism descriptions that use them.
+//! Vendor defined values for this type may also be specified.
+//! See: <https://docs.oasis-open.org/pkcs11/pkcs11-base/v3.0/os/pkcs11-base-v3.0-os.html#_Toc29976545>
```
ionut-arm, line 5: Are vendor mechanisms only supported in v3 of PKCS #11?

Direktor799: It's also supported in v2.4. I put v3.0 link here since we are using v3.0 headers now.

ionut-arm (approved):
👍🏻 thanks! I'm guessing SoftHSM doesn't have support for any (mock) vendor mechanisms, right?

wiktor-k:
> 👍🏻 thanks! I'm guessing SoftHSM doesn't have support for any (mock) vendor mechanisms, right?

My casual skim reveals they don't have any: https://github.com/opendnssec/SoftHSMv2/blob/913e7bfd463194fadcdd28f578087cc9c15643ee/src/lib/SoftHSM.cpp#L723

Too bad, it'd be cool to have tests for this :disappointed:

PR #237, opened 2024-12-21
Add support for vendor defined attributes

jrozner:


cryptoki/src/object.rs
```diff
@@ -254,6 +255,7 @@ impl AttributeType {
             CKA_UNWRAP_TEMPLATE => String::from(stringify!(CKA_UNWRAP_TEMPLATE)),
             CKA_DERIVE_TEMPLATE => String::from(stringify!(CKA_DERIVE_TEMPLATE)),
             CKA_ALLOWED_MECHANISMS => String::from(stringify!(CKA_ALLOWED_MECHANISMS)),
+            CKA_VENDOR_DEFINED => String::from(stringify!(CKA_VENDOR_DEFINED)),
```
hug-dev, line 258: Since vendor defined attributes can go from `CKA_VENDOR_DEFINED` to the last value of `CK_ULONG`, maybe we can specify the range here and put in the string what the value is?

cryptoki/src/object.rs
```diff
@@ -396,6 +399,7 @@ impl TryFrom<CK_ATTRIBUTE_TYPE> for AttributeType {
             CKA_VERIFY_RECOVER => Ok(AttributeType::VerifyRecover),
             CKA_WRAP => Ok(AttributeType::Wrap),
             CKA_WRAP_WITH_TRUSTED => Ok(AttributeType::WrapWithTrusted),
+            0x8000_0000..=0xffff_ffff => Ok(AttributeType::VendorDefined(attribute_type)),
```
hug-dev, line 402: Can you use `CKA_VENDOR_DEFINED` instead of the first value here and could we find a Rust idiomatic way of getting the best value of the `CK_ULONG` type? Could be that on a different architecture it's bigger than `0xffff_ffff` :nerd_face:

cryptoki/src/object.rs
```diff
@@ -518,6 +521,8 @@ pub enum Attribute {
     Value(Vec<u8>),
     /// Length in bytes of the value
     ValueLen(Ulong),
+    /// Vendor defined value
+    VendorDefined((CK_ATTRIBUTE_TYPE, Vec<u8>)),
```
hug-dev, line 525: Why do we need the `CK_ATTRIBUTE_TYPE` value here? If we need it, could we use an `AttributeType` instead, to minimize the `cryptoki-sys` values needed?

hug-dev (commented):
Thanks for adding this! This will complete nicely vendor-defined mechanisms already added here: https://github.com/parallaxsecond/rust-cryptoki/pull/232

hug-dev:
Does SoftHSM support them? Would be nice to add tests if possible.

cryptoki/src/object.rs
jrozner: This could be me not fully understanding how the serialization ends up working in the implementation but the vendor defined type needs to be known along with the data so that the correct value can be specified. I can definitely swap this out with an AttributeType and pull the value out of it.

jrozner:
I'll look into whether SoftHSM supports any vendor defined attributes and can write some tests if it does. I've been testing this with a Yubikey which does rely on them for specifying the touch/pin policy. This set of changes seems to have broken something around the attribute type value, it's working with some values but not others. Need to debug.

jrozner:
Looks like I just missed a compilation for my test code. Everything is working

cryptoki/src/object.rs
hug-dev: Got it, thanks!

cryptoki/src/object.rs
```diff
@@ -14,8 +14,9 @@ use std::fmt::Formatter;
 use std::mem::size_of;
 use std::ops::Deref;
 
+const MAX_CU_ULONG: CK_ULONG = !0;
```
hug-dev, line 17: Smart!

hug-dev (dismissed):
Too bad there are no tests in SoftHsm but good that you tested with Yubikey! Thanks for the changes!

hug-dev:
All good on my side! Seems that the CI is failing (can you see the logs btw? Could be a permission thing). Seems you need to `cargo fmt` all :)

jrozner:
Formatting fixed

hug-dev (dismissed):

cryptoki/src/object.rs
```diff
@@ -405,7 +413,6 @@ impl TryFrom<CK_ATTRIBUTE_TYPE> for AttributeType {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-#[non_exhaustive]
```
wiktor-k, line 408: Why this got removed? Correct me if I'm wrong but vendor defined doesn't cover all space of possible values? 🤔

wiktor-k (commented):
Looks good, just one question though.

cryptoki/src/object.rs
jrozner: Sorry, fixed that. You're correct that it's not exhaustive.

hug-dev:
You forgot to sign-off the last commit :green_apple:

hug-dev (approved):

jrozner:
signed off the commit and force pushed the branch

wiktor-k (approved):
LGTM 👍

PR #238, opened 2024-12-22
chore: remove dependency on psa-crypto

### LINKED ISSUE #100: Remove psa_crypto dependency
The conversions `impl TryFrom<psa_crypto::types::algorithm::Algorithm> for Mechanism` and `pub fn from_psa_crypto_hash` should be removed from cryptoki crate as ideally `cryptoki` should really not depend on `psa-crypto`. Instead we should have these functions implemented in the parsec service.

hug-dev:
Fix #100

When using the new version of `rust-cryptoki` the code removed here can be added in specialized conversion functions directly in Parsec.


wiktor-k (approved):
Cleanups are always welcome! 🥲

gowthamsk-arm (approved):
LGTM. Thanks.

PR #239, opened 2024-12-22
Define CKD_SHA256_KDF transformation

hug-dev:
Originally done by @freedge at https://github.com/parallaxsecond/rust-cryptoki/pull/213

Define CKD_SHA256_KDF transformation to be used with CKM_ECDH1_DERIVE.

Some HSM with FIPS restriction will refuse to derive keys with CKD_NULL. CKD_SHA256_KDF will do fine though.

Unfortunately this is not implemented on softHSM
(https://github.com/opendnssec/SoftHSMv2/pull/599) so I provide no test. This was tested fine against Thales DPOD.

Fix #92

PR #241, opened 2024-12-24
attribute extension defining Unhanled attribute type and EC Kdf improvements

ilkerBaltaci:
For undefined attribute types in PKCS#11 spec, I defined Unhandled attribute type to get attribute type value and data value as Vec<u8> type. I also change visibility of the struct variables to edit outside of the module. ECKdf struct edited for defining KDF type and shared value.

### DIFF
--- a/cryptoki/src/object.rs
+++ b/cryptoki/src/object.rs
@@ -6,7 +6,7 @@ use crate::error::{Error, Result};
 use crate::mechanism::MechanismType;
 use crate::types::{Date, Ulong};
 use cryptoki_sys::*;
-use log::error;
+use log::{error, warn};
 use std::convert::TryFrom;
 use std::convert::TryInto;
 use std::ffi::c_void;
@@ -136,6 +136,8 @@ pub enum AttributeType {
     Wrap,
     /// Indicates that the key can only be wrapped with a wrapping key that has the Trusted attribute
     WrapWithTrusted,
+    /// Wraps undefined and custom attribute types.
+    Unhandled(CK_ATTRIBUTE_TYPE),
 }
 
 impl AttributeType {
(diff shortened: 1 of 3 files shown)


cryptoki/src/object.rs
```diff
@@ -136,6 +136,8 @@ pub enum AttributeType {
     Wrap,
     /// Indicates that the key can only be wrapped with a wrapping key that has the Trusted attribute
     WrapWithTrusted,
+    /// Wraps undefined and custom attribute types.
+    Unhandled(CK_ATTRIBUTE_TYPE),
```
hug-dev, line 140: Hey @ilkerBaltaci , thanks for your PR :pray:!

I understand the idea behind the `Unhandled` types, but I think we strive to only expose in the `cryptoki` crate idiomatic structures and functions which are safe to use and abstracted away from the raw PKCS11 types. That's summarized in our README:

> All the PKCS11 items might not be implemented but everything that is implemented is safe.

I will let other maintainers give their opinions but I think  that if you have new types/algos you want to add it's best to do so in the high-level "way" where you add corresponding `cryptoki` structures/functions for those new types

cryptoki/src/object.rs
```diff
@@ -1006,6 +1017,8 @@ impl ObjectClass {
     pub const MECHANISM: ObjectClass = ObjectClass { val: CKO_MECHANISM };
     /// An OTP key object
     pub const OTP_KEY: ObjectClass = ObjectClass { val: CKO_OTP_KEY };
+    /// A profile object
+    pub const PROFILE: ObjectClass = ObjectClass { val: CKO_PROFILE };
```
hug-dev, line 1021: That is perfect to add though!

hug-dev:
I added new EcKdf options here btw, https://github.com/parallaxsecond/rust-cryptoki/pull/239/files! We could add more if needed

wiktor-k:
@hug-dev so I guess we can close this and extract the `PROFILE` parts? :thinking:

hug-dev:
Agree, @ilkerBaltaci feel free to re-open if you want to add the `PROFILE` parts. Happy to discuss the rest as well though!

PR #244, opened 2025-01-18
feat: revise EdDSA mechanism to support optional params

mcaneris:
This revises the `Mechanism::Eddsa` struct to receive a `EddsaParams` struct as an optional argument in line with the [PKCS#11 3.0 specification](https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061191).

_According to RFC 8032, CKM_EDDSA mechanism has an optional parameter, a CK_EDDSA_PARAMS structure._  

_The absence or presence of the parameter as well as its content is used to identify which signature scheme is to be used. The following table enumerates the five signature schemes defined in RFC 8032 and all supported permutations of the mechanism parameter and its content._

| Signature Scheme | Mechanism Param | phFlag | Context Data |
|------------------|-----------------|--------|--------------|
| Ed25519              | Not Required    | N/A         | N/A          |
| Ed25519ctx        | Required            | False      | Optional     |
| Ed25519ph         | Required            | True       | Optional     |
| Ed448                 | Required            | False      | Optional     |
| Ed448ph             | Required            | True       | Optional     |

The binding for `CK_EDDSA_PARAMS` is already present under the `cryptoki-sys` crate. However, there is no way currently to pass these params to the `Mechanism::Eddsa` struct. 

As the specification places significance on the absence of the parameters (resulting in curve selection), I elected to pass `EddsaParams` as an `Option` to `Mechanism::Eddsa`. Added and modified tests pass for Ed25519 signing with no parameters, as well as default `CK_EDDSA_PARAMS`.

I think the capability to pass these params is necessary to use the various curves allowed under the mechanism according to the specification.

### DIFF
--- a/cryptoki/src/mechanism/eddsa.rs
+++ b/cryptoki/src/mechanism/eddsa.rs
@@ -1,0 +1,175 @@
+//! EdDSA mechanism types
+
+use cryptoki_sys::*;
+use std::{convert::TryInto, ffi::c_void, marker::PhantomData, ptr::null_mut};
+
+/// EdDSA signature schemes.
+///
+/// The EdDSA mechanism, denoted CKM_EDDSA, is a mechanism for
+/// single-part and multipart signatures and verification for
+/// EdDSA.  This mechanism implements the five EdDSA signature
+/// schemes defined in RFC 8032 and RFC 8410.
+///
+/// For curves according to RFC 8032, this mechanism has an
+/// optional parameter, a CK_EDDSA_PARAMS structure.
+///
+/// | Signature Scheme | Mechanism Param | phFlag | Context Data |
+/// |------------------|-----------------|--------|--------------|
+/// | Ed25519          | Not Required    | N/A    | N/A          |
+/// | Ed25519ctx       | Required        | False  | Optional     |
+/// | Ed25519ph        | Required        | True   | Optional     |
+/// | Ed448            | Required        | False  | Optional     |
+/// | Ed448ph          | Required        | True   | Optional     |
+///
+/// The absence or presence of the parameter as well as its
+/// content is used to identify which signature scheme is to be
+/// used.
+#[derive(Debug, Clone, Copy)]
+pub enum EddsaSignatureScheme<'a> {
+    /// Pure EdDSA mode where the scheme is implicitly defined
+    /// by the curve.
+    Pure,
+    /// Ed25519 signature scheme without additional parameters.
+    Ed25519,
+    /// Ed25519 signature scheme with optional context-specific
+    /// data.
+    Ed25519ctx(&'a [u8]),
+    /// Ed25519 signature scheme with pre-hashing and optional
+    /// context-specific data.
+    Ed25519ph(&'a [u8]),
+    /// Ed448 signature scheme with optional context-specific data.
+    Ed448(&'a [u8]),
+    /// Ed448 signature scheme with pre-hashing and optional
+    /// context-specific data.
+    Ed448ph(&'a [u8]),
+}
+
+impl EddsaSignatureScheme<'_> {
+    /// Convert an `EddsaSignatureScheme` into the corresponding
+    /// parameters.
+    ///
+    /// This function prepares the appropriate parameters for
+    /// the mechanism based on the signature scheme variant.
+    ///
+    /// # Returns
+    ///
+    /// A pointer the mechanism-specific parameters.
+    ///
+    /// For `Pure` and `Ed25519`, this returns `null_mut()` as no
+    /// additional parameters are required. For other schemes, a
+    /// pointer to the an `CK_EDDSA_PARAMS` structure is returned.
+    pub fn into_params(&self) -> *mut c_void {
+        match self {
+            EddsaSignatureScheme::Pure | EddsaSignatureScheme::Ed25519 => null_mut(),
+            EddsaSignatureScheme::Ed448(context) | EddsaSignatureScheme::Ed25519ctx(context) => {
+                &CK_EDDSA_PARAMS {
+                    phFlag: false.into(),
+                    pContextData: context.as_ptr() as *mut _,
+                    ulContextDataLen: context
+                        .len()
+                        .try_into()
+                        .expect("usize can not fit in CK_ULONG"),
+                } as *const CK_EDDSA_PARAMS as *mut _
+            }
+            EddsaSignatureScheme::Ed448ph(context) | EddsaSignatureScheme::Ed25519ph(context) => {
+                &CK_EDDSA_PARAMS {
+                    phFlag: true.into(),
+                    pContextData: context.as_ptr() as *mut _,
+                    ulContextDataLen: context
+                        .len()
+                        .try_into()
+                        .expect("usize can not fit in CK_ULONG"),
+                } as *const CK_EDDSA_PARAMS as *mut _
+            }
+        }
+    }
+}
+
+/// EdDSA parameters.
+///
+/// The EdDSA mechanism, denoted CKM_EDDSA, is a mechanism for
+/// single-part and multipart signatures and verification for
+/// EdDSA. This mechanism implements the five EdDSA signature
+/// schemes defined in RFC 8032 and RFC 8410.
+///
+/// For curves according to RFC 8032, this mechanism has an
+/// optional parameter, a CK_EDDSA_PARAMS structure.
+///
+/// The absence or presence of the parameter as well as its
+/// content is used to identify which signature scheme is to be
+/// used.
+///
+/// | Signature Scheme | Mechanism Param | phFlag | Context Data |
+/// |------------------|-----------------|--------|--------------|
+/// | Ed25519          | Not Required    | N/A    | N/A          |
+/// | Ed25519ctx       | Required        | False  | Optional     |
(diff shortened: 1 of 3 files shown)


necmettindev (approved):

cryptoki/src/mechanism/mod.rs
```diff
@@ -902,11 +903,16 @@ pub enum Mechanism<'a> {
...
+    ///
     /// Note: EdDSA is not part of the PKCS#11 v2.40 standard and as
-    /// such may not be understood by the backend. It is included here
-    /// because some vendor implementations support it through the
-    /// v2.40 interface.
-    Eddsa,
+    /// such may not be understood by some backends. It is included
+    /// here because some vendor implementations support it through
+    /// the v2.40 interface.
+    Eddsa(Option<eddsa::EddsaParams<'a>>),
```
wiktor-k, line 915: Hm... here `None` and `Some(Default::default())` is exactly the same state... I wonder if it's too ugly to require `Default::default()` (remove the `Option`)... @hug-dev ?

cryptoki/src/mechanism/eddsa.rs
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+}
+
+impl Default for EddsaParams<'_> {
+    /// Provide a default instance of `EddsaParams`.
+    ///
+    /// This initializes `EddsaParams` with the default value
+    /// of the `CK_EDDSA_PARAMS` structure.
+    fn default() -> Self {
```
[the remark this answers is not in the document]
mcaneris: True, it could. Added derive.

cryptoki/src/mechanism/mod.rs
mcaneris: I agree the passing of an `Option` made the code more ugly, but I think the `None` case is not same as `Default` case.

In the `Default` case, although `CK_EDDSA_PARAMS` is a nullish struct (`phFlag` is `false`, `pContextData` is `null`, and `pContextDataLen` is `0`), params passed to `Eddsa` mechanism is no longer null (in other words, `pParameter` is not `null`, and `ulParameterLen` is no longer `0`).

Reading the specification, I understand that for selection of the `Ed25519` curve, `pParameter` must be `null`, and `ulParameterLen` must be `0`.

That said, I'm open to suggestions for a more elegant solution, I agree the `Option` made it ugly.

wiktor-k: > I agree the passing of an Option made the code more ugly, but I think the None case is not same as the Default case.

I don't mind `None` *that* much but just for the sake of my understanding I'd like to get bottom to the `None` non-equal to Default. I think one thing that tripped me is the manual conversion from `CK_MECHANISM` below. :point_down:

cryptoki/src/mechanism/mod.rs
```diff
@@ -1073,6 +1079,13 @@ impl From<&Mechanism<'_>> for CK_MECHANISM {
             | Mechanism::Sha512RsaPkcsPss(params) => make_mechanism(mechanism, params),
             Mechanism::RsaPkcsOaep(params) => make_mechanism(mechanism, params),
             Mechanism::Ecdh1Derive(params) => make_mechanism(mechanism, params),
+            Mechanism::Eddsa(params) => match params {
+                Some(params) => make_mechanism(mechanism, params),
+                None => CK_MECHANISM {
+                    mechanism,
+                    ..Default::default()
```
wiktor-k, line 1086: Continued:

This line made me suspicious. Maybe it's just me but I think this would be more readable (basically the same code as for `Mechanism::HkdfKeyGen`):

```suggestion
                pParameter: null_mut(),
                ulParameterLen: 0,
```

I wonder if this is just me, though, and yeah, I'm aware that this makes the code longer :sweat_smile:

cryptoki/src/mechanism/eddsa.rs
```diff
@@ -0,0 +1,48 @@
...
+/// content is used to identify which signature scheme is to be
+/// used.
+///
+/// | Signature Scheme | Mechanism Param | phFlag | Context Data |
+/// |------------------|-----------------|--------|--------------|
+/// | Ed25519          | Not Required    | N/A    | N/A          |
+/// | Ed25519ctx       | Required        | False  | Optional     |
+/// | Ed25519ph        | Required        | True   | Optional     |
+/// | Ed448            | Required        | False  | Optional     |
+/// | Ed448ph          | Required        | True   | Optional     |
```
hug-dev, line 26: As I understood the spec one has two ways to choose its curve:
* either use the RFC 8032 curves (which are the ones you listed) and use the `CK_EDDSA_PARAMS`
* or use a RFC 8410 curve and directly select it in the `CKK_EC_EDWARDS` type. Seems to be Eddsa in pure mode.

In order to remove the raw `CK_EDDSA_PARAMS` from `new`, would it work to define an enum in this file listing those 5 schemes + the pure mode option and then remove the `Option` in the `Mechanism` enum?

So something like:

```rust
pub enum EddsaSignatureScheme {
    Ed25519,
    Ed25519ctx(&[u8]),
    Ed25519ph(&[u8]),
    Ed448(&[u8]),
    Ed448ph(&[u8]),
    /// Curves according to RFC 8410, the mechanism is implicitly given by the curve, which is EdDSA in pure mode.
    PureMode,
}
```

and then the `new` method below would take this enum as argument and set the right fields!

What do you think?

cryptoki/src/mechanism/mod.rs
hug-dev: Replied in my other comment on a way to fix this, with an enum :nerd_face:!

hug-dev (commented):
Thanks for the PR!! Really appreciate the effort in the description you added and the comments too, made it easy to review!

Added a suggestion on types, would also appreciate @ionut-arm eyes to make sure all of this is correct :)

cryptoki/src/mechanism/eddsa.rs
mcaneris: I think this can be much more expressive.

On @wiktor-k's suggestion, I tried to test the schemes with SoftHSM, but the signature produced does not seem to differ based on the params (between `Ed25519`, `Ed25519ctx`, and `Ed25519ph`, for example). 

Let me try with your suggestion, and I'll get back.

mcaneris: @hug-dev I implemented the enum variants, and it removes the need to pass the raw `CK_EDDSA_PARAMS` struct to the `EddsaParams` constructor. 

But I'm thinking it still doesn't remove the need to pass an `Option<EddsaParams>` to the mechanism at the final stage, and match on `None` case specifically to set the `pParameter` of the mechanism itself to be `null`, and `ulParameterLen` to be `0`.

Because at the end, `make_mechanism` calculates the `ulParameterLen` based on the type, which will never be `0` so long as `EddsaParams` or `CK_EDDSA_PARAMS` is passed as an argument. 

- `None` case should either pass `null_mut()` or generate CK_MECHANISM directly with `null_mut()` similar to `HkdfKeyGen` 
- `Some` case should pass `EddsaParams`.

I don't see anyway around it, do you have any further suggestions?

hug-dev: > None case should either pass null_mut() or generate CK_MECHANISM directly with null_mut() similar to HkdfKeyGen

I was thinking of doing that in the end, that in the `match params` we would check if the signature scheme work is `EddsaSignatureScheme::PureMode` and if so pass null values?

mcaneris: @hug-dev sorry for the delay. I have pushed some changes to the code, to point out the issue I ran into (quoting below from the prev. message).

> Because at the end, make_mechanism calculates the ulParameterLen based on the type, which will never be 0 so long as EddsaParams or CK_EDDSA_PARAMS is passed as an argument. 

The problem is passing `pParameter: null` and `ulParameterLen: 0` [here](https://github.com/parallaxsecond/rust-cryptoki/pull/244/files#diff-368fb8c0844d89e93641b803c9f980fbef33a83b931a4eee621ac52f52cc06e1R1083-R1087). I have tested with a proprietary HSM and it expects `pParameter: null` and `ulParameterLen: 0` to select the default scheme.

To solve that, the current PR implements the option below.

> Or we could set inner field of EddsaParams to hold an Option<CK_EDDSA_PARAMS> and match on that directly generating a CK_MECHANISM rather than using make_mechanism, which seems to be the best solution, removing the need to pass an Option to the mechanism, and setting the inner Option using the EddsaSignatureScheme.

hug-dev: I think I got it! I guess you don't know about the signature scheme [here](https://github.com/parallaxsecond/rust-cryptoki/pull/244/files#diff-368fb8c0844d89e93641b803c9f980fbef33a83b931a4eee621ac52f52cc06e1R1083-R1087). I guess it could be saved in the `EddsaParams` structure so that you could check it later but it does not matter too much I think for the user.

hug-dev (approved):
Thanks for keeping with it through the reviews :muscle: !

wiktor-k (approved):
Looks really good 👌 I had a couple of nits but then decided they're not that important 😅

Thanks! 🙏

Issue #209, opened 2024-04-10
Add support for C_GetInterfaceList


frankmorgner:
In the [OpenSC](https://github.com/OpenSC/OpenSC) PKCS#11 module we are offering ABIs for both, version 2.4 *and* version 3.0. We're implementing this via `C_GetInterface`. Accessing `C_GetFunctionList` or `C_GetInterface` is not supported by `rust-cryptoki`, which currently only allows access to the *default* interface via `C_GetFunctionList` (i.e. the version 2.4 variant).

Please add wrappers for `C_GetInterface` to allow accessing all interfaces (i.e. version 3.0 as well for OpenSC).

PR #248, opened 2025-03-17
Add support for PKCS#11 3.0

Jakuje:
The PKCS#11 3.0 introduces new APIs with new function list to reach out to new functions. The 3.2 will come with yet another function list. This requires a change in the initialization. While most of the tokens now support also the 2.4 way, using new APIs is not possible through the old function list.

This is adding the support for the new functions and tests against another PKCS#11 module, fixing some incompatibilities that I found in the testsuite here. Its a bit mixed bag and it also found some issues in the kryoptic that we need to handle, such as latchset/kryoptic#182 and latchset/kryoptic#184 so opening as a draft for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,24 @@ jobs:
       - name: "Installs SoftHSM and execute tests"
         uses: ./.github/actions/ci_script
 
+  tests-kryoptic:
+    name: Run tests against Kryoptic
+    runs-on: ubuntu-latest
+    container: fedora:rawhide
+    steps:
+      - name: Install dependencies
+        run: dnf -y install git cargo clang-devel kryoptic
+      - uses: actions/checkout@v2
+      - name: Test script
+        env:
+          KRYOPTIC_CONF: /tmp/kryoptic.sql
+          TEST_PKCS11_MODULE: /usr/lib64/pkcs11/libkryoptic_pkcs11.so
+        run: |
+          RUST_BACKTRACE=1 cargo build &&
+          RUST_BACKTRACE=1 cargo build --all-features &&
+          RUST_BACKTRACE=1 cargo test
+
+
   build-msrv:
     name: MSRV - Execute CI script
     runs-on: ubuntu-latest
--- a/cryptoki/src/context/general_purpose.rs
+++ b/cryptoki/src/context/general_purpose.rs
@@ -44,6 +44,17 @@ macro_rules! check_fn {
     }};
 }
 
+macro_rules! check_30_fn {
+    ($pkcs11:expr, $func_name:ident) => {{
+        let func = paste! { $pkcs11
+            .impl_
+                .function_list_30
+                .map(|f| f.[<C_ $func_name>])
+        };
+        func.is_some()
+    }};
+}
+
 #[allow(missing_docs)]
 #[derive(Debug, Copy, Clone)]
 /// Enumeration of all functions defined by the PKCS11 spec
@@ -116,6 +127,31 @@ pub enum Function {
     GetFunctionStatus,
     CancelFunction,
     WaitForSlotEvent,
+    /* PKCS #11 3.0 */
+    GetInterfaceList,
+    GetInterface,
+    LoginUser,
+    SessionCancel,
+    MessageEncryptInit,
+    EncryptMessage,
+    EncryptMessageBegin,
+    EncryptMessageNext,
+    MessageEncryptFinal,
+    MessageDecryptInit,
+    DecryptMessage,
+    DecryptMessageBegin,
+    DecryptMessageNext,
+    MessageDecryptFinal,
+    MessageSignInit,
+    SignMessage,
+    SignMessageBegin,
+    SignMessageNext,
+    MessageSignFinal,
+    MessageVerifyInit,
+    VerifyMessage,
+    VerifyMessageBegin,
+    VerifyMessageNext,
+    MessageVerifyFinal,
 }
 
 impl Display for Function {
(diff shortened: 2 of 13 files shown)


hug-dev:
Thanks for the PR! Could `kryoptic` provide another way to test our wrapper, similarly than SoftHSM does? The more the merrier :grin:!

Some of the 3.0 compatibility was introduced with #228. I guess this would also fix #209.

Shout if you need early review or anything at this point!

Jakuje:
> Some of the 3.0 compatibility was introduced with #228. 

Right. This added mostly mechanisms which can work through the old API.  I think the new functions added are not reachable in anyway right now.

> I guess this would also fix #209.

Correct. Right now, its probably the least amount of changes to make the functions reachable from inside of rust-cryptoki. Another bunch would be implementing the accessing functions in the API (for example message based operations).

On top of that to be able to support all the versatility of the PKCS#11 3.* GetInterface API, I think we might need some different entry functions than `new()` and `new_from_self()`. Something like `new_version(version)` might do. OTOH its a question whether we really need to access the old API versions when the new function lists alias over the old ones.

Jakuje:
(The previous push had broken yaml for CI. This should fix it but it still waits for approval. It will fail for the kryoptic though)

wiktor-k:
> (The previous push had broken yaml for CI. This should fix it but it still waits for approval. It will fail for the kryoptic though)

:thinking: I think the rest of the jobs fail because of a missing `cargo fmt`. If you don't mind that'd be good to do now :sweat_smile:

Jakuje:
> > (The previous push had broken yaml for CI. This should fix it but it still waits for approval. It will fail for the kryoptic though)
> 
> 🤔 I think the rest of the jobs fail because of a missing `cargo fmt`. If you don't mind that'd be good to do now 😅

Sure, done. I always leave this for last minute and then forget :)

Jakuje:
> Could kryoptic provide another way to test our wrapper, similarly than SoftHSM does? The more the merrier 😁!

Do you mean something like the `ci.sh`? It runs a lot of other things that I did not consider relevant (formatting, clippy, building all the targets) just to run the same tests against different token, but it might be easier to have it in one place even through some duplication.

> Shout if you need early review or anything at this point!

Early review would be always welcomed if there is something conceptually wrong from your point of view or something significant missing to be added.

> Some of these test failures are quite easy to fix it seems: ...

Do you have any strong opinions how this should be addressed? The current tests do test token name, interface version and more that is inherently different in every token and now tied to the softhsm. They can be removed (bad) or made dynamic based on some environment variable (KRYOPTIC_CONF, SOFTHSM2_CONF or some special?), which is still quite fragile. Or did you have something else on your mind? 

> Others, such as unsupported functions, may take a while to make it compatible with both projects but it's definitely something worth pursuing.

These will also need some distinction of the tokens inside of the test.

> (If it takes too much time it's OK to split this PR into two as well).

I just wanted to make sure the new code paths I am adding are executed so thats why I would rather keep them together. But there are more to come :)

wiktor-k:
> Do you have any strong opinions how this should be addressed? The current tests do test token name, interface version and more that is inherently different in every token and now tied to the softhsm. They can be removed (bad) or made dynamic based on some environment variable (KRYOPTIC_CONF, SOFTHSM2_CONF or some special?), which is still quite fragile. Or did you have something else on your mind?

No particular preference. Testing for one exact string that's vendor specific is not the right thing to do anyway IMO. I guess the test was just a sanity check that we *can* get the project name. I think even making it check if the returned string is not empty is good enough. If we can find a different property to test instead that'd be stable that's even better.

In general I think it'd be great if we didn't encode any vendor specific logic in test. If there are differences it'd be nice if the code checked properties at test time and took a different path. I view tests as collection of nice examples that ideally folks could reuse with their own tokens. Yeah, they have been coded against SoftHSM which was OK at that time.

Sorry for a generic response :sweat_smile: my point was to use your best judgement. Adding another token for tests is a tremendous opportunity to make the tests a lot better! :)

Jakuje:
Thank you. Adjusted (and resolved the conflicts).

cryptoki/src/context/mod.rs
```diff
@@ -44,12 +44,14 @@ pub(crate) struct Pkcs11Impl {
     // valid.
     _pkcs11_lib: cryptoki_sys::Pkcs11,
     pub(crate) function_list: cryptoki_sys::CK_FUNCTION_LIST,
+    pub(crate) function_list_30: Option<cryptoki_sys::CK_FUNCTION_LIST_3_0>,
```
hug-dev, line 47: Would it be possible/better to have an enum here which can only be one of the two options and use `C_GetInfo` in the beginning of `new` to check which one the implementation is? That would be instead of trying to call `C_GetInterface`

cryptoki/src/session/object_management.rs
```diff
@@ -454,7 +454,11 @@ impl Session {
                 ))
             } {
                 Rv::Ok => {
-                    results.push(AttributeInfo::Available(template[0].ulValueLen.try_into()?))
+                    if template[0].ulValueLen == CK_UNAVAILABLE_INFORMATION {
+                        results.push(AttributeInfo::Unavailable)
```
hug-dev, line 458: Nice!

.github/workflows/ci.yml
```diff
@@ -41,6 +41,24 @@ jobs:
...
+    runs-on: ubuntu-latest
+    container: fedora:rawhide
+    steps:
+      - name: Install dependencies
+        run: dnf -y install git cargo clang-devel kryoptic
+      - uses: actions/checkout@v2
+      - name: Test script
+        env:
+          KRYOPTIC_CONF: /tmp/kryoptic.sql
+          PKCS11_SOFTHSM2_MODULE: /usr/lib64/pkcs11/libkryoptic_pkcs11.so
```
hug-dev, line 55: I guess you could also define another name for the env variable (like `PKCS11_KRYOPTIC_MODULE`) and then check which one is set to know if the implementation is using softHSM or kyoptic in the tests

cryptoki/tests/common.rs
```diff
@@ -11,12 +11,17 @@ pub static USER_PIN: &str = "fedcba";
 // The default SO pin
 pub static SO_PIN: &str = "abcdef";
 
+fn get_pkcs11_path() -> String {
+    env::var("PKCS11_SOFTHSM2_MODULE")
+        .unwrap_or_else(|_| "/usr/local/lib/softhsm/libsofthsm2.so".to_string())
+}
+
+pub fn is_softhsm() -> bool {
```
hug-dev, line 19: Referring to my other comment, this could check which one of the env variable is set, `PKCS11_SOFTHSM2_MODULE` or `PKCS11_KRYOPTIC_MODULE`

cryptoki/tests/basic.rs
```diff
@@ -1374,15 +1421,27 @@ fn rsa_pkcs_oaep_empty() -> TestResult {
 
 #[test]
 #[serial]
-#[ignore] // it's not clear why the test with data specified fails
 fn rsa_pkcs_oaep_with_data() -> TestResult {
+    /* SoftHSM does not support additional OAEP Source */
```
hug-dev, line 1425: Amazing that this can now be tested!

cryptoki/tests/basic.rs
```diff
@@ -988,6 +1019,11 @@ fn test_clone_initialize() {
 #[test]
 #[serial]
 fn aes_key_attributes_test() -> TestResult {
+    /* FIXME: This is now broken in Kryoptic: https://github.com/latchset/kryoptic/issues/182 */
+    if !is_softhsm() {
```
hug-dev, line 1023: or maybe a `is_kryoptic` function, or a `test_pkcs11_implementation` returning an enum... don't want to be too annoying for the tests though 😅

cryptoki/src/context/general_purpose.rs
```diff
@@ -116,6 +127,30 @@ pub enum Function {
     GetFunctionStatus,
     CancelFunction,
     WaitForSlotEvent,
+    GetInterfaceList,
```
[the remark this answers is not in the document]
Jakuje: Good idea. With 3.2 coming later it will be even more useful to have more comments to be clear what functions come from where.

.github/workflows/ci.yml
Jakuje: Here I also did not want to break everything at once so I reused what was there.  I would probably prefer some name without specific project like TEST_PKCS11_MODULE if we should change it.

cryptoki/tests/basic.rs
Jakuje: I  went with the `is_softhsm` as I do not expect the bugs in softhsm will get ever fixed anymore. I had some luck before 2021, but since then the project is mostly dead.

cryptoki/tests/common.rs
Jakuje: As mentioned in the other comment, would go ahead with something more generic rather than tying the testsuite to specific tokens. There will always have to be some skipping of tests on broken tokens, but I would like to be able to run against different (soft)tokens + modules, rather than a fixed list. Yes, we could have "default"/"other" token too, but not sure how well it would play here.

cryptoki/src/context/mod.rs
Jakuje: I think this wont work. The loaded shared object has only one symbol for PKCS#11 2.40 -- C_GetFunctionList. And the 3.0 has the two additional C_GetInterface and C_GetInterfaceList so we really can not call the C_GetInfo before getting the function list from either of the above functions. And if we call C_GetFunctionList unconditionally, the module will likely return a compatible function list with the version 2.40 in the C_GetInfo for compatibility (or at least some modules do that -- not sure though if it is the right thing to do).

Regarding the Enum, its probably a good suggestion. I initially did not want to break everything, but it might make the code cleaner if the macros and accessors will not get it more complicated. Let me try.

hug-dev: Ah thank you for the explanation, I thought that `C_GetInfo` was also one of those accessible directly. Good to try with the enum but no need if it does not get simpler :)

.github/workflows/ci.yml
hug-dev: Got it, we can change later :)

cryptoki/tests/basic.rs
hug-dev: > I do not expect the bugs in softhsm will get ever fixed anymore

ahh good to know!

cryptoki/src/context/mod.rs
Jakuje: The enum might allow us to store the function pointer only once and just cast it to the correct type based on the interface version. But it might come later.

Jakuje:
> Thanks for the updates! I am really ok with how the code looks now, you decide if you want to test with the enum or we could also improve with another PR.

I started to put something together, but its far from complete (and I got pulled into another tasks) so I would prefer to keep it as it is now and do follow-up in the next PR.

wiktor-k (approved):
Quite a lot of changes but they all seem reasonable. Thanks :+1:

PR #249, opened 2025-03-17
Set ulIvBits and more graceful error handling

jaeparker22:
Author: Jason Parker
Email: parkerje@protonmail.com

Modify GcmParams::new to return a Result rather than panic!() and auto-set ulIvBits to support those HSMs that require it, such as Thales. If the IV length is such that ulIvBits is too long, it will be set to 0 as previously. Either the HSM doesn't care and can work without it, or the operation would fail either way.

Tests using GCM params have been updated to reflect the function now returning a Result and a test has been added for the new operation with ulIvBits.


cryptoki/src/mechanism/aead.rs
```diff
@@ -53,23 +49,35 @@ impl<'a> GcmParams<'a> {
...
+                ulIvLen: match iv_len.try_into() {
+                    Ok(len) => len,
+                    Err(_e) => return Err("iv length does not fit in CK_ULONG"),
+                },
+                // Since this field isn't universally used, set it to 0 if it doesn't fit in CK_ULONG.
+                // If the HSM doesn't require the field, it won't mind; and it it does, it would break anyways.
+                ulIvBits: match iv_bit_len.try_into() {
+                    Ok(len) => len,
+                    Err(_e) => 0,
+                },
```
wiktor-k, line 69:
```suggestion
                ulIvBits: iv_bit_len.try_into().unwrap_or_default(),
```

Would that work? :thinking:

cryptoki/src/mechanism/aead.rs
```diff
@@ -32,11 +32,7 @@ impl<'a> GcmParams<'a> {
     /// be between 0 and 128.  The tag is appended to the end of the
     /// ciphertext.
     ///
-    /// # Panics
-    ///
-    /// This function panics if the length of `iv` or `aad` does not
-    /// fit into an [Ulong].
-    pub fn new(iv: &'a mut [u8], aad: &'a [u8], tag_bits: Ulong) -> Self {
+    pub fn new(iv: &'a mut [u8], aad: &'a [u8], tag_bits: Ulong) -> Result<Self, &'a str> {
```
wiktor-k, line 35: I think I'd leave the "Panics" section but rename it to "Errors" since then it'd be clear when this function fails.

While `&'a str` is not ideal (a custom error would be better) it's not super-bad IMO (but maybe the lifetime of that should be `'static` since it's just static messages? It's not tied to `iv` in any way).

cryptoki/src/mechanism/aead.rs
```diff
@@ -53,23 +49,35 @@ impl<'a> GcmParams<'a> {
...
                 pAAD: aad.as_ptr() as *mut _,
-                ulAADLen: aad
+                ulAADLen: match aad
                     .len()
-                    .try_into()
-                    .expect("aad length does not fit in CK_ULONG"),
+                    .try_into() {
+                        Ok(len) => len,
+                        Err(_e) => return Err("aad length does not fit in CK_ULONG"),
+                    },
```
wiktor-k, line 76:
```suggestion
                    .try_into().map_err(|_| "aad length does not fit in CK_ULONG")?
```

I think something like this would be shorter...

cryptoki/tests/basic.rs
```diff
@@ -1268,6 +1268,28 @@ fn sha256_digest() -> TestResult {
     Ok(())
 }
 
+#[test]
+#[serial]
+fn gcm_param_graceful_failure() -> TestResult {
+    // Try to generate GcmParams with max size IV (2^32-1)
+    // Verify that the ulIvBits doesn't cause failover
+    println!("start");
```
wiktor-k, line 1276: This can be removed.

cryptoki/tests/basic.rs
```diff
@@ -1268,6 +1268,28 @@ fn sha256_digest() -> TestResult {
...
+    println!("iv");
+    let aad = [0; 16];
+    
+    let _gcm_params = match GcmParams::new(&mut iv, &aad, 96.into()) {
+        Ok(params) => params,
+        Err(e) => { 
+            println!("{e}");
+            return Err(e.into());
+        },
+    };
```
wiktor-k, line 1288: I don't think this entire thing is necessary. See if this is better:

```suggestion
    GcmParams::new(&mut iv, &aad, 96.into())?;
```

cryptoki/tests/basic.rs
```diff
@@ -1295,7 +1317,7 @@ fn aes_gcm_no_aad() -> TestResult {
         Attribute::Encrypt(true),
     ];
     let key_handle = session.create_object(&template)?;
-    let mechanism = Mechanism::AesGcm(GcmParams::new(&mut iv, &aad, 96.into()));
+    let mechanism = Mechanism::AesGcm(GcmParams::new(&mut iv, &aad, 96.into()).unwrap());
```
wiktor-k, line 1320:
```suggestion
    let mechanism = Mechanism::AesGcm(GcmParams::new(&mut iv, &aad, 96.into())?);
```

cryptoki/tests/basic.rs
```diff
@@ -1326,7 +1348,14 @@ fn aes_gcm_with_aad() -> TestResult {
...
     ];
     let key_handle = session.create_object(&template)?;
-    let mechanism = Mechanism::AesGcm(GcmParams::new(&mut iv, &aad, 96.into()));
+    let gcm_params = match GcmParams::new(&mut iv, &aad, 96.into()) {
+        Ok(params) => params,
+        Err(e) => { 
+            println!("{e}");
+            return Err(e.into());
+        },
+    };
```
wiktor-k, line 1357:
```suggestion
    let gcm_params = match GcmParams::new(&mut iv, &aad, 96.into())?;
```

`TestResult` should take care of printing and panicking when this fails.

wiktor-k (commented):
Looks good, a couple of nits.

cryptoki/src/mechanism/aead.rs
jaeparker22: There is a TryFromIntError in the crate, which I initially thought to use and is the natural error generated. Only reason I didn't was because I wanted to maintain the original error message.

hug-dev: > Only reason I didn't was because I wanted to maintain the original error message

Do you mean your custom error message?

I think it would be perfect to use `TryFromIntError` here :smiley: I think users would understand based on the context/line number?

hug-dev (approved):
Thanks a lot! It looks all good for me, I would though be more in favour of re-using our `Error` type

wiktor-k:
Looks really great. I'd be nice if all these were squashed (since the code looks really simple now) but... I guess we don't need to block merging for that.

Thanks for sticking with us throughout the entire process. Deeply appreciated! :bow:

wiktor-k (approved):
LGTM :+1:

PR #252, opened 2025-03-27
Add bindings for multi-part operations

### LINKED ISSUE #250: Support for multi-part operations
Hello Cryptoki team!

Is there any plan to add bindings for the multi-part operations (`C_{Digest,Encrypt,Decrypt,Sign,Verify}{Init,Update,Final}`) to the library? Or conversely, is there any reason we should keep them out (say, for example, if the goal is to abstract away most of the more low-level parts of the interface)?

If so, I already have some code written for the digest operations that I can contribute immediately. I would be happy to take on adding support for the others too.

Getting ahead of myself here, but in writing this, would it be preferred to keep the implementation as bare as possible to conform to the spec (i.e. letting the device return `CKR_OPERATION_NOT_INITIALIZED` if a user tries to call an `update` function without having first called an `init` one, essentially leaving it up to human error)? Or is there some benefit in having some form of internal state machine to a `Session` so that we don't allow users to call an `update` or `final` function if the corresponding `init` function has not been called, disallowing these kinds of errors at compile-time instead of runtime? I'm not sure where to stand on this, as it makes for better developer experience, but also strays from the exact expected behaviour of these functions as defined in the spec, and obviates the need for certain errors.

jacobprudhomme:
This PR adds direct bindings for the multi-part operations `C_{Encrypt,Decrypt,Sign,Verify,Digest}{Init,Update,Final}`. Closes #250 

## Summary of Changes

* Added `encrypt_{initialize,update,finalize}()` to `session/encryption.rs`
* Added `decrypt_{initialize,update,finalize}()` to `session/decryption.rs`
* Added `{sign,verify}_{initialize,update,finalize}()` to `session/signing_macing.rs`
* Added `digest_{initialize,update,finalize}()` to `session/digesting.rs`
* Added tests for all of the above to `tests/basic.rs`

## Questions/Feedback Wanted

* Are there any strong feelings about the naming (e.g. would `initialize_encrypt()` be preferred over `encrypt_initialize()`?)? I did things this way so that it was more discoverable, in that a user could start typing `encrypt` and would see all of the encrypt operations grouped together
* I added tests for the sad-path cases of calling an update/finalize operation before it was initialized and initializing an operation that's already in progress. Should I delete these, given that this is an implementation detail of the underlying PKCS#11 device and not this library itself?


hug-dev (commented):
This looks really good, good job :star: ! Very nice with all the tests that you added too! Definitely a good step to have in the code and then we can decide later if we want to add more typing to check the multi-part sequence at compile time.

> Are there any strong feelings about the naming

I like your thinking about discoverability! My only opinion would be to match with the PKCS11 shorter names and have for example with `decrypt`: `decrypt_init` and `decrypt_final`?

> Should I delete these, given that this is an implementation detail of the underlying PKCS#11 device and not this library itself?

I don't think so! Those are really good and are actually not implementation details since they are written in the spec:

> The message-digesting operation MUST have been initialized with C_DigestInit.

so perfectly sensible to check that :)

> Do any version number changes need to be made manually?

We will do that ourselves when we decide to make a new release!

> Should I add myself to [CONTRIBUTORS.md](https://github.com/parallaxsecond/parsec/blob/master/CONTRIBUTORS.md) as mentioned [here](https://parallaxsecond.github.io/parsec-book/contributing/index.html)?

No requirement on our side! That is only if you want it :) I think the original reason is to have people behind what the copyright notice says.

jacobprudhomme:
@hug-dev thanks for the feedback! I changed the function names as you suggested and applied the lint fixes that were preventing CI from passing :)

hug-dev (approved):
Thanks!

jacobprudhomme:
Sorry, it seems there was another CI error to fix. Should all be good now!

hug-dev (approved):

wiktor-k (approved):
This looks very clean and nice, thank you! :+1: 

Would you mind rebasing this on top of `main`? We recently landed a big PR and I wonder if it'll all work together nicely :)

(if @hug-dev want to merge it and potentially collect the pieces of the merge I'm fine with that too :sweat_smile: )

jacobprudhomme:
Just rebased off of `main`!

wiktor-k:
Ugh... it seems I screwed it up this time by pressing "Enable auto-merge" but `Run tests against Kryoptic` is not Required for merging... ugh... maybe I'll just... :face_in_clouds:

wiktor-k:
> Should I open another PR to fix this?

If you could that'd be really appreciated 🙏

hug-dev:
Added the missing CI checks as required now!

jacobprudhomme:
Sorry for the delay in getting it up! I opened a PR at #253

PR #253, opened 2025-03-31
Fixed failing tests for multi-part operations

jacobprudhomme:
Author: Jacob Prud'homme

# Description

In tests relating to multi-part operations, added missing attributes to key templates, since Kryoptic was throwing errors about said keys not having the correct attributes set on them to perform certain operations (see [here](https://github.com/parallaxsecond/rust-cryptoki/pull/252#issuecomment-2761295928))

## Summary of Changes

* Tweaked tests for multipart encrypt/decrypt/verify/sign/digest functionality
  * Added missing attributes to templates for keys
  * Asserted that only a generic PKCS#11 error happens for the `*_not_initialized` tests, since the exact error returned by various backends is different
  * Similarly, changed the `*_already_initialized` tests to account for differing behaviour between backends
  * Ignored `sha256_digest_key` test, since it seems to be impossible to test the desired functionality on all backends for the moment (in particular, extracting the value from a secret key marked `Extractable` and not `Sensitive` is not always supported)

### DIFF
--- a/cryptoki/tests/basic.rs
+++ b/cryptoki/tests/basic.rs
@@ -23,6 +23,8 @@ use std::thread;
 use cryptoki::mechanism::ekdf::AesCbcDeriveParams;
 use testresult::TestResult;
 
+const AES128_BLOCK_SIZE: usize = 128 / 8;
+
 #[test]
 #[serial]
 fn sign_verify() -> TestResult {
@@ -446,7 +448,7 @@ fn encrypt_decrypt_multipart() -> TestResult {
     let template = vec![
         Attribute::Token(true),
         Attribute::Private(false),
-        Attribute::ValueLen((128 / 8).into()),
+        Attribute::ValueLen((AES128_BLOCK_SIZE as u64).into()),
         Attribute::Encrypt(true),
         Attribute::Decrypt(true),
     ];
@@ -455,14 +457,16 @@ fn encrypt_decrypt_multipart() -> TestResult {
     // Data to encrypt
     let data = vec![
         0xFF, 0x55, 0xDD, 0x11, 0xBB, 0x33, 0x99, 0x77, 0xFF, 0x55, 0xDD, 0x11, 0xBB, 0x33, 0x99,
-        0x77,
+        0x77, 0xFF, 0x55, 0xDD, 0x11, 0xBB, 0x33, 0x99, 0x77, 0xFF, 0x55, 0xDD, 0x11, 0xBB, 0x33,
+        0x99, 0x77, 0xFF, 0x55, 0xDD, 0x11, 0xBB, 0x33, 0x99, 0x77, 0xFF, 0x55, 0xDD, 0x11, 0xBB,
+        0x33, 0x99, 0x77,
     ];
 
-    // Encrypt data in parts
+    // // Encrypt data in parts
     session.encrypt_init(&Mechanism::AesEcb, key)?;
 
     let mut encrypted_data = vec![];
-    for part in data.chunks(3) {
+    for part in data.chunks(AES128_BLOCK_SIZE) {
         encrypted_data.extend(session.encrypt_update(part)?);
     }
     encrypted_data.extend(session.encrypt_final()?);
@@ -471,7 +475,7 @@ fn encrypt_decrypt_multipart() -> TestResult {
     session.decrypt_init(&Mechanism::AesEcb, key)?;
 
     let mut decrypted_data = vec![];
-    for part in encrypted_data.chunks(3) {
+    for part in encrypted_data.chunks(AES128_BLOCK_SIZE) {
         decrypted_data.extend(session.decrypt_update(part)?);
     }
     decrypted_data.extend(session.decrypt_final()?);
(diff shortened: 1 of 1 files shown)


jacobprudhomme:
It seems a host of new errors has popped up now. I'll keep exploring

wiktor-k:
I'll CC @Jakuje just in case they have some thoughts on the Kryoptic issues. kthxbai :bow:

Jakuje:
I think we had some issues with the multipart operations recently. They were fixed in main, but not in any release version yet: latchset/kryoptic#179

I will see if we will get a new release out (and update in Fedora) or I can backport the changes to Fedora later today. Sorry for inconvenience! I think the issue is not on your side!

wiktor-k:
> I think we had some issues with the multipart operations recently. They were fixed in main, but not in any release version yet: [latchset/kryoptic#179](https://github.com/latchset/kryoptic/issues/179)
> 
> I will see if we will get a new release out (and update in Fedora) or I can backport the changes to Fedora later today. Sorry for inconvenience! I think the issue is not on your side!

Thank you. :pray:  It's great to hear your remarks directly.

jacobprudhomme:
Fixed the tests! I had to ignore one to get all of them to pass. I'm open to suggestions on how to rewrite that test, though. I already tried manually creating a key with a manually set value and using that to compare against, but it seems Kryoptic doesn't allow using a key created this way to be used in `C_DigestKey`

cryptoki/tests/basic.rs
```diff
@@ -437,21 +449,26 @@ fn encrypt_decrypt_multipart() -> TestResult {
...
     let data = vec![
         0xFF, 0x55, 0xDD, 0x11, 0xBB, 0x33, 0x99, 0x77, 0xFF, 0x55, 0xDD, 0x11, 0xBB, 0x33, 0x99,
-        0x77,
+        0x77, 0xFF, 0x55, 0xDD, 0x11, 0xBB, 0x33, 0x99, 0x77, 0xFF, 0x55, 0xDD, 0x11, 0xBB, 0x33,
+        0x99, 0x77, 0xFF, 0x55, 0xDD, 0x11, 0xBB, 0x33, 0x99, 0x77, 0xFF, 0x55, 0xDD, 0x11, 0xBB,
+        0x33, 0x99, 0x77,
     ];
 
-    // Encrypt data in parts
+    // // Encrypt data in parts
```
wiktor-k, line 467: This is not necessary methinks.

wiktor-k (commented):
Okay, I think it looks good. I guess the ignored test could check for softhsm the same way as some other tests (it's hard to check on mobile for me now).

Let's leave it open for a while as @Jakuje may have an idea on how to better adjust it.

cryptoki/tests/basic.rs
jacobprudhomme: Oops, no idea how _that_ got in there

Jakuje:
I tried to step through this in GDB (running main branch) and for some reason, it looks like this gives 0-length attribute list when the code reaches kryoptic:
```
#0  kryoptic_pkcs11::fn_get_attribute_value (s_handle=2, o_handle=1, template=0x8, count=0) at src/lib.rs:1027
#1  0x000055555560c732 in cryptoki::session::Session::get_attributes (self=0x7ffff7c9bd00, object=..., 
    attributes=&[cryptoki::object::AttributeType](size=1) = {...}) at cryptoki/src/session/object_management.rs:537
```
While I clearly see that the previous frame has there the one attribute we are looking for, which sounds like some bug in rust-cryptoki, but I will have to dig into that deeper to see what is going on there.

Jakuje:
Indeed, the pkcs11-spy run confirms this:
```
13: C_GetAttributeValue
P:380116; T:0x140503810721472 2025-03-31 19:20:30.504
[in] hSession = 0x2
[in] hObject = 0x1
[in] pTemplate[0]: 
[out] pTemplate[0]: 
Returned:  0 CKR_OK
```

Jakuje:
(but still odd that the rust-cryptoki tries to call it without any template)

Ok, I see the bug. With the optimizations, we do not pull all attributes from db for get_attributes and the cka_sensitive defaults to true. Filled in latchset/kryoptic#193 -- I will try to have a look into that.

For the time being, feel free to skip this test for kryoptic, similarly as I skipped some others (with the link to the above issue). Once I will have a fix and new release, we can remove the exclusion.

wiktor-k:
> For the time being, feel free to skip this test for kryoptic, similarly as I skipped some others (with the link to the above issue). Once I will have a fix and new release, we can remove the exclusion.

Hah, it seems including Kryoptic makes for a nice cross-testing experience. Great that we have you here to help debug the issues.

@jacobprudhomme Would you be so kind to add a simple `if !softhsm { return }` code branch there? I think otherwise this is good to go (unless @hug-dev spots something :sweat_smile: ).

jacobprudhomme:
Alright, done!

wiktor-k (approved):
Much appreciated :pray:

hug-dev (approved):
Thank you!

PR #255, opened 2025-04-04
Add support for message-based encryption and decryption (PKCS#11 3.0)

Jakuje:
Toward this change, I had to go through some hops.
 * Update the headers to PKCS#11 3.1
 * Fix the function list initialization to be able to use the new PKCS#11 3.0 functions
 * Added the new API functions
 * Added tests (work kryoptic)


Jakuje:
The clippy failure is something in the code that I did not touch so I assume rust/clippy was updated somewhere and new warnings are showing up. On the first sight, its not clear to me how to correctly fix it.

cryptoki/src/mechanism/mod.rs
```diff
@@ -986,6 +989,7 @@ impl Mechanism<'_> {
             Mechanism::AesKeyWrap => MechanismType::AES_KEY_WRAP,
             Mechanism::AesKeyWrapPad => MechanismType::AES_KEY_WRAP_PAD,
             Mechanism::AesGcm(_) => MechanismType::AES_GCM,
+            Mechanism::AesGcmMessage(_) => MechanismType::AES_GCM,
```
hug-dev, line 992: Should there be a corresponding `MechanismType::AES_GCM_MESSAGE`?

hug-dev (commented):
Thank you!!

I think the clippy warning is only about aligning lists with two spaces instead of more

cryptoki/src/mechanism/mod.rs
Jakuje: That is the odd part that the same mechanism is used with different parameter (and different functions), see the example in https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203302

If it would be more clear, we can do "phantom" mechanism for the message based parameter that would map to the same PKCS#11 mechanism though. But not sure if it would be easier, which is why I did not do that.

cryptoki/src/context/mod.rs
```diff
@@ -38,31 +38,51 @@ use std::ptr;
...
 
+/// Enum for various function lists
+/// Each following is super-set of the previous one with overlapping start so we store them
+/// in the largest one so we can reference also potentially NULL/non-existing functions
+#[derive(Debug)]
+enum FunctionList {
+    /// PKCS #11 2.40 CK_FUNCTION_LIST
+    V2(cryptoki_sys::CK_FUNCTION_LIST_3_0),
+    /// PKCS #11 3.0 CK_FUNCTION_LIST_3_0
+    V3_0(cryptoki_sys::CK_FUNCTION_LIST_3_0),
```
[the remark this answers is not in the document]
Jakuje: Correct. No new functions in 3.1 so no new mechanism list.

Jakuje:
> I think the clippy warning is only about aligning lists with two spaces instead of more

Yeah, thats how much I got from the message, but its the continuation of the list item so I am not sure how it will render in the docs when I will remove the indent for the list. But I can try. It will break at worst :)

Jakuje:
> > I think the clippy warning is only about aligning lists with two spaces instead of more
> 
> Yeah, thats how much I got from the message, but its the continuation of the list item so I am not sure how it will render in the docs when I will remove the indent for the list. But I can try. It will break at worst :)

Turned out it just needed one less space than I would expect in the RSA case (and few less than in the other). Fixed in the last commit.

cryptoki/src/mechanism/mod.rs
hug-dev: I see, thanks for the precision! I think it's fine to keep it like this if that is consistent with the spec :)

hug-dev (dismissed):

cryptoki/src/mechanism/aead.rs
```diff
@@ -89,3 +89,105 @@ impl<'a> GcmParams<'a> {
         self.inner.ulTagBits.into()
     }
 }
+
+/// The GCM generator function for the Initialization Vector
+#[derive(Debug, Clone, Copy)]
+pub enum GeneratorFunction {
+    /// CKG_NO_GENERATE no IV generation is done.
```
wiktor-k, line 96:
```suggestion
    /// `CKG_NO_GENERATE` no IV generation is done.
```

This is truly minor but I like to use backticks around stuff like this so that the look okay in the generated docs.

cryptoki/src/mechanism/mod.rs
```diff
@@ -805,6 +805,9 @@ pub enum Mechanism<'a> {
     AesKeyWrapPad,
     /// AES-GCM mechanism
     AesGcm(aead::GcmParams<'a>),
+    /// AES-GCM mechanism with message based API and parameters
+    /// TODO should we reuse the AesGcm and use Option<> to select parameter?
```
wiktor-k, line 809: I think the TODO should use normal comment (`//`) not to be rendered as docs :thinking:

cryptoki/src/mechanism/mechanism_info.rs
```diff
@@ -228,6 +231,27 @@ impl MechanismInfo {
     pub fn ec_compressed(&self) -> bool {
         self.flags.contains(MechanismInfoFlags::EC_COMPRESS)
     }
+
+    /// True if the mechanism can be used to encrypt messages
+    ///
+    /// See [`Session::encrypt_message`](crate::session::Session::encrypt_message)
```
wiktor-k, line 237: I wonder why this line couldn't just be:

```suggestion
    /// See [`Session::encrypt_message`].
```

Is it because `Session` is not in scope? (then it can be imported above with `#[cfg(doc)]` attribute) (we do something similar [in signstar](https://gitlab.archlinux.org/archlinux/signstar/-/merge_requests/202/diffs#07c064b304098b8d77cafe580429e647de21dc76_12_11))

cryptoki/src/mechanism/mechanism_info.rs
```diff
@@ -228,6 +231,27 @@ impl MechanismInfo {
...
+    pub fn message_encrypt(&self) -> bool {
+        self.flags.contains(MechanismInfoFlags::MESSAGE_ENCRYPT)
+    }
+
+    /// True if the mechanism can be used to decrypt encrypted messages
+    ///
+    /// See [`Session::decrypt`](crate::session::Session::decrypt_message)
```
wiktor-k, line 244: The below is important to catch stuff like here (`decrypt` vs `decrypt_message`).

cryptoki/src/mechanism/mechanism_info.rs
```diff
@@ -228,6 +231,27 @@ impl MechanismInfo {
...
+    pub fn message_decrypt(&self) -> bool {
+        self.flags.contains(MechanismInfoFlags::MESSAGE_DECRYPT)
+    }
+
+    /// True if the mechanism can be used with encrypt/decrypt_bessage_begin API.
```
wiktor-k, line 249: `message` typo

cryptoki/src/session/message_decryption.rs
```diff
@@ -0,0 +1,142 @@
...
+                aad.len().try_into()?,
+                encrypted_data.as_ptr() as *mut u8,
+                encrypted_data.len().try_into()?,
+                data.as_mut_ptr(),
+                &mut data_len,
+            ))
+            .into_result(Function::DecryptMessage)?;
+        }
+
+        data.resize(data_len.try_into()?, 0);
```
wiktor-k, line 71: Is this call necessary if the vec is already created with correct size in the first place? :thinking:

cryptoki/tests/basic.rs
```diff
@@ -1862,6 +1862,123 @@ fn aes_gcm_with_aad() -> TestResult {
     Ok(())
 }
 
+#[test]
+#[serial]
+fn encrypt_decrypt_gcm_message_no_aad() -> TestResult {
+    let (pkcs11, slot) = init_pins();
+    // PKCS#11 3.0 API is not supported by this token. Skip
+    if !pkcs11.is_fn_supported(Function::MessageEncryptInit) {
+        /* return Ignore(); */
```
wiktor-k, line 1871: I think it'd be good to add a `println` here saying that the test is ignored or something (just to give a way for the dev to see what was executed or not).

cryptoki/tests/basic.rs
```diff
@@ -1862,6 +1862,123 @@ fn aes_gcm_with_aad() -> TestResult {
...
+    Ok(())
+}
+
+#[test]
+#[serial]
+fn encrypt_decrypt_gcm_message_with_aad() -> TestResult {
+    let (pkcs11, slot) = init_pins();
+    // PKCS#11 3.0 API is not supported by this token. Skip
+    if !pkcs11.is_fn_supported(Function::MessageEncryptInit) {
+        /* return Ignore(); */
```
wiktor-k, line 1930: The same here.

wiktor-k (approved):
LGTM :+1: thanks! A couple of smaller things :)

cryptoki/src/mechanism/mechanism_info.rs
Jakuje: I just copy-pasted what other flags have to make it consistent. We can do something you proposed in separate issue or PR if it can be simplified, though, as it would be touching more parts not related to the changes here.

cryptoki/src/session/message_decryption.rs
Jakuje: I used the same approach as in the `Session::decrypt()` as tried and tested :)

I agree that this might not be necessary for the PKCS#11 modules that return the right value on the first try, but the PKCS#11 specs says that the value does not have to be exact and could be larger so shrinking the vector is probably a good thing to do here:

> 1.     If pBuf is NULL_PTR, then all that the function does is return (in *pulBufLen) a number of bytes which would suffice to hold the cryptographic output produced from the input to the function.� This number may somewhat exceed the precise number of bytes needed, but should not exceed it by a large amount.� CKR_OK is returned by the function.

From [5.2 Conventions for functions returning output in a variable-length buffer](
https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203243)

cryptoki/tests/basic.rs
wiktor-k: Yes, this (along others) is one of the issues I have with `cargo test`.

FWIW it seems the issue is tracked at https://github.com/rust-lang/rust/issues/68007.

cryptoki/src/session/message_decryption.rs
Jakuje: No, it should work only the other way round -- the size query can return larger value, but the operation itself will return the right value so no buffer should overrun in this code, unless I miss something.

cryptoki/tests/basic.rs
Jakuje: I hope I will be able to remove the other skips as soon as I will update kryptic with the recent fixes so I changed it only now in these two cases I added.

wiktor-k (approved):

hug-dev (approved):

PR #257, opened 2025-04-14
Feat: Add NIST SP800-108 KDF mechanisms

jacobprudhomme:
Author: Jacob Prud'homme

# Description

This PR adds support for the NIST SP800-108 KDF (aka KBKDF) in counter-, feedback- and double pipeline-modes

## Summary of Changes

* Added `SP800_108_{COUNTER,FEEDBACK,DOUBLE_PIPELINE}_KDF` mechanism types
  * Also, their corresponding `Mechanism` variants and associated parameters (`KbkdfParams` for counter- and double pipeline-mode, `KbkdfFeedbackParams` for feedback-mode)
  * Also, the types used within these parameter types (`PrfDataParam`, `KbkdfDkmLengthFormat`, `KbkdfDkmLengthType`, `KbkdfCounterFormat`, `DerivedKey`)
  * Also, tests for each mechanism (don't be afraid of the size of the diff! This is where the largest part of the changes comes from)
* Updated warning on pointer casts in [`mechanism::make_mechanism()`](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/src/mechanism/mod.rs#L1252)
* Unrelated lint errors that popped up (not sure where they came from...)

# Design Justifications

Since there's a decent amount of changes in this PR, I figured I'd start by justifying some of the choices I made, to preemptively answer any questions that might pop up.

1. I didn't make [`KbkdfParams`](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/src/mechanism/kbkdf.rs#L233)/[`KbkdfFeedbackParams`](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/src/mechanism/kbkdf.rs#L302) transparent wrappers around the corresponding PKCS#11 struct, because it simply wasn't possible. Since we need to pass a `&[Attribute]` to the [constructor for `DerivedKey`](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/src/mechanism/kbkdf.rs#L195) (unless we want to expose the end-user to the internal `CK_ATTRIBUTE` struct), and `Attribute` itself isn't a transparent wrapper around `CK_ATTRIBUTE`, we have to convert to a slice of `CK_ATTRIBUTE`s and store it internally, so that that contiguous section of memory stays valid for the needed duration. The consequences of this choice bubble up to the top level, and requires that the `Kbkdf[Feedback]Params` themselves can not be transparent for the same reason. Because of this, I had to add an accessor for the wrapped PKCS#11 type to get things to work with `make_mechanism()`. I have suggestions below about making this more flexible for the future
2. I've settled on mutating the original `DerivedKey` structs the end-user creates to update them with the handle of the created key, and so it is up to that user to go back and collect these `ObjectHandle`s for the additional derived keys after calling `derive_key()` (see [this test](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/tests/basic.rs#L2479) for an example). While slightly less convenient, this more closely matches the paradigm of the spec itself (where pointers to handles are passed in via the params, and these are then mutated when the corresponding key is derived). The `additional_derived_keys` argument is also correspondingly marked `mut`, to signal that it will be mutated on the PKCS#11 backend
   1. As you can see in commit f1a15f7f375b1869a2dd1a437c69d3eff791eee2, my prior solution involved creating a new `derive_keys()` function, not present in the spec, as well as a `HasAdditionalDerivedKeys` trait to allow accessing such data on arbitrary param types. This clunky solution was significantly less attractive too
3. I wrapped the internal boxed slices in [`DerivedKey`](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/src/mechanism/kbkdf.rs#L184) and [`KbkdfParams`](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/src/mechanism/kbkdf.rs#L233)/[`KbkdfFeedbackParams`](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/src/mechanism/kbkdf.rs#L302) in a `Pin`. While pinning a box does nothing, I figured it was important to mark that this data should have a stable address, since the C side relies on pointers to it remaining valid

# Questions/Feedback Wanted

## General Questions

- [ ] When looking at the HKDF implementation as an example, there are a bunch of accessors on [`HkdfParams`](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/src/mechanism/hkdf.rs#L36) that return different parts of the internal data. I'm unsure of their purpose, however. Should I also add these to `Kbkdf[Feedback]Params`?

## Coding Style Questions

- [ ] The [`Endianness`](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/src/mechanism/kbkdf.rs#L24) enum is quite general, should I place this somewhere else?
- [ ] The tests I wrote are quite big, I'd be happy to break them up if deemed necessary

## Design Questions

- [ ] I named all external-facing types as `Kbkdf(...)` instead of something like `Sp800108(...)`, because the naming of the latter is atrocious. However, the term KBKDF is never used in the spec and may lead to people not easily discovering the functionality they need. How do you feel about this naming?
- [ ] As you can see in commit 8ac48ed215e1ce056ae34321fcb0d84bcf677966, I previously had an implementation with separate `PrfCounterDataParam` and `PrfDataParam` types, as there are some differences between the data parameters allowed by the KDF in counter-mode vs the other modes. Notably, in counter-mode, `IterationVariable` must take a `KbkdfCounterFormat` argument, and the `Counter` data parameter type is not allowed. While distinguishing these two types of params ensures the end-user could not make mistakes when configuring the KDF, it goes against the spec, which has just a single `CK_PRF_DATA_PARAM` type. Would the type-safety focus of the original design be preferred over the current 1-to-1 mapping with the spec? In the latter case, the backend should return `CKR_MECHANISM_PARAM_INVALID` for these types of errors anyway, but it may be nice to stop these errors at the type system-level

# Suggestions for Future

Both of these came up while trying to work around the limitations I ran into on [Design Justifications (1)](#design-justifications).

1. What if we made `Attribute` a transparent wrapper around `CK_ATTRIBUTE`? That way we could get rid of the inconsistencies between constructing a `Kbkdf[Feedback]Params` vs other parameter types, i.e. have a tree of just transparent wrappers all the way to the root params object, which we can directly pass as a pointer to the C side, as was done up until now
2. Solving _part_ of the same issue as above, what if we made some sort of trait `AsPkcsParams` defining a `as_pkcs_params_ptr()` method with the following default implementation that covers 99% of the cases so far (i.e. all cases except this one):
```rust
fn as_pkcs_params_ptr(&self) -> *mut c_void {
    self as *const T as *mut c_void
}
```
While this partially overlaps with (1) in terms of solving some inconsistencies that I've had to introduce to make my changes fit, I think it has other benefits. While not so important because `make_mechanism()` is internal, it would still prevent developers from making a mistake and enforce that the function is only used with parameter objects that explicitly implement the trait, instead of arbitrary types. As well, the impact of implementing this is quite small. I could even make that a part of this PR, if you're interested!


cryptoki/src/mechanism/kbkdf.rs
```diff
@@ -0,0 +1,371 @@
...
+
+impl KbkdfCounterFormat {
+    /// Construct encoding format for KDF's internal counter variable.
+    ///
+    /// # Arguments
+    ///
+    /// * `endianness` - The endianness of the counter's bit representation.
+    ///
+    /// * `width_in_bits` - The number of bits used to represent the counter value.
+    pub fn new(endianness: Endianness, width_in_bits: usize) -> Self {
```
wiktor-k, line 45: Does `width_in_bits` have any value constraints? Maybe using [`NonZeroUsize`](https://doc.rust-lang.org/stable/std/num/type.NonZeroUsize.html) would be a good idea? :thinking:

cryptoki/src/mechanism/kbkdf.rs
```diff
@@ -0,0 +1,371 @@
...
+        dkm_length_method: KbkdfDkmLengthMethod,
+        endianness: Endianness,
+        width_in_bits: usize,
+    ) -> Self {
+        Self(CK_SP800_108_DKM_LENGTH_FORMAT {
+            dkmLengthMethod: match dkm_length_method {
+                KbkdfDkmLengthMethod::SumOfKeys => CK_SP800_108_DKM_LENGTH_SUM_OF_KEYS,
+                KbkdfDkmLengthMethod::SumOfSegments => CK_SP800_108_DKM_LENGTH_SUM_OF_SEGMENTS,
+            },
+            bLittleEndian: (endianness == Endianness::Little).into(),
```
wiktor-k, line 93: I wonder if we want to have a `From<Endianness> for bool` conversion so that all thess could be replaced with `endianness.into()` :thinking:

cryptoki/src/mechanism/kbkdf.rs
```diff
@@ -0,0 +1,371 @@
...
+            },
+            _marker: PhantomData,
+        }
+    }
+}
+
+/// Container for information on an additional key to be derived.
+#[derive(Debug)]
+pub struct DerivedKey {
+    template: Pin<Box<[CK_ATTRIBUTE]>>,
```
wiktor-k, line 182: Just out of curiosity why this is pinned?

cryptoki/src/mechanism/kbkdf.rs
```diff
@@ -0,0 +1,371 @@
...
+    }
+}
+
+/// NIST SP 800-108 (aka KBKDF) counter and double pipeline-mode parameters.
+///
+/// This structure wraps a `CK_SP800_108_KDF_PARAMS` structure.
+#[derive(Debug)]
+pub struct KbkdfParams<'a> {
+    /// Holds own data so that we have a contiguous memory region to give to backend
+    _additional_derived_keys: Option<Pin<Box<[CK_DERIVED_KEY]>>>,
```
wiktor-k, line 232: Why the underscore?

cryptoki/src/mechanism/kbkdf.rs
```diff
@@ -0,0 +1,371 @@
...
+
+    inner: CK_SP800_108_KDF_PARAMS,
+    /// Marker type to ensure we don't outlive the data
+    _marker: PhantomData<&'a [u8]>,
+}
+
+impl<'a> KbkdfParams<'a> {
+    /// Construct parameters for NIST SP 800-108 KDF (aka KBKDF) pseuderandom function-based key
```
wiktor-k, line 240:
```suggestion
    /// Construct parameters for NIST SP 800-108 KDF (aka KBKDF) pseudorandom function-based key
```

cryptoki/tests/basic.rs
```diff
@@ -2170,6 +2175,796 @@ fn ekdf_aes_cbc_encrypt_data() -> TestResult {
...
+    // Derive key
+    let derived_key_feedback_iv = session.derive_key(
+        &Mechanism::KbkdfFeedback(params),
+        base_key,
+        &derived_key_template,
+    )?;
+
+    /* DOUBLE PIPELINE-MODE */
+
+    /* FIXME: NIST SP800-108 in double-pipeline mode is not yet supported by SoftHSM or Kryoptic */
```
wiktor-k, line 2276: I think it'd be good to move all these unsupported cases into a separate test and mark them with `#[ignore = "unsupported by both softhsm and kryoptic"]`.

wiktor-k (approved):
I think it looks good. It's quite a sizable change :sweat_smile: 

The `make_mechanism` made me think of https://github.com/parallaxsecond/rust-cryptoki/issues/225 :thinking:

cryptoki/src/mechanism/kbkdf.rs
jacobprudhomme: The spec doesn't explicitly say anything: https://docs.oasis-open.org/pkcs11/pkcs11-curr/v3.0/os/pkcs11-curr-v3.0-os.html#_Toc30061446, but I think it's logical that if you didn't want a `Counter` or `DkmLength` format object to be present in the input data, you would just leave it out, instead of providing one with length 0. I'll make this change

cryptoki/src/mechanism/kbkdf.rs
jacobprudhomme: That feels a little wrong to me because there's no obvious equivalence between `Little` and `true` or `Big` and `false` 🤔 If some other type were to expect `true` for big endian instead (say, because it was a more obvious default), then this would break down. What do you think?

cryptoki/src/mechanism/kbkdf.rs
jacobprudhomme: Since we don't use this member, the lint check complains without it. It's just an internal member needed to hold the slice of additional key PKCS#11 structs so that it remains valid for the lifetime of the params object (this is a consequence of what I explained in the Design Justifications section of the PR body)

cryptoki/src/mechanism/kbkdf.rs
wiktor-k: > What do you think?

Good points! As a middle-ground maybe we could have something like this:

```rust
impl Endianness {
  /// Converts this instance into a boolean acceptable in `bLittleEndian` parameters.
  pub(crate) is_little_endian(&self) -> the param {
    *self == Endianness::Little
  }
}
```

`pub(crate)` means this would not leak and the code would be simpler (all comparisons to Little removed).

If you feel it's too much then I'm not bent into making it happen, too :sweat_smile: Hmm... the more I think about it the more unsure I am if this is even a good idea...

Maybe we can just resolve this conversation as if it never happened :face_in_clouds:

cryptoki/tests/basic.rs
jacobprudhomme: Done!

jacobprudhomme:
@wiktor-k glad to know I was on the same wavelength as previous work using mutable parameters!

Also, did you have any comments/remarks on what I've written in the PR body?

wiktor-k:
> Also, did you have any comments/remarks on what I've written in the PR body?

Tbh I want to calmly re-read it as I didn't have much time for open-source reviews today. I'll add a more concrete comment soon.

jacobprudhomme:
@wiktor-k sorry to ping again, but do you know when you might have time for a re-review?

wiktor-k (dismissed):
> @wiktor-k sorry to ping again, but do you know when you might have time for a re-review?

No, it's good. I've got tons of other things and I prefer if folks ping me.

I've re-read the diff and your description and I'm good with merging this as is. There are quite a few questions but IMO they should not block inclusion of this PR.

I'll get back to answers in a sec.

wiktor-k:
> When looking at the HKDF implementation as an example, there are a bunch of accessors on [HkdfParams](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/src/mechanism/hkdf.rs#L36) that return different parts of the internal data. I'm unsure of their purpose, however. Should I also add these to Kbkdf[Feedback]Params?

The accessors there are for providing safe access to internals of the struct. I don't think you need to add them. If anyone feels the need they can always expand the current code, it's open-source! :)

> The [Endianness](https://github.com/jacobprudhomme/rust-cryptoki/blob/bea29ba5bab8e5b70e009b3b004c44c99ec6847d/cryptoki/src/mechanism/kbkdf.rs#L24) enum is quite general, should I place this somewhere else?

I'm not sure since the use is currently limited but I'm also okay with moving it.

> The tests I wrote are quite big, I'd be happy to break them up if deemed necessary

They are quite readable. I'm more worried with special-casing more and more of our testing backends but it's not a problem for this PR.

> I named all external-facing types as Kbkdf(...) instead of something like Sp800108(...), because the naming of the latter is atrocious. However, the term KBKDF is never used in the spec and may lead to people not easily discovering the functionality they need. How do you feel about this naming?

I think you made the right choice. They're readable and thanks to doc comments you've inserted if someone searches for the ugly name they'd find it too. :+1: 

> As you can see in commit https://github.com/parallaxsecond/rust-cryptoki/commit/8ac48ed215e1ce056ae34321fcb0d84bcf677966, I previously had an implementation with separate PrfCounterDataParam and PrfDataParam types, as there are some differences between the data parameters allowed by the KDF in counter-mode vs the other modes. Notably, in counter-mode, IterationVariable must take a KbkdfCounterFormat argument, and the Counter data parameter type is not allowed. While distinguishing these two types of params ensures the end-user could not make mistakes when configuring the KDF, it goes against the spec, which has just a single CK_PRF_DATA_PARAM type. Would the type-safety focus of the original design be preferred over the current 1-to-1 mapping with the spec? In the latter case, the backend should return CKR_MECHANISM_PARAM_INVALID for these types of errors anyway, but it may be nice to stop these errors at the type system-level

I think safety is more important than the 1-1 mapping. I wonder what does @hug-dev think about it?

jacobprudhomme:
@wiktor-k thanks for the answers! I agree that maybe it's best to address any comments in follow-up PRs since it'd be nice to get this one through

cryptoki/src/mechanism/mod.rs
```diff
@@ -1213,9 +1252,11 @@ impl From<&Mechanism<'_>> for CK_MECHANISM {
...
     CK_MECHANISM {
         mechanism,
-        // SAFETY: Although the type signature says *mut, none of the
-        // mechanisms we support involve mutating the parameter, so
-        // this cast is OK.
+        // SAFETY: Parameters that expect to have some part of themselves
+        // mutated (such as additional_derived_keys in Kbkdf{*}Params) should
+        // indicate this to the end user by marking the relevant constructor
+        // parameters as mut. Otherwise, we should generally not expect the
+        // backend to mutate the parameters, so this cast is fine.
```
hug-dev, line 1259: Very nice that you added that because I think we already had some parameters that could be mutated. I am wondering if we should keep a bullet point list here to be even more transparent?

cryptoki/src/mechanism/kbkdf.rs
```diff
@@ -0,0 +1,374 @@
...
+    CK_SP800_108_ITERATION_VARIABLE, CK_SP800_108_KDF_PARAMS, CK_ULONG,
+};
+
+use crate::object::{Attribute, ObjectHandle};
+
+use super::MechanismType;
+
+/// Endianness of byte representation of data.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Endianness {
```
hug-dev, line 24: It seems that only the SP 800-108 functions use endianness as a parameter, good to keep in mind that this is here and could be lifted up to a higher level later on if needed!

cryptoki/src/mechanism/kbkdf.rs
hug-dev: Since this is a private field, I think it's fine with only the `Box` but writing the reasoning in a comment above it :)

cryptoki/src/mechanism/kbkdf.rs
hug-dev: Don't you use though on line 286 when setting its value? If so I think it's ok to remove the underscore.

hug-dev (commented):
Thank you!! The code looks very clean :ok:
Mostly minor comments on my side!

hug-dev:
> I think safety is more important than the 1-1 mapping.

Agree! Trust you with making the right choices here, since you are from all of us the one knowing the best those parts :sweat_smile: We can always adapt in the future if new contributors/users have diverging opinions!

cryptoki/src/mechanism/kbkdf.rs
jacobprudhomme: It seems because I'm deriving `Debug`, it complains: https://github.com/rust-lang/rust/issues/88900. I would have to access the field at least once to avoid the compiler warning. I'm guessing keeping the `Debug` trait is important, right?

jacobprudhomme:
Only small changes for you to review this time, thankfully!

Regarding the 1-to-1 mapping with the spec vs. being more type-safe, I decided in the end to stick with the first option. With the 2nd option, the end-user would still be able to make mistakes (i.e. initializing a `KbkdfParams` without a `PrfDataParam::IterationVariable`, which is always required), while making things more verbose and ugly imo. As a tradeoff, the doc comment on `PrfDataParam` is quite complete and describes these rules, so people will at least have the relevant information accessible right away.

hug-dev (approved):

hug-dev (approved):

PR #258, opened 2025-04-14
Remove skipping of kryoptic tests as they should work now

Jakuje:
When I introduced kryoptic tests, some of them were failing because of kryoptic bugs. These were fixed now and the skipping should not be needed anymore. I just updated the kryoptic package in Fedora, but it might take some time before it will land in mirrors so it might need some rerun later. I will keep an eye on that.

Contains also fix for the clippy as the #255 is not yet merged (this fix should have been a separate PR ...).


wiktor-k (approved):
Great to see Kryoptic growing more and more features 🚀

hug-dev (approved):

PR #261, opened 2025-04-18
Bump libloading version to 0.8.6

Jakuje:
the libloading version 0.7.0 is a bit dated and makes downstream packaging (for example for Fedora) more complicated. I did not see any huge incompatible changes locally (but it might need some more love for windows -- hope the CI has a coverage).


wiktor-k (approved):
LGTM 👍 thanks!

hug-dev (approved):

hug-dev:
Kryoptic is not happy :crying_cat_face:

wiktor-k:
> Kryoptic is not happy 😿

I think the test is overly tied to specific versions. Converting them to check if the version is *greater than* some number is better. The minor check can be dropped IMO.

As a compromise I think the version could be just `println`d so that we have something in logs (AFAICT `Version` impls `Display`) :)

Jakuje:
> Kryoptic is not happy 😿

As mentioned in the previous comment, the kryoptic failure is fixed in #258, which is now merged so merging this should be green (but I can rebase it here if you prefer).

Jakuje:
> I think the test is overly tied to specific versions. Converting them to check if the version is _greater than_ some number is better. The minor check can be dropped IMO.

The version numbers should not change that fast with the pace Oasis works on them. So I would be for keeping them as they are now.

Actually, we have now the PKCS#11 3.2 in kryoptic now, while rust-cryptoki is still on 3.1 but with the backward compatibility, it should not matter functionally (just the this test complains).

wiktor-k:
> (but I can rebase it here if you prefer).

Please do. I work in security so I know "red means bad, green means good" :melting_face: :wink: 

> Actually, we have now the PKCS#11 3.2 in kryoptic now, while rust-cryptoki is still on 3.1 but with the backward compatibility, it should not matter functionally (just the this test complains).

Ah, right, this is for PKCS#11 version :man_facepalming: still, changing the test when the backend returns higher numbers seems kind of off :thinking:

Jakuje:
> > (but I can rebase it here if you prefer).
> 
> Please do. I work in security so I know "red means bad, green means good" 🫠 😉

done

> > Actually, we have now the PKCS#11 3.2 in kryoptic now, while rust-cryptoki is still on 3.1 but with the backward compatibility, it should not matter functionally (just the this test complains).
> 
> Ah, right, this is for PKCS#11 version 🤦‍♂️ still, changing the test when the backend returns higher numbers seems kind of off 🤔

Actually, its a pkcs11 interface version. There is only 3.0 and 3.2 now.

I took this test failing as a good feedback that the version is properly propagated to the calling application (as for now it is the only indication as we do not have anything else to test) and that I do not mess up in the future for example by not enabling the (currently non-default) pkcs11_3_2 feature of kryoptic in Fedora build in the future.

Doing the comparison on greater_equal or something will match a lot of bad numbers, which is also something I would like to avoid. Just printing would make it really `/dev/null` as I do not think anyone reads the logs if they pass (they are hidden by the tooling).

So the only reasonable option I can find would be just to allow 3.0 and 3.2 values right now. This will also allow the tests to work for other people who are not on 3.2 yet. I will do it in separate PR as this is a bit off-topic here.

Jakuje:
Filled #262.

wiktor-k (approved):

PR #262, opened 2025-04-22
tests: Be less strict in accepted interface versions

Jakuje:
Resolves the discussion that started in the following PR:

https://github.com/parallaxsecond/rust-cryptoki/pull/261#issuecomment-2820570013

### DIFF
--- a/cryptoki/tests/basic.rs
+++ b/cryptoki/tests/basic.rs
@@ -1037,7 +1037,11 @@ fn get_info_test() -> TestResult {
         assert_eq!(info.cryptoki_version().minor(), 40);
     } else {
         assert_eq!(info.cryptoki_version().major(), 3);
-        assert_eq!(info.cryptoki_version().minor(), 2);
+        let minor = info.cryptoki_version().minor();
+        assert!(
+            [0, 2].contains(&minor),
+            "Only 3.0 and 3.2 versions are expected but got 3.{minor}"
+        );
     }
     Ok(())
 }


cryptoki/tests/basic.rs
```diff
@@ -1037,7 +1037,7 @@ fn get_info_test() -> TestResult {
         assert_eq!(info.cryptoki_version().minor(), 40);
     } else {
         assert_eq!(info.cryptoki_version().major(), 3);
-        assert_eq!(info.cryptoki_version().minor(), 2);
+        assert!([0, 2].contains(&info.cryptoki_version().minor()));
```
wiktor-k, line 1040:
```suggestion
        let minor = &info.cryptoki_version().minor();
        assert!([0, 2].contains(&minor), "Only 0 and 2 are accepted but got {minor}");
```

This would make it faster to debug :)

wiktor-k (dismissed):
LGTM :+1: thanks!

cryptoki/tests/basic.rs
Jakuje: Good point. Updated.

cryptoki/tests/basic.rs
```diff
@@ -1037,7 +1037,11 @@ fn get_info_test() -> TestResult {
         assert_eq!(info.cryptoki_version().minor(), 40);
     } else {
         assert_eq!(info.cryptoki_version().major(), 3);
-        assert_eq!(info.cryptoki_version().minor(), 2);
+        let minor = info.cryptoki_version().minor();
+        assert!(
+            [0, 2].contains(&minor),
+            "Only 3.0 and 3.2 versions are expected but got {minor}"
```
wiktor-k, line 1043:
```suggestion
            "Only 3.0 and 3.2 versions are expected but got 3.{minor}"
```

I think this will be more in line with the actual version numbers.

wiktor-k (approved):
LGTM 👍 thanks!

hug-dev (approved):

cryptoki/tests/basic.rs
Jakuje: Just noticed that we would need edition 2021 for this as currently if warns as:
```
warning: panic message contains an unused formatting placeholder
    --> cryptoki/tests/basic.rs:1043:63
     |
1043 |             "Only 3.0 and 3.2 versions are expected but got 3.{minor}"
     |                                                               ^^^^^^^
     |
     = note: this message is not used as a format string when given without arguments, but will be in Rust 2021
     = note: `#[warn(non_fmt_panics)]` on by default
help: add the missing argument
     |
1043 |             "Only 3.0 and 3.2 versions are expected but got 3.{minor}", ...
     |                                                                       +++++
help: or add a "{}" format string to use the message literally
     |
1043 |             "{}", "Only 3.0 and 3.2 versions are expected but got 3.{minor}"
     |             +++++
```
Anything against changing the edition to 2021? I think the bindgen is not yet ready for 2024 edition though.

Issue #260, opened 2025-04-17
Thread Safety


daxpedda:
AFAIK you can't initialize Cryptoki while one instance already is and you can't create sessions to the same slot. However, currently doing so doesn't always return `CryptokiAlreadyInitialized` or `SessionExists`, but instead sometimes causes various UBs.

Here is a simple reproducible integration test I created:
```rust
mod common;

#[test]
fn test1() {
    common::init_pins();
}

#[test]
fn test2() {
    common::init_pins();
}
```

This almost never succeeds in my environment.

hug-dev:
Do you think it could be an error of the underlying PKCS11 implementation not returning those errors?

daxpedda:
I didn't look into it, but I was just using SoftHSM. The integration test I've outlined above is using the same setup as in the CI in this repo.

I will make a PR to see if we can reproduce it in the CI here.

bcheidemann:
Hey, FWIW I am also seeing thread unsafe behaviour when using cryptoki and SoftHSM (I haven't tried other libraries).

Sometimes I get:

```
signal: 11, SIGSEGV: invalid memory reference
```

And other times I get:

```
Function::Initialize: PKCS11 error: This value can only be returned by C_Initialize.  It means that the Cryptoki library has already been initialized (by a previous call to C_Initialize which did not have a matching C_Finalize call).
```

Is it intended that I should not create multiple instances of the library across different threads?

Jakuje:
Did you have a look at my comment in https://github.com/parallaxsecond/rust-cryptoki/pull/263#issuecomment-2830158320? Did it answer your questions? Or is there something else here to follow-up?

tldr is that the softhsm is broken. Using Initialize from different threads is probably ok, but you might need to work on some synchronization. If one of your threads will call Finalize, SoftHSM will crash.

bcheidemann:
> Did you have a look at my comment in [#263 (comment)](https://github.com/parallaxsecond/rust-cryptoki/pull/263#issuecomment-2830158320)? Did it answer your questions? Or is there something else here to follow-up?
> 
> tldr is that the softhsm is broken. Using Initialize from different threads is probably ok, but you might need to work on some synchronization. If one of your threads will call Finalize, SoftHSM will crash.

Hey, thanks for the link - somehow I didn't see the PR or comment 🤦‍♂️

[...]

PR #263, opened 2025-04-22
Test parallel access

daxpedda:
This PR just tests #260 in the CI.
It might be turned into a useful test later down the line when the bug can be reproduced and fixed.

### DIFF
--- a/cryptoki/tests/parallel.rs
+++ b/cryptoki/tests/parallel.rs
@@ -1,0 +1,13 @@
+// Copyright 2021 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+mod common;
+
+#[test]
+fn test1() {
+    common::init_pins();
+}
+
+#[test]
+fn test2() {
+    common::init_pins();
+}


daxpedda:
It is indeed [reproducible in the CI](https://github.com/parallaxsecond/rust-cryptoki/actions/runs/14592032848/job/40929291570?pr=263#step:4:474):
```
      Running tests/parallel.rs (target/debug/deps/parallel-9574901f18cd4e9f)

running 2 tests
error: test failed, to rerun pass `-p cryptoki --test parallel`

Caused by:
  process didn't exit successfully: `/home/runner/work/rust-cryptoki/rust-cryptoki/target/debug/deps/parallel-9574901f18cd4e9f` (signal: 11, SIGSEGV: invalid memory reference)
```

hug-dev:
Thanks for putting it together!

Does Rust run `test1` and `test2` as two different binaries? Or different threads? If that is the earlier, this could work as written in the documentation:

> If several applications are using Cryptoki, each one should call C_Initialize.  Every call to C_Initialize should (eventually) be succeeded by a single call to C_Finalize.  See [PKCS11-UG] for further details.

But if that is the later then I agree that the `initialize` call of one of them should have returned an error :thinking: Could we double check that this is working correctly by having a test trying to initialize twice in a row?

daxpedda:
> Does Rust run `test1` and `test2` as two different binaries? Or different threads?

Each test is run in its own thread as part of the same process. See https://doc.rust-lang.org/1.86.0/cargo/commands/cargo-test.html#description.

> Could we double check that this is working correctly by having a test trying to initialize twice in a row?

```rust
#[test]
fn test() {
    let (pkcs11, slot) = common::init_pins();
    let _session = pkcs11.open_rw_session(slot).unwrap();

    let (pkcs11, slot) = common::init_pins();
    let _session = pkcs11.open_rw_session(slot).unwrap();
}
```
This correctly always returns `CryptokiAlreadyInitialized`. Considering we can only recreate UB in a threaded scenario, I assume its some kind of race condition.

Jakuje:
I think this can be an issue on three different levels:

 * this library
 * the pkcs11 module you are using
 * the PKCS#11 specification

I did not manage to dig into the PKCS#11 specs deep enough, but there is a locking mechanism when you need to access a single token from different threads

The softhsm tests with your reproducer crash (I think interface functions do not have enough checks for threaded-access and finalize frees something they expect it will be around):

```
process didn't exit successfully: `/home/runner/work/rust-cryptoki/rust-cryptoki/target/debug/deps/parallel-9574901f18cd4e9f` (signal: 11, SIGSEGV: invalid memory reference)
```

but kryoptic just fails:

```
thread 'test1' panicked at cryptoki/tests/common.rs:41:52:
called `Result::unwrap()` on an `Err` value: Pkcs11(CryptokiNotInitialized, OpenSession)
```

This means that the tests got further, but the first test likely called c_finalize between the second one called the `init_token()` and `open_session()`. If you want to run run the tests in different threads, the initialization and finalization needs to be adjusted. Likely something like we do in kryoptic now:

 * the TestToken makes sure the cryptoki is initialized (just once)
 * for each test separate slot is created
 * the finalize is called only when the last test ends

https://github.com/latchset/kryoptic/blob/main/src/tests/keys.rs#L11

This way, we are able to run the tests in parallel. But we have a bit more control of the token as we call some internals to create separate slot for each test.

With generic pkcs11 tokens, you will likely have issues if one thread will be creating objects and the other will be expecting different ones in the single slot, but I think it should still be possible to adjust the tests to work.

hug-dev:
Closing since the related issue has been closed. Thanks!

PR #264, opened 2025-04-25
Add PKCS#11 3.2 bindings + plug them into the cryptoki

Jakuje:
The PKCS#11 3.2 headers are available (pre-release though, but final version should be out in coming weeks):

https://github.com/latchset/pkcs11-headers/tree/main/public-domain/3.2-prerelease

This pulls the new headers, regenerates bindings and adjusts the initialization so they can be used.

So far opened only as a draft as the specs is not published yet, but planning to add some more functions using kryoptic, that will verify the whole process works.

I took also the chance to update bindgen to 0.71.1 once we generate new binding version


wiktor-k:
Hmm fortunately this is just a clippy issue. I guess suppressing it somewhere would be an okayish workaround. (maybe in lib.rs of -sys?)

I'd still report it to bindgen. This may improve their code generation or at least make them consider adding the lint suppression in the generated code.

wiktor-k:
> I am not sure if there is some way how to improve the code generation to avoid this.

FWIW I didn't mean that. Slapping the lint suppression there is sufficient enough for me.

Jakuje:
Kryoptic tests now fail as there are few bugs in the last release that were already fixed in main. I think this is already good for first reviews, but I would probably wait for the pkcs11 3.2 to get released officially and I will likely implement also the SLH-DSA (they look quite simple and similar to the ML-DSA, but I do not have an implementation to test against yet).

hug-dev:
Feel free to ping us when you want us to review this again!

Jakuje:
I think this should be ready for reviews! Rebased and updated links to the final headers.

Jakuje:
> I found a small issue. When I ask for an `AttributeType::MlDsaParameterSet`, the token returns an `Attribute::ParameterSet`.
> 
> I think the patch introduces `AttributeType::MlDsaParameterSet` as an alias for `AttributeType::ParameterSet`. Both `MlDsaParameterSet` and `ParameterSet` map to `CKA_PARAMETER_SET`, and therefore the distinction is lost, and the response will be a `ParameterSet`.

Yeah, I was aware of this issue, but I am not sure how this could be easily handled given that this attribute has different values for different key types. Any thoughts?

I think it would require some logic in the GetAttributeValue mapping the type to the correct aliased "subtype" based on the key type, but I really did not look into that yet.

teythoon:
I think the aliasing is most surprising.  The Rust model is a refinement of the interface model, i.e. it is more expressive.  I think that is a mistake at this point.  It is also surprising for people coming from the PKCS#11 spec.

Instead, I think we should offer explicit conversion functions from `ParameterSetType` (doesn't exist yet, and should hold a u64, and that is what `Attribute::ParameterSet` should use) to `MlDsaParameterSet`.

cryptoki/src/context/general_purpose.rs
```diff
@@ -245,5 +258,18 @@ pub(super) fn is_fn_supported(ctx: &Pkcs11, function: Function) -> bool {
         Function::VerifyMessageBegin => check_fn!(ctx, VerifyMessageBegin),
         Function::VerifyMessageNext => check_fn!(ctx, VerifyMessageNext),
         Function::MessageVerifyFinal => check_fn!(ctx, MessageVerifyFinal),
+        /* PKCS #11 3.0 */
+        Function::EncapsulateKey => check_fn!(ctx, EncapsulateKey),
```
wiktor-k, line 262: Wait, this says `PKCS #11 3.0` but the import above `PKCS #11 3.2`? :thinking:

cryptoki/src/mechanism/mldsa.rs
```diff
@@ -0,0 +1,152 @@
...
+
+use crate::mechanism::{Mechanism, MechanismType};
+
+use cryptoki_sys::*;
+use std::{convert::TryInto, marker::PhantomData, ptr::null_mut};
+
+/// The hedge type for ML-DSA signature
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum HedgeType {
+    /// CKH_HEDGE_PREFERRED: Default
```
wiktor-k: Hmm... I wonder if that should also be marked with `#[default]` and the enum deriving `Default`? :thinking: 

Also I'd wrap `CKH_HEDGE_PREFERRED` in backticks... it looks weird in rendered docs without them :sweat_smile:

cryptoki/src/mechanism/mldsa.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// token must produce a hedged signature or fail
+    Required,
+    /// CKH_DETERMINISTIC_REQUIRED
+    /// token must produce a deterministic signature or fail
+    DeterministicRequired,
+}
+
+impl HedgeType {
+    /// Convert the enum to Cryptoki value
+    pub fn into_ck(&self) -> CK_HEDGE_TYPE {
```
wiktor-k: I think we could do the `From` thingy here (since `HedgeType` is defined in this crate).

cryptoki/src/mechanism/mldsa.rs
```diff
@@ -0,0 +1,152 @@
...
+    ///
+    /// A new SignAdditionalContext struct.
+    pub fn new(hedge: HedgeType, context: Option<&[u8]>, hash: MechanismType) -> Self {
+        let (p_context, ul_context_len) = match context {
+            Some(c) => (
+                c.as_ptr() as *mut _,
+                c.len().try_into().expect("usize can not fit in CK_ULONG"),
+            ),
+            None => (null_mut(), 0),
+        };
```
wiktor-k, line 126:
```suggestion
        let (p_context, ul_context_len) = if let Some(c) = context {
                (c.as_ptr() as *mut _,
                c.len().try_into().expect("usize can not fit in CK_ULONG")),
            } else {
                (null_mut(), 0),
            };
        };
```

I tend to avoid `match`es which then destructure into Some/None as I think `if let` is more readable... but it's not a blocker :)

cryptoki/tests/common/mod.rs
```diff
@@ -16,6 +16,7 @@ fn get_pkcs11_path() -> String {
         .unwrap_or_else(|_| "/usr/local/lib/softhsm/libsofthsm2.so".to_string())
 }
 
+#[allow(dead_code)]
```
wiktor-k, line 19: Is that still needed?

Jakuje:
```
warning: function `is_softhsm` is never used
  --> cryptoki/tests/common/mod.rs:19:8
   |
19 | pub fn is_softhsm() -> bool {
   |        ^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `cryptoki` (test "ml_kem") generated 1 warning
warning: `cryptoki` (test "ml_dsa") generated 1 warning (1 duplicate)
```
I did some restructure of the tests and this reports failure for the two now (mlkem and mldsa) as they do not use this public function.

I think the whole test structure can be improved as now it builds separate test binaries, which is not very effective.

So as a after-thought I think we should really have one test binary and source code split to separate files, probably something similar as we have in kryoptic (but that lives under `src/tests`, which I understand is not a standard place to push tests according to rust.

wiktor-k: > which is not very effective.

Oh yeah, this is definitely an annoying aspect of `cargo test` and part of the reason why in other projects we're instead using `cargo nextest run` instead of bare `cargo test`.

cryptoki/src/context/general_purpose.rs
Jakuje: Fixed! Thanks for your sharp eyes!

cryptoki/src/mechanism/mldsa.rs
Jakuje: Changed, fixed

cryptoki/src/mechanism/mldsa.rs
Jakuje: Done. Thanks for nudge!

cryptoki/src/context/mod.rs
```diff
@@ -43,12 +43,11 @@ use std::sync::RwLock;
 #[derive(Debug)]
 enum FunctionList {
```
hug-dev, line 44: Maybe I am the one suggesting to create such a structure but I am now wondering if it's useful since all the variants contain the same type :sweat_smile:

hug-dev (approved):
Thank you for pushing for this!

Jakuje:
@teythoon I did rehash the `ParameterSetType` a bit in the last commit. Is it something you had in your mind or is there something missing that you would add or change?

Jakuje:
> But, I remember Simo suggesting to keep the aliases, always make sure to request the key type as well, and when wrapping the response in Rust types, convert the generic parameter set response to a specific one based on the reported key type. If that is feasible to implement, that'd be most convenient for the downstream users.

I guess I will have to think about that a bit more (and will probably not get back to that before I will get back from holiday). It will either have to be some ad-hoc change in the C_GetAttribute to do both of the pulling of the key type and converting the parameter set to the key-specific one. Or some new more generic logic to do this. But as far as I know the ParameterSet of the three new PQC algorithms is the only one so the ad-hoc solution might be enough.

Jakuje:
@teythoon I though about implementing the aliases, but though it would be more confusing now. With the `.into()` conversion, the inner types are just one step away. If we decide to do this later, it can be done though.

@hug-dev @wiktor-k I think this is ready for final reviews.

cryptoki/src/mechanism/mechanism_info.rs
```diff
@@ -252,6 +254,18 @@ impl MechanismInfo {
     pub fn multi_message(&self) -> bool {
         self.flags.contains(MechanismInfoFlags::MULTI_MESSAGE)
     }
+
+    /// True if the mechanism can be used to encapsulate other keys
+    ///
```
wiktor-k, line 259: I don't think we need this extra line.

cryptoki/src/mechanism/mechanism_info.rs
```diff
@@ -252,6 +254,18 @@ impl MechanismInfo {
...
+    pub fn encapsulate(&self) -> bool {
+        self.flags.contains(MechanismInfoFlags::ENCAPSULATE)
+    }
+
+    /// True if the mechanism can be used to decapsulate other keys
+    ///
```
wiktor-k, line 266: Likewise.

cryptoki/src/mechanism/mldsa.rs
```diff
@@ -0,0 +1,152 @@
...
+use crate::mechanism::{Mechanism, MechanismType};
+
+use cryptoki_sys::*;
+use std::{convert::TryInto, marker::PhantomData, ptr::null_mut};
+
+/// The hedge type for ML-DSA signature
+#[derive(Debug, Clone, Copy, PartialEq, Default)]
+pub enum HedgeType {
+    /// `CKH_HEDGE_PREFERRED`: Default
+    /// token may create either a hedged signature or a deterministic signature
```
wiktor-k:
```suggestion
    /// Token may create either a hedged signature or a deterministic signature
```

cryptoki/src/mechanism/mldsa.rs
```diff
@@ -0,0 +1,152 @@
...
+    #[default]
+    Preferred,
+    /// `CKH_HEDGE_REQUIRED`
+    /// token must produce a hedged signature or fail
```
wiktor-k:
```suggestion
    /// Token must produce a hedged signature or fail
```

cryptoki/src/mechanism/mldsa.rs
```diff
@@ -0,0 +1,152 @@
...
+    Required,
+    /// `CKH_DETERMINISTIC_REQUIRED`
+    /// token must produce a deterministic signature or fail
+    DeterministicRequired,
```
wiktor-k: Alternatively:

```suggestion
    /// Token must produce a deterministic signature or fail.
    ///
    /// This variant maps to `CKH_DETERMINISTIC_REQUIRED`.
    DeterministicRequired,
```

cryptoki/src/mechanism/mldsa.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// Marker type to ensure we don't outlive the data
+    _marker: PhantomData<&'a [u8]>,
+}
+
+impl SignAdditionalContext<'_> {
+    /// Construct ML-DSA signature parameters.
+    ///
+    /// # Arguments
+    ///
+    /// * `hedge` - The HedgeType.
```
wiktor-k:
```suggestion
    /// * `hedge` - The [`HedgeType`].
```

cryptoki/src/mechanism/mldsa.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// * `context` - The context.
+    ///
+    /// # Returns
+    ///
+    /// A new SignAdditionalContext struct.
```
wiktor-k:
```suggestion
    /// A new [`SignAdditionalContext`] struct.
```

cryptoki/src/session/encapsulation.rs
```diff
@@ -0,0 +1,90 @@
...
+                template.as_mut_ptr(),
+                template.len().try_into()?,
+                encapsulated.as_mut_ptr(),
+                &mut encapsulated_len,
+                &mut handle,
+            ))
+            .into_result(Function::EncapsulateKey)?;
+        }
+
+        encapsulated.resize(encapsulated_len.try_into()?, 0);
```
wiktor-k: Do we need `resize` here or would [`truncate`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.truncate) suffice? :thinking: (I'm wondering if it's possible that the first call would return smaller number than the second)

wiktor-k (dismissed):
LGTM :+1: a couple of nits/questions.

cryptoki/src/mechanism/mechanism_info.rs
Jakuje: I think I wanted to add the link to the `Session::decapsulate` as in the other operations. Will add.

cryptoki/src/mechanism/mldsa.rs
Jakuje: This sounds more readable. Will go towards this direction.

cryptoki/src/session/encapsulation.rs
Jakuje: We discussed the same in https://github.com/parallaxsecond/rust-cryptoki/pull/255#discussion_r2041565129 and came to the conclusion that it should really be the truncate only, but we never followed-up. I will sneak one more commit in fixing this in all the places that follow this semantics.

wiktor-k: Thanks for digging this up, I had a feeling we had this conversation :sweat_smile:

Jakuje:
Applied all the suggestions. Thanks for prompt review @wiktor-k !

wiktor-k (approved):

PR #270, opened 2025-05-06
Implement PKCS#11 3.0 MessageSign API

Jakuje:
This is separated from #264 as it really does not depends on the 3.2 API.

Currently, this API is not implemented by any of the software tokens we have this tracked for kryoptic in latchset/kryoptic#74 but given that this really does not add new functionality, it is still quite a low priority.

For that reason, I do not have any test coverage there (but our understanding is that we should be able to use this API with existing mechanisms, just without passing any parameters to the `sign_message_begin()`, which would make it more handy to have the possibility to pass the `param` optional (or just define `MessageParam::None`. Thoughts?

### DIFF
--- a/cryptoki/src/session/message_decryption.rs
+++ b/cryptoki/src/session/message_decryption.rs
@@ -1,6 +1,6 @@
 // Copyright 2025 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
-//! Encrypting data
+//! Decrypting messages
 
 use crate::context::Function;
 use crate::error::{Result, Rv};
--- a/cryptoki/src/session/message_signing.rs
+++ b/cryptoki/src/session/message_signing.rs
@@ -1,0 +1,246 @@
+// Copyright 2025 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+//! Message Signing/Verification and authentication functions
+
+use crate::context::Function;
+use crate::error::{Result, Rv};
+use crate::mechanism::{Mechanism, MessageParam};
+use crate::object::ObjectHandle;
+use crate::session::Session;
+use cryptoki_sys::*;
+use std::convert::TryInto;
+
+impl Session {
+    /// Prepare a session for one or more Message-based signature using the same mechanism and key
+    pub fn message_sign_init(&self, mechanism: &Mechanism, key: ObjectHandle) -> Result<()> {
+        let mut mechanism: CK_MECHANISM = mechanism.into();
+
+        unsafe {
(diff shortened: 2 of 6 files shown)


cryptoki/src/session/message_decryption.rs
```diff
@@ -1,6 +1,6 @@
 // Copyright 2025 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
-//! Encrypting data
+//! Decrypting messages
```
hug-dev, line 3: :detective:

cryptoki/src/session/message_signing.rs
```diff
@@ -0,0 +1,246 @@
...
+
+use crate::context::Function;
+use crate::error::{Result, Rv};
+use crate::mechanism::{Mechanism, MessageParam};
+use crate::object::ObjectHandle;
+use crate::session::Session;
+use cryptoki_sys::*;
+use std::convert::TryInto;
+
+impl Session {
```
hug-dev, line 13: Could we re-organize the methods here to split single-part and multi-part?

hug-dev (commented):
> For that reason, I do not have any test coverage there

Even SoftHSM does not implement those? I was thinking that maybe a simple single-part test would be enough.
Is it also possible to write a test but not run it on CI or mark it as failable?

Jakuje:
> > For that reason, I do not have any test coverage there
> 
> Even SoftHSM does not implement those?

No. This is PKCS#11 3.0 API, which is not present in SoftHSM.

> I was thinking that maybe a simple single-part test would be enough. Is it also possible to write a test but not run it on CI or mark it as failable?

I hope we will get some implementation in kryoptic in foreseeable future, but we got stuck with the https://github.com/oasis-tcs/pkcs11/issues/16 as it is not clear to neither of us what mechanisms should this API be usable with and in what semantics.

I will probably change it back to draft for now. To be able to write some tests, we will need some new parameters and some mechanisms to use with this API. Adding unusable API makes no sense.

cryptoki/src/session/message_signing.rs
Jakuje: I was starting from the sign/verify API, which was in a single file. The semantics is also interleaved as we need to initialize the operation first and then call either one-shot or multipart operations, which really does not make a good split as for both of the ways, we need the operation initialization.

cryptoki/src/session/message_decryption.rs
Jakuje: Sorry, did not write it right the first time :)

Issue #268, opened 2025-05-05
CKA_UNIQUE_ID not exposed


nwalfield:
According to version 3.2 draft 13 of the standard, `CKA_UNIQUE_ID` is a way to unambiguously identify objects across sessions:

> 4.4.1 The CKA_UNIQUE_ID attribute
>
> Any time a new object is created, a value for CKA_UNIQUE_ID MUST be generated by the token and stored with the object. The specific algorithm used to generate unique ID values for objects is token-specific, but values generated MUST be unique across all objects visible to any particular session, and SHOULD be unique across all objects created by the token. Reinitializing the token, such as by calling C_InitToken, MAY cause reuse of CKA_UNIQUE_ID values.
>
> Any attempt to modify the CKA_UNIQUE_ID attribute of an existing object or to specify the value of the CKA_UNIQUE_ID attribute in the template for an operation that creates one or more objects MUST fail. Operations failing for this reason return the error code CKR_ATTRIBUTE_READ_ONLY.

As far as I can tell, (I'm looking at 19bb7b9ca827064b4fcc8a0b3f3f382110a47f47), this attribute is not exposed.  It is possible to see that it is present by passing `AttributeType::VendorDefined(0x00000004)` (the value according to the `-sys` crate) to `Session::get_attribute_info`, but if I try to fetch the value using `Session::get_attributes`, I get `Error getting value: Feature not supported`.

Is not exposing `CKA_UNIQUE_ID` an oversight?  Is it intentional?

Thanks for any insight.

Jakuje:
We discussed this over IRC. Adding the attribute is quite trivial (done in #271). This was not done before, likely because it is not supported by SoftHSM.

PR #271, opened 2025-05-06
 Add UniqueId Attribute

Jakuje:
Fixes: #268

### DIFF
--- a/cryptoki/src/object.rs
+++ b/cryptoki/src/object.rs
@@ -122,6 +122,8 @@ pub enum AttributeType {
     Token,
     /// Determines if the object is trusted
     Trusted,
+    /// Unique Object Id
+    UniqueId,
     /// Determines if a key supports unwrapping
     Unwrap,
     /// Gives the URL where the complete certificate can be obtained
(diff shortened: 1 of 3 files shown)


wiktor-k (approved):
Looks simple enough. I wonder if we should keep filing SoftHSM tickets for all things that are missing... who knows maybe someone will implement them :sweat_smile:

Jakuje:
> Looks simple enough. I wonder if we should keep filing SoftHSM tickets for all things that are missing... who knows maybe someone will implement them 😅

Thats what I was trying since 2019 (also with PRs): https://github.com/softhsm/SoftHSMv2/issues?q=is%3Aissue%20author%3AJakuje%20is%3Apr%20 feel free to continue, but their issue tracker looks like dead end for at least last 4 years.

wiktor-k:
Hmm... so the project is effectively abandoned? :thinking: Maybe it's time for SoftHSMv3 :wink: okay. that's good to know (and have in writing in case someone else asks :sweat_smile: ).

Thanks for your time! :wave:

Jakuje:
That is my understanding. There were some attempts to resurect it last year, but it looks like it did not go anywhere.

hug-dev (approved):

PR #272, opened 2025-05-09
#266: Added conditional attribute to pack the PkcsOaepParams struct o…

### LINKED ISSUE #266: PkcsOaepParams mis-aligned
I was looking at why RSA PKCS OAEP fails with data and while debugging on Windows I noticed that CK_RSA_PKCS_OAEP_PARAMS should be 24 bytes, while checking `std::mem::size_of::<PkcsOaepParams>()` it was returning 32.

The struct is defined as follows:

```
#[derive(Copy, Debug, Clone)]
#[repr(C)]
pub struct PkcsOaepParams<'a> {
    /// mechanism ID of the message digest algorithm used to calculate the digest of the encoding
    /// parameter
    hash_alg: MechanismType,
    /// mask generation function to use on the encoded block
    mgf: PkcsMgfType,
    /// source of the encoding parameter
    source: cryptoki_sys::CK_RSA_PKCS_OAEP_SOURCE_TYPE,
    /// data used as the input for the encoding parameter source
    source_data: *const c_void,
    /// length of the encoding parameter source input
    source_data_len: Ulong,
    /// marker type to ensure we don't outlive the source_data
    _marker: PhantomData<&'a [u8]>,
}
```

After a closer look it turns out that without `repr(C)` it's returning 24, meaning that it's being aligned to the next 8 bytes (green is expected while red is what appears to be happening):

Image

Adding packed (`#[repr(C, packed)]`) appears to remove the inter-field padding and the `rsa_pkcs_oaep_with_data` test passes.

I need to test on linux but would welcome any thoughts :).

ivozeba:
While debugging a separate issue I discovered the cryptoki-sys build script and noticed that you can conditionally pack structs on windows by simply adding `#[cfg_attr(windows, repr(packed))]`.

This resolves #266 and only affects the PkcsOaepParams struct (until we find another one).

### DIFF
--- a/cryptoki/src/mechanism/rsa.rs
+++ b/cryptoki/src/mechanism/rsa.rs
@@ -134,6 +134,7 @@ pub struct PkcsPssParams {
 /// Parameters of the RsaPkcsOaep mechanism
 #[derive(Copy, Debug, Clone)]
 #[repr(C)]
+#[cfg_attr(windows, repr(packed))]
 pub struct PkcsOaepParams<'a> {
     /// mechanism ID of the message digest algorithm used to calculate the digest of the encoding
     /// parameter


hug-dev (approved):
Seems to align well with your related issue. Thanks a lot for the detailed description there!

wiktor-k (approved):
It seems to be improving things so... let's go with it :rocket:

Issue #267, opened 2025-05-05
Consider changing get_attribute_info_map from taking a Vec to taking a slice


nwalfield:
[`Session::get_attribute_info_map`](https://docs.rs/cryptoki/latest/cryptoki/session/struct.Session.html#method.get_attribute_info_map) takes a `Vec<AttributeType>`.  I looked at [the implementation](https://github.com/parallaxsecond/rust-cryptoki/blob/614c946/cryptoki/src/session/object_management.rs#L486) and I think passing a slice would work just as well and could allow the caller to avoid allocating a `Vec`.

wiktor-k:
Sounds sensible. PRs welcome, I guess 😅

PR #273, opened 2025-05-09
#267 Modify get_attribute_info_map to take slice instead of vec

ivozeba:
Modified the argument from vec to slice as highlighted in #267 and updated the only test case that used get_attribute_info_map.


hug-dev (dismissed):
Thanks for picking this up :star: !

wiktor-k:
Yep, it looks good, but please `cargo fmt` and add `Fixes: https://github.com/parallaxsecond/rust-cryptoki/issues/267` to the commit message.

hug-dev (approved):

wiktor-k (approved):

Issue #282, opened 2025-06-09
Building on ix86 fails


Jakuje:
Some changes to the tests fail to build on 32bit architecture now (the code released in 0.10).

To prevent this in future, we should consider adding a CI job testing on 32b architecture.

Found this only while trying to update the package in Fedora.

```
error[E0277]: the trait bound `Ulong: From<u64>` is not satisfied
   --> tests/basic.rs:458:56
    |
458 |         Attribute::ValueLen((AES128_BLOCK_SIZE as u64).into()),
    |                                                        ^^^^ the trait `From<u64>` is not implemented for `Ulong`
    |
    = help: the trait `From<u64>` is not implemented for `Ulong`
            but trait `From<u32>` is implemented for it
    = help: for that trait implementation, expected `u32`, found `u64`
    = note: required for `u64` to implement `Into<Ulong>`

error[E0277]: the trait bound `Ulong: From<u64>` is not satisfied
   --> tests/basic.rs:569:56
    |
569 |         Attribute::ValueLen((AES128_BLOCK_SIZE as u64).into()),
    |                                                        ^^^^ the trait `From<u64>` is not implemented for `Ulong`
    |
    = help: the trait `From<u64>` is not implemented for `Ulong`
            but trait `From<u32>` is implemented for it
    = help: for that trait implementation, expected `u32`, found `u64`
    = note: required for `u64` to implement `Into<Ulong>`

error[E0277]: the trait bound `Ulong: From<u64>` is not satisfied
    --> tests/basic.rs:1681:56
     |
1681 |         Attribute::ValueLen((AES128_BLOCK_SIZE as u64).into()),
     |                                                        ^^^^ the trait `From<u64>` is not implemented for `Ulong`
     |
     = help: the trait `From<u64>` is not implemented for `Ulong`
             but trait `From<u32>` is implemented for it
     = help: for that trait implementation, expected `u32`, found `u64`
     = note: required for `u64` to implement `Into<Ulong>`

error[E0277]: the trait bound `Ulong: From<u64>` is not satisfied
    --> tests/basic.rs:2182:56
     |
2182 |         Attribute::ValueLen((AES256_BLOCK_SIZE as u64).into()),
     |                                                        ^^^^ the trait `From<u64>` is not implemented for `Ulong`
     |
     = help: the trait `From<u64>` is not implemented for `Ulong`
             but trait `From<u32>` is implemented for it
     = help: for that trait implementation, expected `u32`, found `u64`
[...]

PR #283, opened 2025-06-09
ci: Run tests also on i686

Jakuje:
Fixes: #285

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,7 @@ jobs:
       matrix:
         target:
           - x86_64-unknown-linux-gnu
+          - i686-unknown-linux-gnu
         toolchain:
           - stable
           - "1.77" # MSRV
@@ -107,8 +108,14 @@ jobs:
           default: true
       - name: Install SoftHSM
         run: |
-          sudo apt-get update -y -qq &&
-          sudo apt-get install -y -qq libsofthsm2 &&
+          if [ "${{ matrix.target }}" = "i686-unknown-linux-gnu" ]; then
+            sudo dpkg --add-architecture i386
+            sudo apt-get update -y -qq
+            sudo apt-get install -y -qq gcc-multilib:i386 libsofthsm2:i386 gcc:i386
+          else
+            sudo apt-get update -y -qq
+            sudo apt-get install -y -qq libsofthsm2
+          fi
           mkdir /tmp/tokens
           echo "directories.tokendir = /tmp/tokens" > /tmp/softhsm2.conf
       - name: Install Rust target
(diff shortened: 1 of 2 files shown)


Jakuje:
> #282 has already been fixed... Should this be closed or is there anything valuable that we can scavenge? 😁

This is about running the tests on i386, not just building as with the ulong size on 32b arch, some things might explode even during runtime. I tried to rework it, but did not manage to finalize the linking on multiarch yet.

Jakuje:
And it was a missing dollar sign all the time :facepalm: 

image

So this should be good to review.

If not done early, I will start adding arm builds :)

wiktor-k:
Okay, I think this looks good and should be merged :rocket: 

One question though, couldn't we adjust the test to use [`usize::MAX`](https://doc.rust-lang.org/std/primitive.usize.html#associatedconstant.MAX) and have one thing that works regardless of the target pointer width?

I'll approve the thing anyway :sweat_smile:

hug-dev (approved):

Jakuje:
> One question though, couldn't we adjust the test to use [`usize::MAX`](https://doc.rust-lang.org/std/primitive.usize.html#associatedconstant.MAX) and have one thing that works regardless of the target pointer width?

Good idea! Its always better to fix the test than skip it. I will adjust it. I was probably too focused on fixing the CI, rather than fixing the tests.

Jakuje:
> > One question though, couldn't we adjust the test to use [`usize::MAX`](https://doc.rust-lang.org/std/primitive.usize.html#associatedconstant.MAX) and have one thing that works regardless of the target pointer width?
> 
> Good idea! Its always better to fix the test than skip it. I will adjust it. I was probably too focused on fixing the CI, rather than fixing the tests.

Actually the value `4294967295` is `u32::MAX-1`.

But actually this already fails non-gracefully in the test itself if changed to this value, as it looks like the max vector capacity is different on the 32b arch rust:
```
---- gcm_param_graceful_failure stdout ----

thread 'gcm_param_graceful_failure' panicked at cryptoki/tests/basic.rs:1865:18:
capacity overflow
```
So yes, I can probably try to find the vector capacity, but this would be already different test. This test tries this input specially made for the specs (commented in the code)

https://github.com/parallaxsecond/rust-cryptoki/blob/83790f29cabed346089faffa91a5f59abe16ccde/cryptoki/src/mechanism/aead.rs#L47-L49

So reverting back my first comment. Its probably best to skip it on 32b arch in the end.

Good to get merged.

wiktor-k:
Oh yes, definitely. Sorry to send you on such a rabbit chase I thought it will be "just a simple fix" :wink: 

Merging, thanks for your work! :bow:

Issue #285, opened 2025-06-09
Run CI on more different architectures in some more systematic manner


Jakuje:
Currently, the CI runs only on amd64. The #283 added i386 builds (on amd64 cpus). Github provides also arm64 runners which might be interested.

https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

 _Originally posted by @Jakuje in [#282](https://github.com/parallaxsecond/rust-cryptoki/issues/282#issuecomment-2956572322)_

PR #286, opened 2025-06-11
Build and test in a matrix

wiktor-k:


Jakuje:
To catch the #282, we would need to also run the tests on this target.

wiktor-k:
> To catch the https://github.com/parallaxsecond/rust-cryptoki/issues/282, we would need to also run the tests on this target.

In this case, to be precise we don't actually need to *run* tests we just need to *compile* them. `cargo check ... --all-targets` checks tests and examples too (see [cargo targets](https://doc.rust-lang.org/cargo/reference/cargo-targets.html)) and IMO we were missing that.

I think we don't need to strictly `cargo build` since we throw away the artifacts so `check` is sufficient. Running the tests may also be a good idea (but separate from the check) although I'm not sure how does it work (I don't think it emulates, say, aarch64 to run the test).

I've updated the matrix job and it's clearly visible it's not just that particular 32bit target that was failing.

To be honest I'd move the `ci.sh` to yaml since it's pretty straightforward and GitHub shows separate sections for all steps but I don't know how much is @hug-dev attached to that :wink:

wiktor-k:
I've cherry-picked your commits from the other PR... let's see how many tests will be green now :)

ci.sh
```diff
@@ -19,22 +19,10 @@ if cargo clippy --version; then
...
-RUST_BACKTRACE=1 cargo build --target x86_64-unknown-freebsd
-
-RUST_BACKTRACE=1 cargo test
+cargo build
+
+cargo build --all-features
+
+cargo build --target "$TARGET"
+
+cargo test
```
Jakuje: I would probably change this to run only if the arch triplet matches the system we run on. Either simply by running this when TARGET is `x86_64-unknown-linux-gnu` . Or somehow dynamically if we will use different runners.

Otherwise we run the same tests 13 times in the same configuration here.

.github/workflows/ci.yml
```diff
@@ -32,14 +32,50 @@ jobs:
...
+          - aarch64-unknown-linux-gnu
+          - i686-unknown-linux-gnu
+          - loongarch64-unknown-linux-gnu
+          - powerpc64-unknown-linux-gnu
+          - powerpc64le-unknown-linux-gnu
+          - x86_64-pc-windows-msvc
+          - x86_64-apple-darwin
+          - aarch64-apple-darwin
+          - x86_64-unknown-freebsd
+          - riscv64gc-unknown-linux-gnu
```
Jakuje, line 60: we are missing the "default" target here now. Something like `x86_64-unknown-linux-gnu`, which was probably excluded from the previous list as it was running when there was no `--target` provided (and which now builds when we call `cargo test`.

Jakuje (commented):
Sounds like all of them were fixed :)

I like the idea of running on the whole matrix. It makes it easier to catch all possible oddness (but really the most common one is the ulong size which demonstrates on 32b arches). The linux-x86_64 is missing though.

On the other hand, I would probably be interested in running the test on 32b arch too (can be separate PR based on what I did in #283). We can probably do it easily only on i386 though. We might get the arm64 runner as mentioned in #285 and we might be able to run there the  arm32 binaries, but I never played around this and I an skeptical it will show other errors than the i386 ones.

ci.sh
wiktor-k: Ooh, nice catch! Thanks :) I split that job into a separate task and now we've got checks and unit tests in different jobs with different targets.

wiktor-k:
Okay, I've simplified it further: noticed that we're checking formatting twice (once in a job, second time in `ci.sh`) and removed the one in `ci.sh`. I've also extracted clippy into separate job. The same for unit tests which (for now?) use a very limited `target` list (which should mimic the old behavior).

See if this looks any better and thanks for review! :wave: 

Btw, we won't be able to merge this anyway until @hug-dev changes the definition of "Required" checks (they're hardcoded for now to look for old jobs).

Jakuje (dismissed):
looks good!

hug-dev (dismissed):
yooohooo amazing job! Looks much better than our `ci.sh` script! Definitely in favour of removing it if we can do everything from the yaml file :) Maintaining a big shell file is not fun...

hug-dev:
Modified the branch protection rules to remove the old check. I however need to manually add all matrix variations in the branch protection rules (that is 38 checks). There are some [way around](https://stackoverflow.com/questions/70275958/how-do-i-set-a-github-branch-protection-rule-based-on-the-success-or-failure-of) but they do not look too good.

Shall I manually add all of them as required or do we only require a subset?

hug-dev (approved):

wiktor-k:
Phew, great to hear that @hug-dev, I was worried how would you react :wink: but I think this should vastly improve our testing coverage :)

I've dropped the two commits from @Jakuje as they seem to be addressed by the Windows PR merged recently.

Thanks for help, deeply appreciated :pray:

PR #288, opened 2025-06-12
Fix typos in doc texts

Jakuje:
I found and fixed some typos in #270, but the whole PR turned out that it won't get merged any time soon so moved the non-controversial documentation fixes from there to separate PR, that should be easy to review and merge.


wiktor-k (approved):
Good idea. Thanks! :ok_hand:

paulhowardarm (approved):
LGTM - good catches, thanks.

Issue #279, opened 2025-05-22
Unmaintained Dependency `paste` (RUSTSEC-2024-0436)


Nicoretti:
Hi Cryptoki maintainers,

Thanks for providing this library!

I noticed that `cryptoki` uses the `paste` crate (https://github.com/dtolnay/paste), and I wanted to bring to your attention RUSTSEC-2024-0436, which notes that `paste` is currently unmaintained.

I've been thinking about possible solutions, and given how small the `paste` crate is, vendoring it directly into `cryptoki` seems like a reasonable and straightforward option. This would remove the external dependency and give you more control over its maintenance and security going forward.  Plus, it wouldn't negatively impact the current security posture.

**To that end, I was wondering if you'd be open to a PR where I vendor the `paste` code directly into `cryptoki`?** 

Thanks for your time and all your work on `cryptoki`!

best
Nico

wiktor-k:
Others [had the same problems](https://github.com/aws/aws-lc-rs/issues/722#issuecomment-2713120628) and it seems there are quite a few forks already, including `pastey`. I don't have any suggestions here, on one hand inlining is annoying, on another, if that's just a small piece of code... 🤷‍♂

Nicoretti:
We are talking 4 files, overall ~800-900 Loc (including doc comments).
Files could be merged into a single one though from my pov.

hug-dev:
Thanks for bringing this up! If there are already forks available I would also prefer using them directly instead of having to maintain the crate ourselves 😅 Do you see any advantages of having the crate in our repo?

Would definitely be open for a PR trying out `pastey` or another suitable replacement!

Nicoretti:
Thanks for the timely replies. I understand the ups and downs from a maintainer's perspective (In the end it is your decision). I was also taken aback that the vendoring is more significant than expected due to the large test suite. I have created a PR which can serve as a basis for discussion. If you decide against vendoring, feel free to close/reject it. If you use a maintained alternative, that would also be a good solution from a user's perspective 👍.

Nicoretti:
@hug-dev 
> Thanks for bringing this up! If there are already forks available I would also prefer using them directly instead of having to maintain the crate ourselves 😅 

I can fully understand that.

> Do you see any advantages of having the crate in our repo? 

Compared to having another well maintained dependency?  -> **No** 

If there are no alternatives available, it does not make the status quo worse as you likely don't touch it,
but if you need to fix or adjust something, you can.

[...]

PR #289, opened 2025-06-19
Fix Pkcs11::new bug preventing PKCS#11 library loading and bump rust-version and update deps

mematthias:
# Update crates: `cargo update` and fix Clippy warnings

This pull request was originally intended to cover multiple changes, but it has now been split into several smaller parts for clarity and easier review.

This PR is focused solely on crate updates and cleanup:

- Re-ran `cargo update` to ensure dependencies are up to date
- Fixed all warnings reported by `cargo clippy`

Further related changes will follow in separate pull requests.

Old description
```text
// # Summary
//
// This PR fixes a critical issue in `Pkcs11::new` that prevented my PKCS#11 library from being loaded correctly. The root cause was the use of `MaybeUninit::uninit()` in a context where zero-initialized or `null`-assigned variables were required. This has been resolved by using explicitly `null`-assigned variables instead.
//
// # Changes
//
// - Fixed bug in `Pkcs11::new` related to improper initialization, resolving library loading failure
// - Made the `basic.rs` test module compile again
// - Ran `cargo update` to bring dependencies up to date
// - Raised `rust-version` to `1.77` in `Cargo.toml`, as `mem::offset_of` is used and only available from that version onward
```

### DIFF
--- a/cryptoki/src/object.rs
+++ b/cryptoki/src/object.rs
@@ -1156,7 +1156,7 @@ impl TryFrom<CK_ML_KEM_PARAMETER_SET_TYPE> for MlKemParameterSetType {
             CKP_ML_KEM_768 => Ok(MlKemParameterSetType::ML_KEM_768),
             CKP_ML_KEM_1024 => Ok(MlKemParameterSetType::ML_KEM_1024),
             _ => {
-                error!("ML-KEM parameter set {} is not supported.", val);
+                error!("ML-KEM parameter set {val} is not supported.");
                 Err(Error::NotSupported)
             }
         }
@@ -1225,7 +1225,7 @@ impl TryFrom<CK_ML_DSA_PARAMETER_SET_TYPE> for MlDsaParameterSetType {
             CKP_ML_DSA_65 => Ok(MlDsaParameterSetType::ML_DSA_65),
             CKP_ML_DSA_87 => Ok(MlDsaParameterSetType::ML_DSA_87),
             _ => {
-                error!("ML-DSA parameter set {} is not supported.", val);
+                error!("ML-DSA parameter set {val} is not supported.");
                 Err(Error::NotSupported)
             }
         }


Jakuje:
Thank you for the fix for this issue!

It would be good if you could provide the changes at least in separate commits (or ideally as separate PRs) . Now there is everything mixed up in one commit (that is missing the sing-off) and then there is a fixup commit reverting one of the changes. We are certainly interested in the fix, but I am not sure if we want to pull all the changes of dependencies.

Do you know why the CI still works on the rust 1.66 when the `offset_of` is used? Or is it only on some configurations?

Jakuje:
Interesting, The CI runs only on Linux now. From what I see, it looks the the `offset_of` was there before 1.77, but was stabilized with 1.77. Is there some way to run (at least build of rust crate) in GH Actions CI on Windows? It might help to catch some differences.

cryptoki-sys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cryptoki-sys"
-version = "0.4.0"
+version = "0.4.1"
```
hug-dev: You don't need to change that one now, we usually bump that manually when we release a new version :)

cryptoki-sys/Cargo.toml
```diff
@@ -10,13 +10,13 @@ categories = ["api-bindings", "external-ffi-bindings", "cryptography", "hardware
 license = "Apache-2.0"
 repository = "https://github.com/parallaxsecond/rust-cryptoki"
 documentation = "https://docs.rs/crate/cryptoki-sys"
-rust-version = "1.66.0"
+rust-version = "1.77.0"
```
hug-dev: We currently have a MSRV of 1.66.0 so I guess we would need to decide to bump it before moving on... I am also surprised that the "MSRV - Execute CI script" CI step passed successfully with this change... Looking at its logs it does not seem to use the good Rust version :cry:

hug-dev (commented):
Thank you!! Looks good to me but we need to:

* decide if it's ok to bump the MRSV version. This mostly depends on where cryptoki is used: Ubuntu, Fedora, Microsoft, MacOS. We should have a look at what's the rustc version in the most common distribution
* investigate why the MRSV CI check does not work :sweat_smile: 

You might also need to rebase to latest `main` since I have now added the Windows build as required!

cryptoki-sys/Cargo.toml
wiktor-k: Whoops! I guess some lazy developer didn't check that!

/me checks out who touched CI lately

... 😽

I've read the docs of the action and filed https://github.com/parallaxsecond/rust-cryptoki/pull/294 on mobile. Sadly there seem to be quite a few problems (I've enabled web sign off in repo settings but I guess I won't be able to fix everything like this :()

(if you want to feel free to adjust, if not I'll look at it on Monday the earliest).

hug-dev: #294 is merged! You can now rebase and we should be able to go forward with this :rocket:

cryptoki/Cargo.toml
hug-dev, line 1: Not related with your PR and you don't need to change anything here but this made me thinking that we could add a dependabot check in some nightly CI which could check that we use the latest dependencies, and update if not!

hug-dev (approved):
Thank you that's great!

cryptoki/Cargo.toml
wiktor-k: Agreed. It took me quite a while to locate the actual change, even though I've seen it earlier :sweat_smile:

Maybe it's best to split this into several smaller PRs and keep this one only about the fix? :thinking: 

(Sorry for putting you for even more work :bow: )

hug-dev: @mematthias made a good job of splitting between different commit and there seems to be one just to address the bug, would that be enough :detective:?

wiktor-k: That first commit is still kind-of packed: https://github.com/parallaxsecond/rust-cryptoki/pull/289/commits/860d94555aeb137fec076cda69b6eace2eb8c44d and it includes stuff such as raising MRSV to 1.77 that has been done in another PR...

I just itches me to restructure these commits differently but... maybe it's just my OCD :sweat_smile:

cryptoki/Cargo.toml
```diff
@@ -13,17 +13,16 @@ documentation = "https://docs.rs/crate/cryptoki"
...
 [dependencies]
-bitflags = "1.3"
-libloading = "0.8.6"
-log = "0.4.14"
+bitflags = "2.9.1"
+libloading = "0.8.8"
+log = "0.4.27"
 cryptoki-sys = { path = "../cryptoki-sys", version = "0.4.0" }
-paste = "1.0.6"
+paste = "1.0.15"
```
Jakuje, line 20: for paste we have separate issue #279 where it is considered unmaintained (or feature complete -- depending on the point of view) and @hug-dev had a proposal to remove the dependency . Given that it is used only in one place and audit will show the issues with that, I would rather go with removing it than updating. @hug-dev can you open a PR with the proposed change in there to remove it?

cryptoki/Cargo.toml
Jakuje: I think it should be just rebased (instead of doing the merge commits on the way.

cryptoki/Cargo.toml
hug-dev: My proposal can only be done once the related change has been stabilized so I don't think we can rely on it yet :(

mematthias:
@hug-dev Thanks for the rebase and the new PR! Much appreciated. Should I go ahead and close this PR then?

Issue #208, opened 2024-04-09
Do not call C_Finalize if not initialized


Geobert:
Hi,

I’m in a mixed env with Rust code used in a C++ program. C++ side initialize the pkcs11 lib so Rust side can’t.
If Rust side calls `C_Finalize`, I get a random crash, which, I think, is because of a double free?

hug-dev:
Hello! Sorry for the long time without answer!

`C_Finalize` is called when then `Pkcs11` instance is dropped. We could potentially add a new constructor which builds an instance that does not auto-finalize :)

wiktor-k:
Hi @Geobert, I've made a slight adjustment to the API in #290 and would be super-happy if you could take a look if this solves your issue. The new function would be `Pkcs11Impl::new_unchecked` where you can pass the raw object and it would *not* be finalized.

Thanks in advance!

hug-dev:
> as it can seriously disrupt concurrent token access (or force applications to keep the Pkcs11 structure in a global indefinitely simply to avoid issues)

What do you mean? By using the `Arc` encapsulation (which is now done on our side but maybe soon should be done by the caller), one can pass a `Pkcs11` to different threads and be sure that it will on be finalized once, when all threads are done using it. That's what we do in the `login_feast` [test](https://github.com/parallaxsecond/rust-cryptoki/blob/2702bbaad3c6202803801361c9166d761192ff87/cryptoki/tests/basic.rs#L1465).

> There should also definitely be a way to initialize Pkcs11 without calling C_Initialize (for the same reason).

That is a good point!

simo5:
> What do you mean? By using the `Arc` encapsulation (which is now done on our side but maybe soon should be done by the caller), one can pass a `Pkcs11` to different threads and be sure that it will on be finalized once, when all threads are done using it. That's what we do in the `login_feast` [test](https://github.com/parallaxsecond/rust-cryptoki/blob/2702bbaad3c6202803801361c9166d761192ff87/cryptoki/tests/basic.rs#L1465).

This works fine and well when a single crate depends on Cryptoki, but as soon as a large project have two dependencies that independently use Cryptoki (or just independent implementations of PKCS11, like via C bindings) you may have conflicts.

For the most part you can use two separate Pkcs11 structures w/o much interference as long as they do not do untoward things like: uncontrollably C_Initialize, uncontrollably C_Finalize and uncontrollably C_CloseAllSessions on the same slot.

[...]

PR #290, opened 2025-06-23
Allow working with a non-finalizing `Pkcs11`

wiktor-k:
Hi folks,

There is much interest in a Pkcs11 that does not finalize the underlying implementation and I thought I'll try out one idea I had: splitting the `Pkcs11` into an owning and ref structs (similarly to how Rust does with `String` and `str`). I've made the `Pkcs11` finalize itself and the `Pkcs11Impl` just work on what's passed in. This way, folks that manage the lifecycle themselves can construct a new instance via `Pkcs11Impl::new_unchecked` which should resolve https://github.com/parallaxsecond/rust-cryptoki/issues/208.

I've also dropped inner Arc from `Pkcs11` (so that the concurrency is now managed by the client outside of Pkcs11) and made `Session` own a reference to the client. I'd further rename `Pkcs11Impl` into something like `Pkcs11Ref` but I want to keep the diff as small as possible, and in case you don't like it it's not worth the refactor.

See what you think about it @hug-dev and @Jakuje.

The alternative posted on Slack by @testingapisname is adding additional finalization flag: https://github.com/testingapisname/rust-cryptoki/commit/6a4e58df0229312c8967a5d1aaab825f1a63615d


Jakuje:
I do not have this use case and I do not feel knowledgeable enough of rust to be able to judge which of the approaches is better. But this one looks more straightforward, while the other more "bolted on top of the existing things".

Generally, I think it is a good idea, especially when we have people asking for this.

cryptoki/src/context/mod.rs
```diff
@@ -69,19 +68,78 @@ impl fmt::Debug for Pkcs11Impl {
 }
 
 impl Pkcs11Impl {
+    /// Initializes Pkcs11 using raw Pkcs11 object.
+    ///
+    /// # Safety
+    ///
+    /// `pkcs11_lib` must point to a valid Pkcs11 object.
+    pub unsafe fn new_unchecked(pkcs11_lib: cryptoki_sys::Pkcs11) -> Result<Self> {
```
hug-dev, line 90: I like the idea here! Was only wondering if we could expand the documentation (and maybe change the name?) to make clear that this is done so that `finalize` is not called on `Drop` and maybe explain the use-case.

wiktor-k: I've added better docs... happy to hear suggestions about the naming though... `new_non_finalizing`? `new_manage_your_own_lifecycle`? I'm kind-of bad at naming :sweat_smile:

Jakuje (approved):
I do not have any concerns here. Looks good! Thank you!

PR #292, opened 2025-06-25
Add more setters for EcKdf.

### LINKED ISSUE #281: EcKdf documentation unclear
The documentation for EcKdf [says](https://github.com/parallaxsecond/rust-cryptoki/blob/7844005/cryptoki/src/mechanism/elliptic_curve.rs#L76):

>  In the current version of this crate, only the null KDF is
> supported, which takes no shared data.  Therefore `'a` can always be inferred
> `'static`.

But there are two methods: `null` and [`sha256`](https://github.com/parallaxsecond/rust-cryptoki/blob/7844005/cryptoki/src/mechanism/elliptic_curve.rs#L100).  `sha256`'s signature is:

```
fn sha256() -> Self
```

That is, it doesn't have an elegant way to fail fast.  Is `sha256` supported? Is the error returned later?  Is the above comment out of date?

Thanks!

nwalfield:
Adds setters for the ANSI X9.63 KDFs (`EcKdf::sha1`, etc) and the NIST SP800-56A KDFs (`EcKdf::sha1_sp800`) to `EcKdf`.

Fixes `EcKdf::sha256`.

Fixes #281.

The SP800 variants were [recently added to Kryoptic](https://github.com/latchset/kryoptic/pull/273).

I've tested this using Sequoia, but I haven't included any unit tests.  First, the unit tests will fail, because the current released version of Kryoptic doesn't include support for the SP800 yet.  Second, creating test vectors will be a huge amount of work.  Given how simple the implementation is, do you want unit tests for these functions?

Thanks for taking a look.


cryptoki/tests/basic.rs
```diff
@@ -687,6 +687,91 @@ fn derive_key() -> TestResult {
...
+    let session = pkcs11.open_rw_session(slot)?;
+
+    // log in the session
+    session.login(UserType::User, Some(&AuthPin::new(USER_PIN.into())))?;
+
+    // sha256_sp800
+    let key = dehex("F00670C5F139E7E6C2511EA04FF507AFBBE237CE3C71A89CA59A1C5CF8856562").expect("valid hex");
+    let ephemeral_key = dehex("533B3F09E53B3DEED661A13F7A7D9694AB71CE156C728E00DEE87A1EE3A14C4A").expect("valid hex");
+    let kdf_param = dehex("0A2B0601040197550105011203010807416E6F6E796D6F75732053656E646572202020205633A4C5AE4305BC0FE2ABB699A8EE54632790A0").expect("valid hex");
+    let derivation = dehex("AF8CE51D0139A6D60831A9BABAB20186").expect("valid hex");
```
Jakuje: I would use `hex::decode()` as we have `hex` already in the list of dependencies (but looks unused though).

nwalfield: Good point.  Fixed.

cryptoki/tests/basic.rs
```diff
@@ -687,6 +687,82 @@ fn derive_key() -> TestResult {
...
+
+    let params = Ecdh1DeriveParams::new(kdf, &ephemeral_key);
+
+    let shared_secret = session.derive_key(
+        &Mechanism::Ecdh1Derive(params),
+        private,
+        &[
+            Attribute::Class(ObjectClass::SECRET_KEY),
+            Attribute::KeyType(KeyType::GENERIC_SECRET),
+            Attribute::ValueLen((derivation.len() as u64).into()),
```
Jakuje: This still does not work on 32b arch. We had more of them fixed recently in 80b22b2b58ce138304b1e0e228e17edada88a195, but I think all of them are constants.

Not sure what would be the cleanest way to do this, but probably some From for u32 to keep it safe? Or just Ulong::new(derivation.len()), but it makes the API ugly, i think...

nwalfield: Thanks for the hint.  I switched to `Ulong::new(derivation.len())`.

Jakuje: Still it probably needs
```
Attribute::ValueLen(Ulong::new(derivation.len() as u64)),
```
or the `try_into().unwrap()` as suggested by the rust ...

nwalfield: I updated the change to use `Attribute::ValueLen(Ulong::new(derivation.len() as u64)),`.

Jakuje: Still not right. Sorry for misleading you in the wrong direction. I think it would be good to have the Ulong::new() constructor with some platform-agnostic type as an argument. The CK_ULONG is problematic here.

nwalfield: It seems I've finally figured out the right incantation :D

nwalfield:
> > The MR is currently failing, because this relies on Kryoptic support the SP 800 KDF, which has been added to Kryoptic, but has not yet been released.
> 
> I think the cleanest way would be now to expect it to fail now (with appropriate comment or even ideally in separate commit that can be reverted later) and change it (revert that separate commit) once the new kryoptict will be released.
> 
> The other option would be to start building kryoptic from source, but that is also outside of the scope of this PR.

I added two helper functions: one to check if we're using Kryoptic and one to get the token's version.  I then modified the test to check if we're using at least Kryopic 1.3, which I assume will be the next version of Kryoptic.  What do you think?

cryptoki/tests/basic.rs
```diff
@@ -687,6 +687,90 @@ fn derive_key() -> TestResult {
...
+
+    let (pkcs11, slot) = init_pins();
+
+    if is_kryoptic() {
+        let (major, minor) = get_firmware_version(&pkcs11, slot);
+        // Kryoptic added support for sha256_sp800 in version 1.3.
+        if !(major > 1 || minor >= 3) {
+            return Ok(());
+        }
+    }
```
Jakuje: One last note I would have would be adding a print output that we are actually skipping the test as the rust does not distinguish the passed an skipped tests (similarly as in the other tests that are skipped like this).

nwalfield: I added that, but it is not clear how much it will help: `stdout` / `stderr` are only shown if `--no-capture` is used.

Jakuje (approved):

hug-dev (approved):

PR #294, opened 2025-06-27
Set installed toolchain as default and bump MSRV to 1.77

wiktor-k:
The docs for the action indicate that unless specified the toolchain will not be used by default.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
           - riscv64gc-unknown-linux-gnu
         toolchain:
           - stable
-          - "1.66.0" # MSRV
+          - "1.77" # MSRV
     steps:
       - uses: actions/checkout@v4
       - name: Setup Rust toolchain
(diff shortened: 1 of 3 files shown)


wiktor-k:
Hmm... okay, fixing this uncovered another issue: the bindings are using `std::mem::offset_of` which, [according to docs](https://dev-doc.rust-lang.org/stable/std/mem/macro.offset_of.html) was introduced in Rust 1.77.

Do you think we should upgrade our MSRV or rather find a way to generate sources without `offset_of` or something else entirely? :thinking: 

Apparently it is possible to set Rust target when generating bindings: https://github.com/rust-lang/rust-bindgen/issues/2960#issuecomment-2452253259

CC @Jakuje @hug-dev

Jakuje (approved):
no concerns here

wiktor-k:
Thanks :wave: . Let's wait for @hug-dev since he touched these parts previously and may have valuable input.

hug-dev:
Checking at [Rust packages in Ubuntu](https://documentation.ubuntu.com/ubuntu-for-developers/reference/availability/rust/), seems like 1.77 is available on old distributions too so I think we can go with it!

hug-dev (approved):
Thanks! I am thinking that we could as well write what the MSRV is somewhere in a README!

PR #298, opened 2025-07-10
Remove ci.sh script to run all from workflow

hug-dev:
The ci.sh script was used in the main script, but only to execute cargo test. It was also used in the nightly workflow but it's not needed since the same steps are run in the merge workflow. Also prevents the nightly from failing.


hug-dev:
True but I thought that the "Check for errors" workflow would check it with the `--all-features` flag in the following command?

```
cargo check --target ${{ matrix.target }} --all-features --workspace --all-targets
```

Issue #299, opened 2025-07-17
Support vendor defined error codes


hug-dev:
> Vendor defined error codes are still not handled nicely. As a vendor, you probably do something like this:
> 
> ```c
> #define CKR_MY_ERROR_A (CKR_VENDOR_DEFINED+1)
> #define CKR_MY_ERROR_B (CKR_VENDOR_DEFINED+2)
> #define CKR_MY_ERROR_C (CKR_VENDOR_DEFINED+3)
> ```
> 
> In cryptoki 0.10.0, these [will be mapped](https://github.com/parallaxsecond/rust-cryptoki/blob/cryptoki-0.10.0/cryptoki/src/error/rv.rs#L120) to RvError::GeneralError. Therefore, as a cryptoki user I cannot distinguish between the vendor-defined errors.
> 
> From [Section 3.6 of the standard v3.1](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203196):
> 
> Return values CKR_VENDOR_DEFINED and above are permanently reserved for token vendors.
> 
> So it should be safe to map these to something like `RvError::VendorDefined(code)`. 

 _Originally posted by @thgoebel in [#54](https://github.com/parallaxsecond/rust-cryptoki/issues/54#issuecomment-3083797101)_

A similar approach than #246 should probably be acceptable there!

PR #300, opened 2025-07-19
Expose vendor defined error code

thgoebel:
Fixes https://github.com/parallaxsecond/rust-cryptoki/issues/299

Tested with:

- The unit tests I added
- A Securosys [CloudHSM Sandbox](https://cloud.securosys.com/cloudhsm) environment

This contribution is in personal capacity.

### DIFF
--- a/cryptoki/src/error/rv.rs
+++ b/cryptoki/src/error/rv.rs
@@ -8,7 +8,7 @@ use super::{Error, Result, RvError};
 use cryptoki_sys::*;
 use log::error;
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 /// Return value of a PKCS11 function
 pub enum Rv {
     /// The function exited successfully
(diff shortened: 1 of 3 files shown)


cryptoki/src/error/rv.rs
```diff
@@ -137,3 +138,32 @@ impl Rv {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
```
hug-dev, line 143: Thanks for the tests!

cryptoki/src/object.rs
```diff
@@ -14,7 +14,7 @@ use std::fmt::Formatter;
 use std::mem::size_of;
 use std::ops::Deref;
 
-const MAX_CU_ULONG: CK_ULONG = !0;
+pub(crate) const MAX_CU_ULONG: CK_ULONG = !0;
```
hug-dev: It's not your doing but I am wondering why this is not called `MAX_CK_ULONG` :thinking: ?

hug-dev (dismissed):
Thank you!

wiktor-k (dismissed):
LGTM :+1: thanks!

cryptoki/src/object.rs
thgoebel: Done, renamed and moved to the crate root.

cryptoki/src/lib.rs
```diff
@@ -49,6 +49,8 @@ pub mod types;
 
 use cryptoki_sys::CK_UTF8CHAR;
 
+const MAX_CK_ULONG: cryptoki_sys::CK_ULONG = !0;
```
Jakuje: Whats wrong with using `CK_ULONG::MAX` directly rather than exporting the alias like this?

hug-dev (dismissed):
Nice addition!

cryptoki/src/lib.rs
thgoebel: Good question.

CK_ULONG is a type alias for std::os::raw::c_ulong. For that, Rust Analyzer's autocomplete in VS Code doesn't show ::MAX to me. When I add it anyway, it compiles. Looking one step deeper: c_ulong is always either u32 or u64:

- https://doc.rust-lang.org/std/os/raw/type.c_ulong.html
- https://github.com/rust-lang/rust/blob/master/library/core/src/ffi/primitives.rs

And in the end, it will do the same thing anyway: https://github.com/rust-lang/rust/blob/0162cc5/library/core/src/num/uint_macros.rs#L45

So I think it should be safe to use ::MAX directly. I'll change it.

PR #302, opened 2025-07-29
Fix the link checker job

wiktor-k:


wiktor-k:
Thanks for the explanation! I think that by now these two are pretty stable... So... I propose we make them required too. Objections? CC @hug-dev

PR #303, opened 2025-07-30
Fixup PKCS#11 headers definition (`C_UnwrapKey`)

Jakuje:
We identified this issue in the public headers. Fortunately, turns out it should not have any functional difference. Also updates the bindgen to 0.72 while I regenerate the bindings.


wiktor-k (approved):

hug-dev (approved):
Nice spot!

PR #304, opened 2025-08-12
Add support for Pkcs#11 Profile objects

Jakuje:
This allows selecting new object types, profiles, from the token and reading their attributes, consisting of a profile ID the token implements:

https://docs.oasis-open.org/pkcs11/pkcs11-profiles/v3.1/os/pkcs11-profiles-v3.1-os.html

The specs is a bit vague and my understanding is that the profiles can overlap, but are not always superset of some other profile, which makes me thinking the token can have several profile objects.

Currently, this is not implemented by neither kryoptic nor softhsm, but I filled a RFE for kryoptic (https://github.com/latchset/kryoptic/issues/305) so once we will have it in, I will try to write some test.

### DIFF
--- a/cryptoki/src/object.rs
+++ b/cryptoki/src/object.rs
@@ -110,6 +110,8 @@ pub enum AttributeType {
     PublicExponent,
     /// DER-encoding of the SubjectPublicKeyInfo
     PublicKeyInfo,
+    /// Profile ID
+    ProfileId,
     /// Seed to derive private key
     Seed,
     /// Determines if the key is sensitive
@@ -269,6 +271,7 @@ impl AttributeType {
             CKA_UNIQUE_ID => String::from(stringify!(CKA_UNIQUE_ID)),
             CKA_SEED => String::from(stringify!(CKA_SEED)),
             CKA_PARAMETER_SET => String::from(stringify!(CKA_PARAMETER_SET)),
+            CKA_PROFILE_ID => String::from(stringify!(CKA_PROFILE_ID)),
             CKA_VENDOR_DEFINED..=CK_ULONG::MAX => {
                 format!("{}_{}", stringify!(CKA_VENDOR_DEFINED), val)
             }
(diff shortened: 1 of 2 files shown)


cryptoki/src/object.rs
```diff
@@ -1708,3 +1731,94 @@ impl TryFrom<CK_CERTIFICATE_TYPE> for CertificateType {
...
+            _ => format!("unknown ({profile_id:08x})"),
+        }
+    }
+}
+
+impl std::fmt::Display for ProfileIdType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", ProfileIdType::stringify(self.val))
+    }
+}
```
wiktor-k: Is there a reason the `ProfileIdType::stringify` function is not just inlined like so:

```suggestion
}

impl std::fmt::Display for ProfileIdType {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}", match self.val {
            CKP_BASELINE_PROVIDER => stringify!(CKP_BASELINE_PROVIDER),
            CKP_EXTENDED_PROVIDER => stringify!(CKP_EXTENDED_PROVIDER),
            CKP_AUTHENTICATION_TOKEN => stringify!(CKP_AUTHENTICATION_TOKEN),
            CKP_PUBLIC_CERTIFICATES_TOKEN => {
                stringify!(CKP_PUBLIC_CERTIFICATES_TOKEN)
            }
            CKP_COMPLETE_PROVIDER => stringify!(CKP_COMPLETE_PROVIDER),
            CKP_HKDF_TLS_TOKEN => stringify!(CKP_HKDF_TLS_TOKEN),
            profile_id => return write!(f, "unknown ({profile_id:08x})"),
        })
    }
}
```

?

cryptoki/src/object.rs
```diff
@@ -1708,3 +1731,94 @@ impl TryFrom<CK_CERTIFICATE_TYPE> for CertificateType {
...
+
+impl Deref for ProfileIdType {
```
wiktor-k: I know we have a couple of these but I'd refrain from adding more if possible.

IMO Deref should be used for smart pointers, not new-types and some projects [already regret abusing it](https://gitlab.com/sequoia-pgp/sequoia/-/issues/1092).

Maybe in this case `AsRef` could be used as it is a lighter alternative, with no compiler magic:

```suggestion
impl AsRef<CK_PROFILE_ID> for ProfileIdType {
```

wiktor-k (commented):
Looks good in general, a couple of nits if you don't mind :)

cryptoki/src/object.rs
Jakuje: Likely copy&paste from the previous structure above. I do not have a preference to either way, but I agree that doing it directly is less complicated so I will apply your suggestion.

cryptoki/src/object.rs
Jakuje: Good point! Fixed!

hug-dev (approved):

PR #306, opened 2025-08-18
Add support for PKCS#11 3.2 validation objects

Jakuje:
The PKCS#11 3.2 introduces a way to query the session for the validation flags of the last operation. This is done with the new API `C_GetSessionValidationFlags` which is being exposed now also to the users of this crate.

It also defines the new attribute specifying if the given object is matching the requirements for the validation.

Last but not least there is new validation object exposing information about the validation itself.

There is currenly almost no coverage for these, as this is implemented only by kryoptic (as far as I know) and not enabled in the default build we are using in CI.

Opening as draft as it depends on some fixes in kryoptic (latchset/kryoptic#315) as well its based on other code changes here (#304).

### DIFF
--- a/.github/workflows/kryoptic-fips.yml
+++ b/.github/workflows/kryoptic-fips.yml
@@ -1,0 +1,136 @@
+---
+name: Test kryoptic FIPS module
+
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  build:
+    name: Test kryoptic FIPS module
+    runs-on: ubuntu-22.04
+    container: quay.io/fedora/fedora:latest
+    steps:
+      #################
+      ### DNF cache ###
+      #################
+      - name: Get Date for DNF cache entry
+        id: get-date
+        run: |
+          echo "date=$(/bin/date -u "+%Y%V")" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: Restore DNF cache
+        uses: actions/cache/restore@v4
+        id: cache-dnf
+        with:
+          path: "/var/cache/libdnf5"
+          key: fedora-dnf-${{ steps.get-date.outputs.date }}
+
+      - name: Install Dependencies
+        run: |
+          dnf -y install git cargo clang-devel openssl-devel sqlite-devel \
+            'perl(FindBin)' 'perl(lib)' 'perl(File::Compare)' \
+            'perl(File::Copy)' 'perl(bigint)' 'perl(Time::HiRes)' \
+            'perl(IPC::Cmd)' 'perl(Pod::Html)' 'perl(Digest::SHA)' \
+            'perl(Module::Load::Conditional)' 'perl(File::Temp)' \
+            'perl(Test::Harness)' 'perl(Test::More)' 'perl(Math::BigInt)' \
+            'perl(Time::Piece)' zlib-devel sed sqlite-devel
+
+      - name: DNF cache
+        if: ${{ steps.cache-dnf.outputs.cache-hit != 'true' }}
+        uses: actions/cache/save@v4
+        with:
+          path: "/var/cache/libdnf5"
+          key: fedora-dnf-${{ steps.get-date.outputs.date }}
+
+      #####################
+      ### OpenSSL build ###
+      #####################
+      - name: Setup OpenSSL 3.5
+        id: ossl-setup
+        run: |
+          OPENSSL_BRANCH="openssl-3.5"
+
+          cd ..
+          git clone https://github.com/openssl/openssl.git \
+                    --single-branch --branch $OPENSSL_BRANCH openssl
+          cd openssl
+          echo "KRYOPTIC_OPENSSL_SOURCES=$PWD" >> "$GITHUB_ENV"
+          echo "cacheid=${{ runner.os }}-ossl-$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+
+      - name: Restore OpenSSL build if cached
+        uses: actions/cache/restore@v4
+        id: cache
+        with:
+          path: ${{ env.KRYOPTIC_OPENSSL_SOURCES }}
+          key: ${{ steps.ossl-setup.outputs.cacheid }}
+
+      - name: Build OpenSSL
+        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
+        run: |
+          pushd ${{ env.KRYOPTIC_OPENSSL_SOURCES }}
+          ./Configure
+          make
+
+      - name: Cache OpenSSL 3.5 build
+        if: ${{ steps.cache.outputs.cache-hit != 'true' }}
+        uses: actions/cache/save@v4
+        with:
+          path: ${{ env.KRYOPTIC_OPENSSL_SOURCES }}
+          key: ${{ steps.ossl-setup.outputs.cacheid }}
+
+      ######################
+      ### kryoptic build ###
+      ######################
+      - name: Setup kryoptic
+        run: |
+          KRYOPTIC_REVISION="v1.3.1"
+
+          cd ..
+          git clone https://github.com/latchset/kryoptic.git \
+              --depth 1 --single-branch --revision $KRYOPTIC_REVISION kryoptic
+
+      - name: Generate lock file
+        run: |
+          cd ../kryoptic &&
+          cargo generate-lockfile
+
+      - name: Cache Rust dependencies
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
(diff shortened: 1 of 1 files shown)


cryptoki/src/session/validation.rs
```diff
@@ -0,0 +1,68 @@
...
+    val: CK_SESSION_VALIDATION_FLAGS_TYPE,
+}
+
+impl ValidationFlagsType {
+    /// Check the last operation met all requirements of a validated mechanism.
+    pub const VALIDATION_OK: ValidationFlagsType = ValidationFlagsType {
+        val: CKS_LAST_VALIDATION_OK,
+    };
+
+    pub(crate) fn stringify(flags: CK_SESSION_VALIDATION_FLAGS_TYPE) -> String {
```
wiktor-k: I think this can be inlined into `Display`, like so: https://github.com/parallaxsecond/rust-cryptoki/pull/304/files#diff-8eed47b951056624bcbe744076eac12fab93c4fe6c655c937351243a0db816daR1775

cryptoki/src/session/validation.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+}
+
+impl std::fmt::Display for ValidationFlagsType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", ValidationFlagsType::stringify(self.val))
+    }
+}
+
+impl Deref for ValidationFlagsType {
```
wiktor-k: `AsRef` would be safer IMO.

cryptoki/tests/basic.rs
```diff
@@ -3822,13 +3821,87 @@ fn unique_id() -> TestResult {
...
+    // generate a secret key
+    let key = session.generate_key(&Mechanism::AesKeyGen, &generate_template)?;
+
+    // we can get the ObjectValidationFlags attribute
+    let attrs = session.get_attributes(key, &[AttributeType::ObjectValidationFlags])?;
+    if is_softhsm() {
+        // SoftHSM does not support this attribute at all
+        assert_eq!(attrs.len(), 0);
+    } else {
+        // Kryoptic supports the ObjectValidationFlag only if it is built as a FIPS provider
```
wiktor-k, line 4265: :thinking: is it possible to built it this way for our CI? I think it'd be cool to have some real results here :)

Jakuje: Sure. But its slow as it builds also the whole OpenSSL :) I was trying this locally and there are some more limitations and corner cases, which do not work or work differently and which we would have to workaround. But with the validation API, it should be quite easy to distinguish the FIPS and non-FIPS kryoptic version.

cryptoki/src/session/validation.rs
Jakuje: Resolved.

cryptoki/src/session/validation.rs
Jakuje: Agreed.  Done.

cryptoki/tests/basic.rs
Jakuje: Added the FIPS build of kryoptic and adjusted the tests to work with it. The build is slow, but after OpenSSL build will be cached, it should be faster (but still slower than other tests).

cryptoki-sys/build.rs
```diff
@@ -31,6 +31,7 @@ mod generate {
                 ("CKC_", "CK_CERTIFICATE_TYPE"),
                 ("CKD_", "CK_EC_KDF_TYPE"),
                 ("CKF_", "CK_FLAGS"),
+                ("CKV_", "CK_ULONG"),
```
hug-dev, line 34: Good catch!

cryptoki/src/session/validation.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn from(val: ValidationFlagsType) -> Self {
+        *val.as_ref()
+    }
+}
+
+impl Session {
+    /// Get requested validation flags from the session
+    ///
+    /// The only supported flag as for PKCS#11 3.2 is `ValidationFlagsType::VALIDATION_OK`
+    pub fn get_validation_flags(&self, flags_type: ValidationFlagsType) -> Result<CK_FLAGS> {
```
hug-dev, line 53: Could we use any of the existing abstraction for the `CK_FLAGS` returned here?

.github/workflows/kryoptic-fips.yml
```diff
@@ -0,0 +1,137 @@
+---
+name: Test kryoptic FIPS module
```
hug-dev, line 2: Shall I add it to the required workflow to pass for a PR?

hug-dev (dismissed):
Thank you!

.github/workflows/kryoptic-fips.yml
wiktor-k: I think now this can be done using the "[Check if all checks succeeded (pull_request)](https://github.com/parallaxsecond/rust-cryptoki/actions/runs/17860753123/job/50880857592?pr=306)" job, thus, Jakub can adjust it himself.

cryptoki/src/session/validation.rs
Jakuje: The resulting `CK_FLAGS` value here are vendor specific that are based on the  validation object (`CKA_VALIDATION_FLAG` attribute). That said, we would have to define the ValidationFlags abstraction, but the user would have to define its own vendor specific values. I did not work with the vendor specific atrributes in rust-cryptoki so far, but is probably the cleanest approach. It will also make it more clear that these two flags should be related to each other.

https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/pkcs11-spec-v3.2.html#_Toc195693109

Jakuje: ugh ... not sure if we can easily make a vendor defined bitfield though ...

hug-dev: I see.. we can probably keep it like that for now and find a good abstraction when it's getting used more?

.github/workflows/kryoptic-fips.yml
Jakuje: I am not sure if it would not need to be in the same workflow file to work though to be able to use in the "check if all checks succeeded". Given that how large this workflow is, I did not want to mess the main ci.yml with it for now.

I hope this job will be stable, but I would rather keep it for some time non-mandatory and add it to the required just after some time we will see it will work as expected to avoid working around some required jobs.

cryptoki/src/session/validation.rs
Jakuje: Yeah, that would be my preference as I am not sure where to start defining the vendor specific bit field.

hug-dev (dismissed):

.github/workflows/kryoptic-fips.yml
```diff
@@ -0,0 +1,137 @@
...
+      #################
+      ### the tests ###
+      #################
+      - name: Run test script
+        env:
+          KRYOPTIC_CONF: /tmp/kryoptic.sql
+          TEST_PKCS11_MODULE: /__w/rust-cryptoki/kryoptic/target/debug/libkryoptic_pkcs11.so
+          OUT_DIR: /__w/rust-cryptoki/kryoptic/target/debug/deps/
+        run: |
+          RUST_BACKTRACE=1 cargo build --all-features &&
```
wiktor-k: I think `RUST_BACKTRACE=1` can also be moved to `env`.
I don't think we need `cargo build` here... `cargo check` should be sufficient :thinking:

cryptoki/tests/basic.rs
```diff
@@ -2378,7 +2484,7 @@ fn kbkdf_feedback_mode() -> TestResult {
     let dkm_length_format = KbkdfDkmLengthFormat::new(
         KbkdfDkmLengthMethod::SumOfKeys,
```
wiktor-k, line 2768: Could you explain why `SumOfKeys` is OK here but in the fragment above it was changed to `SumOfSegments`? :thinking: Thanks!

Jakuje: I did not dig deep enough, but generally, kryoptic implements all possible PKCS#11 combinations in Rust. But for the FIPS we want to use the OpenSSL implementation, which has limitation that the Counter mode supports only sum of segments, while Feedback mode supports only the sum of keys.

I reverted the changes to the double-pipeline KDFs as they are untested (I think not implemented in OpenSSL) and I am not sure how they will work.

.github/workflows/kryoptic-fips.yml
Jakuje: Moved the env + removed the separate build

wiktor-k (approved):
LGTM :+1: thanks!

Jakuje:
Thanks! That was fast. Lest hope the CI with FIPS kryoptic will work :) Otherwise expect follow-up PR :)

PR #307, opened 2025-09-09
Add SLH-DSA mechanism

Jakuje:
This moves the ML-DSA mechanisms params around as they are shared by these two mechanism groups (I hope its ok, given that we did not do a release with these yet).

Currently, the mechanisms are supported by kryoptic (latchset/kryoptic#316), but I do not have a Fedora build yet so its not testable in CI.

### DIFF
--- a/cryptoki/src/object.rs
+++ b/cryptoki/src/object.rs
@@ -1096,7 +1096,7 @@ impl TryFrom<Ulong> for ParameterSetType {
 impl From<MlKemParameterSetType> for ParameterSetType {
     fn from(val: MlKemParameterSetType) -> Self {
         ParameterSetType {
-            val: Ulong::new(*val),
+            val: Ulong::new(*val.as_ref()),
         }
     }
 }
@@ -1104,7 +1104,15 @@ impl From<MlKemParameterSetType> for ParameterSetType {
 impl From<MlDsaParameterSetType> for ParameterSetType {
     fn from(val: MlDsaParameterSetType) -> Self {
         ParameterSetType {
-            val: Ulong::new(*val),
+            val: Ulong::new(*val.as_ref()),
+        }
+    }
+}
+
+impl From<SlhDsaParameterSetType> for ParameterSetType {
+    fn from(val: SlhDsaParameterSetType) -> Self {
+        ParameterSetType {
+            val: Ulong::new(*val.as_ref()),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


cryptoki/src/object.rs
```diff
@@ -1253,6 +1257,133 @@ impl From<ParameterSetType> for MlDsaParameterSetType {
...
+    /// SLH-DSA-SHA2-256f
+    pub const SHA2_256F: SlhDsaParameterSetType = SlhDsaParameterSetType {
+        val: CKP_SLH_DSA_SHA2_256F,
+    };
+    /// SLH-DSA-SHAKE-256f
+    pub const SHAKE_256F: SlhDsaParameterSetType = SlhDsaParameterSetType {
+        val: CKP_SLH_DSA_SHAKE_256F,
+    };
+
+    pub(crate) fn stringify(val: CK_SLH_DSA_PARAMETER_SET_TYPE) -> String {
```
wiktor-k: Sorry for bothering you with the same thing but I think we can just inline this into `Display`'s formatter, just like in https://github.com/parallaxsecond/rust-cryptoki/pull/304/files#diff-8eed47b951056624bcbe744076eac12fab93c4fe6c655c937351243a0db816daR1775

Jakuje: Sure. Can change this too here.

cryptoki/src/object.rs
```diff
@@ -1253,6 +1257,133 @@ impl From<ParameterSetType> for MlDsaParameterSetType {
...
+    /// SLH-DSA-SHAKE-256s
+    pub const SHAKE_256S: SlhDsaParameterSetType = SlhDsaParameterSetType {
+        val: CKP_SLH_DSA_SHAKE_256S,
+    };
+    /// SLH-DSA-SHA2-256f
+    pub const SHA2_256F: SlhDsaParameterSetType = SlhDsaParameterSetType {
+        val: CKP_SLH_DSA_SHA2_256F,
+    };
+    /// SLH-DSA-SHAKE-256f
+    pub const SHAKE_256F: SlhDsaParameterSetType = SlhDsaParameterSetType {
```
Jakuje, line 1313: One thing I was not sure here was the short names here, `SHAKE_256F`, but given that the name really does not conflict with anything else and from context it is clear it is s SLH-DSA parameter `SlhDsaParameterSetType::SHAKE_256F`, I think it is ok and we do not need to repeat the SLH_DSA again (as we need to do in the C code with global defines).

wiktor-k: > I think it is ok and we do not need to repeat the SLH_DSA again (as we need to do in the C code with global defines).

Yes, agreed. The type effectively forms a namespace here so I think it's :ok_hand:

wiktor-k (approved):

PR #310, opened 2025-09-11
Feat: Add mechanisms for simple key derivations

jacobprudhomme:
Author: Jacob Prud'homme

# Description

This PR adds support for the simple key derivation mechanisms (defined in section 6.43 of v3.1 of PKCS#11)

# Motivation

We would like to use some of these mechanisms in a cryptography library we are building, and figured it would make sense to add support for all for them since they are all simple

## Summary of Changes

* Added `CONCATENATE_BASE_AND_KEY`, `CONCATENATE_BASE_AND_DATA`, `CONCATENATE_DATA_AND_BASE`, `XOR_BASE_AND_DATA`, `EXTRACT_KEY_FROM_KEY` mechanism types and their corresponding `Mechanism` variants, as well as new parameter types for some of these (in a new `misc` module)
* Added corresponding tests for these new mechanisms
  * Currently untestable, as Kryoptic does not support these. Issue opened at https://github.com/latchset/kryoptic/issues/323
* Made unrelated code cleanup (start using `make_mechanism()` for `GcmParams` and `GcmMessageParams`, for some reason this was not the case before)

### DIFF
--- a/cryptoki/tests/basic.rs
+++ b/cryptoki/tests/basic.rs
@@ -798,6 +798,8 @@ fn derive_key_concatenation_two_keys() -> TestResult {
         Attribute::KeyType(KeyType::GENERIC_SECRET),
         Attribute::Value(key_value[..3].to_vec()),
         Attribute::Derive(true),
+        Attribute::Sensitive(false),
+        Attribute::Extractable(true),
     ];
     let key2_template = [
         Attribute::Token(true),
@@ -806,6 +808,8 @@ fn derive_key_concatenation_two_keys() -> TestResult {
         Attribute::KeyType(KeyType::GENERIC_SECRET),
         Attribute::Value(key_value[3..].to_vec()),
         Attribute::Derive(true),
+        Attribute::Sensitive(false),
+        Attribute::Extractable(true),
     ];
 
     let key1 = session.create_object(&key1_template)?;


cryptoki/tests/basic.rs
```diff
@@ -772,6 +773,254 @@ fn derive_key_sp800() -> TestResult {
...
+
+    let key = session.create_object(&key_template)?;
+
+    // Derive key by extracting subset of base key
+    let derived_key_template = [
+        Attribute::Token(true),
+        Attribute::Private(false),
+        Attribute::Sensitive(false),
+        Attribute::Extractable(true),
+    ];
```
Jakuje, line 1027: From my reading of the specs, you need to have either key type with defined size or VALUE_LEN attribute for this mechanism to properly work

cryptoki/tests/basic.rs
```diff
@@ -772,6 +773,254 @@ fn derive_key_sp800() -> TestResult {
...
+    let derived_key_value = session
+        .get_attributes(derived_key, &[AttributeType::Value])?
+        .remove(0);
+    let derived_key_value = if let Attribute::Value(value) = derived_key_value {
+        value
+    } else {
+        panic!("Expected value attribute.");
+    };
+
+    assert_eq!(&derived_key_value, &data_value[3..]);
```
Jakuje: The Extract params has bit offset, not a byte offset, so this assert will fail too.

cryptoki/src/mechanism/misc.rs
```diff
@@ -0,0 +1,63 @@
...
+#[repr(transparent)]
+pub struct ExtractParams(CK_EXTRACT_PARAMS);
+
+impl ExtractParams {
+    /// Construct parameter from index to extract the derived key from the base
+    /// key.
+    ///
+    /// # Arguments
+    ///
+    /// * `index` - The index from which to extract the derived key from the base key.
```
Jakuje: this should probably be clarified that it is a bit index, not a byte index.

Jakuje:
I just pushed update of kryoptic 1.3.1 into Fedora so next time you run the CI, you should have these algorithms supported.

cryptoki/tests/basic.rs
jacobprudhomme: Good catch!

Jakuje (approved):
ci is happy now. Code looks good. Thank you for the contribution and finding bugs in kryoptic!

PR #312, opened 2025-09-15
Fix WrapKeyAuthenticated parameter type from pkcs11 header

Jakuje:
These functions are not used by rust-cryptoki and not in any release versions so far.

https://github.com/latchset/pkcs11-headers/pull/2


Jakuje:
Used #305 to re-generate the new bindings.

wiktor-k (approved):
The change looks good but it'd be even better if, instead of "fix another bug" it had at least a single sentence about the bug.

Thanks for the reference I got to the bottom of it :)

hug-dev (approved):

Issue #316, opened 2025-10-01
Consider opening visibility on ObjectHandle and SessionHandle


jack-shannon:
I am working on handling a vendor extension that adds additional functions to the standard and so far all of the vendor specific accommodations work well. The one issue I've run into is with handles - accessing the underlying `SessionHandle` is `pub(crate)` as is instantiating an `ObjectHandle` or accessing the underlying handle, all of which I need to be able to do to add the extension cleanly.

Would changing these to `pub` be an unreasonable request/contribution?

jack-shannon:
A small thing could be to put pub accessors behind a feature that would force users to opt in/acknowledge they're doing something potentially less safe.

PR #317, opened 2025-10-15
Add pub unsafe accessors for object/session handles and ObjectHandle new

jhshannon17:
As detailed in this [issue](https://github.com/parallaxsecond/rust-cryptoki/issues/316), in order to allow for vendor extensions to the standard, raw session/object handles need to be accessible outside of the crate. Creating new methods marked `unsafe` was the preferred way to allow this extension.

- adds `new_from_raw` method to ObjectHandle to allow creation
- adds `raw_handle` to ObjectHandle to get the underlying handle
- adds `raw_handle` to Session to get the underlying handle

### DIFF
--- a/cryptoki/src/object.rs
+++ b/cryptoki/src/object.rs
@@ -1193,7 +1193,15 @@ impl ObjectHandle {
         ObjectHandle { handle }
     }
 
-    pub(crate) fn handle(&self) -> CK_OBJECT_HANDLE {
+    /// Create a new object handle from a raw handle.
+    /// # Safety
+    /// Considered unsafe due to ability for client to arbitrarily create object handles.
+    pub unsafe fn new_from_raw(handle: CK_OBJECT_HANDLE) -> Self {
+        ObjectHandle { handle }
+    }
+
+    /// Get the raw handle of the object.
+    pub fn handle(&self) -> CK_OBJECT_HANDLE {
         self.handle
     }
 }
--- a/cryptoki/src/session/mod.rs
+++ b/cryptoki/src/session/mod.rs
@@ -78,7 +78,8 @@ impl Session {
     /// This will be called on drop as well.
     pub fn close(self) {}
 
-    pub(crate) fn handle(&self) -> CK_SESSION_HANDLE {
+    /// Get the raw handle of the session.
+    pub fn handle(&self) -> CK_SESSION_HANDLE {
         self.handle
     }
 
--- a/cryptoki/tests/basic.rs
+++ b/cryptoki/tests/basic.rs
@@ -4296,3 +4296,55 @@ fn validation() -> TestResult {
 
     Ok(())
 }
+
+#[test]
+#[serial]
+fn object_handle_new_from_raw() -> TestResult {
+    let (pkcs11, slot) = init_pins();
+
+    // open a session
+    let session = pkcs11.open_rw_session(slot)?;
+
+    // log in the session
+    session.login(UserType::User, Some(&AuthPin::new(USER_PIN.into())))?;
+
+    // get mechanism
+    let mechanism = Mechanism::RsaPkcsKeyPairGen;
+
+    let public_exponent: Vec<u8> = vec![0x01, 0x00, 0x01];
+    let modulus_bits = 2048;
+
+    // pub key template
+    let pub_key_template = vec![
+        Attribute::Token(true),
+        Attribute::Private(false),
+        Attribute::PublicExponent(public_exponent),
+        Attribute::ModulusBits(modulus_bits.into()),
+        Attribute::Verify(true),
+    ];
+
+    // priv key template
+    let priv_key_template = vec![Attribute::Token(true), Attribute::Sign(true)];
+
+    // generate a key pair
+    let (public, private) =
+        session.generate_key_pair(&mechanism, &pub_key_template, &priv_key_template)?;
+
+    let private_cloned = unsafe { ObjectHandle::new_from_raw(private.handle()) };
+    let public_cloned = unsafe { ObjectHandle::new_from_raw(public.handle()) };
+
+    // data to sign
+    let data = [0xFF, 0x55, 0xDD];
+
+    // sign something with it
+    let signature = session.sign(&Mechanism::RsaPkcs, private_cloned, &data)?;
+
+    // verify the signature
+    session.verify(&Mechanism::RsaPkcs, public_cloned, &data, &signature)?;
+
+    // delete keys
+    session.destroy_object(public)?;
+    session.destroy_object(private)?;
+
+    Ok(())
+}


hug-dev (approved):
Looks nice! Thank you!

Was wondering if you could add a small test using those if that would make sense?

cryptoki/tests/vendor_extensions.rs
```diff
@@ -0,0 +1,168 @@
...
+    arg2: *mut CK_MECHANISM,
+    arg3: CK_OBJECT_HANDLE,
+    arg4: *mut CK_ATTRIBUTE,
+    arg5: CK_ULONG,
+    arg6: *mut CK_OBJECT_HANDLE,
+) -> CK_RV;
+
+fn load_from_library(library_path: String) -> Result<(libloading::Library, C_DeriveKey)> {
+    let library = unsafe {
+        libloading::os::unix::Library::open(Some(library_path), RTLD_NOW | RTLD_GLOBAL)
```
hug-dev: I guess here you will need some `cfg` to make it work on Windows as well to make the test pass :) Would it be possible to use 
```
let library = ::libloading::Library::new(path)?;
```
as we do in `cryptoki-sys` [bindings](https://github.com/parallaxsecond/rust-cryptoki/blob/3ac27669d0b6320cdc7dd6a721f0ed50c1e9df7b/cryptoki-sys/src/bindings/generic.rs#L4839)?

jhshannon17: Ah.

I don't seem to be able to use `Library::new` when trying to load the library in two separate places, I've only had success loading eagerly.

I added a separate load for windows, hopefully that is sufficient, would that cover all cases? I'm not particularly familiar with working cross-platform.

cryptoki/tests/vendor_extensions.rs
```diff
@@ -0,0 +1,179 @@
+// Copyright 2021 Contributors to the Parsec project.
```
Jakuje: nit: 2025

cryptoki/tests/vendor_extensions.rs
Jakuje: Why the `RTLD_GLOBAL`? We recently hit an issue with the dlopening with RTLD_GLOBAL (default on mac) in OpenSC (OpenSC/OpenSC#3487), which went into recursion when the pkcs11-spy was used so I am wondering if this is intentional. If not, I think the RTLD_LOCAL should be used to prevent resolving the symbols from the pkcs11 modules in the subsequent loadings.

cryptoki/tests/vendor_extensions.rs
```diff
@@ -0,0 +1,179 @@
...
+
+#[cfg(target_os = "windows")]
+fn load_library(library_path: String) -> libloading::Library {
+    unsafe { ::libloading::Library::new(library_path).unwrap().into() }
+}
+
+fn load_from_library(library_path: String) -> Result<(libloading::Library, C_DeriveKey)> {
+    let library = load_library(library_path);
+
+    let C_DeriveKey = unsafe { library.get(b"C_DeriveKey\0").map(|sym| *sym) }?;
```
Jakuje: nit: I like the c-strings more than explicit null bytes in rust:
```suggestion
    let C_DeriveKey = unsafe { library.get(c"C_DeriveKey").map(|sym| *sym) }?;
```

cryptoki/tests/vendor_extensions.rs
Jakuje: sounds like the windows is still not happy in the CI

wiktor-k:
Folks, is there a reason _getting_ the raw handle is considered unsafe?

As far as I saw in Rust, usually it's the constructor functions that are `unsafe` (since the compiler cannot check the input for validity, e.g. [`CStr::from_bytes_with_nul_unchecked`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_with_nul_unchecked)) but the accessor functions are *not* `unsafe` since it's assumed that the check for validity has already been done elsewhere (e.g. [`Cstr::as_ptr`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.as_ptr)).

(As a side-note, quite a lot of structures already have inner accessors via `AsRef<INNER_TYPE`> and friends).

Thanks for your input!

cryptoki/tests/vendor_extensions.rs
jhshannon17: I used `RTLD_GLOBAL` to be able to load the symbols in a second place from the already loaded symbols instead of trying to load the library twice, but for them to be discoverable they (afaik) have to be loaded globally the first time (and based on testing eagerly).

So the referenced call has to occur first and actually loads the library into global, then when I initialize `Pkcs11` I use `new_from_self` which loads with `dlopen(NULL, ..)`, pulling from already loaded symbols instead of reloading the lib.

I was wary of loading the lib twice, even if both `RTLD_LOCAL`,  I don't want to assume that's okay across implementations.

Jakuje: True. For tests this is probably ok. Thanks for clarification!

cryptoki/tests/vendor_extensions.rs
```diff
@@ -0,0 +1,179 @@
...
+use cryptoki::error::{Result, Rv};
+use cryptoki::mechanism::Mechanism;
+use cryptoki::object::{Attribute, AttributeType, KeyType, ObjectClass, ObjectHandle};
+use cryptoki::session::{Session, UserType};
+use cryptoki::types::AuthPin;
+use cryptoki_sys::{
+    CK_ATTRIBUTE, CK_MECHANISM, CK_MECHANISM_PTR, CK_OBJECT_HANDLE, CK_RV, CK_SESSION_HANDLE,
+    CK_ULONG,
+};
+use libloading::os::unix::{RTLD_GLOBAL, RTLD_NOW};
```
hug-dev: This line should also be under `cfg` for the CI to feel good

jhshannon17:
> > Are there any other feelings for the ObjectHandle `new`?
> 
> `new_from_raw` is good 👌 I wonder if we could/should replace all calls to `new` within the library to `new_from_raw` because that's basically what it does... not sure how that change would propagate.
> 
> FWIW I'm okay with what's in here already, just wondering if there are more "low hanging fruit" to be harvested ;)

Happy to harvest whatever is helpful - there are a number of places where `new` is currently being used that would introduce new unsafe blocks (mechanism/hkdf.rs, mechanism/kbkdf.rs) is that fine?

The other places it's used are just outside unsafe blocks so small to change.

cryptoki/tests/vendor_extensions.rs
jhshannon17: No problem!

hug-dev (approved):
Thank you for going through the comments! For me it's good as is. We can change all calls from `new_from_raw` to `new` in a later PR for all handle types.

The CI is failing, maybe https://github.com/parallaxsecond/rust-cryptoki/pull/318 should be merged before?

hug-dev:
You can try now!

jhshannon17:
pushed!

Jakuje (approved):

PR #318, opened 2025-10-16
Run tests against fips kryoptic current main

Jakuje:
The last released version is missing some fixes for several Simple KDFs.

### DIFF
--- a/.github/workflows/kryoptic-fips.yml
+++ b/.github/workflows/kryoptic-fips.yml
@@ -82,12 +82,12 @@ jobs:
       ### kryoptic build ###
       ######################
       - name: Setup kryoptic
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/kryoptic-fips.yml
```diff
@@ -83,11 +83,9 @@ jobs:
       ######################
       - name: Setup kryoptic
         run: |
-          KRYOPTIC_REVISION="v1.3.1"
```
hug-dev, line 86: Could we use a specific commit number as revision? I am afraid that tracking a branch we might fail the CI if a bug is introduced in `kryoptic` main (I hope not :heart: )

Jakuje: While we run a CI too I understand it can be annoying to see unexpected breakages. Changed to the current commit in main.

PR #322, opened 2025-11-15
Accelerate fetching attributes from an object handle

keldonin:
Hello team!

this PR is all about optimizing calls when fetching attributes from a (potentially network-based, slow) token. The design goal is to minimize the number of calls to the PKCS#11 layer, using the following strategies:
 - regroup all attributes within one template instead of making individual calls to `C_GetAttributeValue`
 - for those attributes with a known size, pre-allocate buffer
 - for the others, and only if needed, regroup them in a secondary template, and fetch them all at once
 - Once done, build up resulting array.

I have tried this patch on a PKCS#11 scanning utility. In some cases, the number of invocations to PKCS#11 APIs is an order of magnitude below the original version, I found it a massive improvement in terms of performance and execution time.

A small benchmarking tool has been added to this PR - the intent is to give the ability to test it against different tokens and see how it goes. It will be removed at a later stage. The former `get_attributes()` has been temporarily renamed `get_attributes_old()` to allow for the benchmarking.

## Results with SoftHSM
Note: SoftHSM is likely to be a worst-case scenario, as fetching attributes is extremely fast on a software-based token. Playing the benchmarking tool against other equipments (smart cards, HSMs) will yield better results.

Different situations have been tested below:
- multiple: several, mixed attributes are requested
- single-fixed: a single attribute which size is well known ( e.g. `CKA_ENCRYPT`)
- single-variable: a single attribute with a variable size (e.g. `CKA_MODULUS`)
- single-nonexist: a single attribute that does not exists for the object

The single-variable case is typically on parity ( the stats varies from run to run between 0.95 and 1.05, which range is within standard deviation). All other cases provide better performance.

```
╔═══════════════════════════════════════════════════════════════════════════════════════════════════╗
║                                  BENCHMARK SUMMARY TABLE                                          ║
╠═══════════════════╦═════════════╦═════════════╦═════════════╦═════════════╦═══════╦═══════════════╣
║     Test Case     ║   Orig Mean ║    Orig p95 ║    Opt Mean ║     Opt p95 ║ Unit  ║    Speedup    ║
╠═══════════════════╬═════════════╬═════════════╬═════════════╬═════════════╬═══════╬═══════════════╣
║ Multiple          ║     1585.61 ║     2867.11 ║      257.49 ║      409.27 ║    µs ║        x 6.16 ║
║ Single-fixed      ║      281.74 ║      800.74 ║      123.31 ║      160.49 ║    µs ║        x 2.28 ║
║ Single-variable   ║      243.94 ║      308.98 ║      245.94 ║      358.56 ║    µs ║        x 0.99 ║
║ Single-nonexist   ║      279.43 ║      762.66 ║      128.42 ║      159.12 ║    µs ║        x 2.18 ║
╚═══════════════════╩═════════════╩═════════════╩═════════════╩═════════════╩═══════╩═══════════════╝
```

### Benchmarking and Developer Tooling

* Added a new example file `benchmark_attributes.rs` that benchmarks and compares the performance and correctness of the original (`get_attributes_old`) and optimized (`get_attributes`) attribute retrieval implementations. It reports statistics and speedups for various attribute types and scenarios.

### Attribute Retrieval Optimization

* Implemented a new `Session::get_attributes` method that optimizes attribute retrieval by pre-allocating buffers for attributes with known fixed sizes and minimizing the number of PKCS#11 calls, while maintaining correctness and filtering out unavailable attributes.
* Renamed the old attribute retrieval method to `get_attributes_old` for benchmarking and comparison purposes.

### Utility and Type Improvements

* Added the `AttributeType::fixed_size` method to determine if an attribute type has a known fixed size, supporting buffer pre-allocation and optimization in the new retrieval method.
* Minor code improvements such as consistent use of `c_void` and import adjustments to support the new implementation. [[1]](diffhunk://#diff-43b4fab1121a336ac8c581abc283016f03197be35029d6a2a8a79f9533e7e3edR12) [[2]](diffhunk://#diff-43b4fab1121a336ac8c581abc283016f03197be35029d6a2a8a79f9533e7e3edL528-R529)

### DIFF
--- a/cryptoki/src/object.rs
+++ b/cryptoki/src/object.rs
@@ -2086,8 +2086,6 @@ pub enum AttributeInfo {
     Sensitive,
     /// The attribute is available to get from the object and has the specified size in bytes.
     Available(usize),
-    /// The attribute exists but has no value
-    NoValue,
     /// The attribute is not available.
     Unavailable,
 }
--- a/cryptoki/src/session/object_management.rs
+++ b/cryptoki/src/session/object_management.rs
@@ -454,11 +454,13 @@ impl Session {
                     template.len().try_into()?,
                 ))
             } {
-                Rv::Ok => match template[0].ulValueLen {
-                    CK_UNAVAILABLE_INFORMATION => results.push(AttributeInfo::Unavailable),
-                    0 => results.push(AttributeInfo::NoValue),
-                    len => results.push(AttributeInfo::Available(len.try_into()?)),
-                },
+                Rv::Ok => {
+                    if template[0].ulValueLen == CK_UNAVAILABLE_INFORMATION {
+                        results.push(AttributeInfo::Unavailable)
+                    } else {
+                        results.push(AttributeInfo::Available(template[0].ulValueLen.try_into()?))
+                    }
+                }
                 Rv::Error(RvError::AttributeSensitive) => results.push(AttributeInfo::Sensitive),
                 Rv::Error(RvError::AttributeTypeInvalid) => {
                     results.push(AttributeInfo::TypeInvalid)
@@ -511,13 +513,10 @@ impl Session {
             .iter()
             .zip(attributes.iter())
             .filter_map(|(attr_info, attr_type)| {
-                match attr_info {
-                    // Regular case, we allocate a buffer of the right size
-                    AttributeInfo::Available(size) => Some((*attr_type, vec![0; *size])),
-                    // No value case, we allocate an empty buffer
-                    AttributeInfo::NoValue => Some((*attr_type, vec![])),
-                    // all other cases are skipped
-                    _ => None,
+                if let AttributeInfo::Available(size) = attr_info {
+                    Some((*attr_type, vec![0; *size]))
+                } else {
+                    None
                 }
             })
             .collect();
@@ -527,7 +526,7 @@ impl Session {
             .map(|(attr_type, memory)| {
                 Ok(CK_ATTRIBUTE {
                     type_: (*attr_type).into(),
-                    pValue: as_cptr!(memory),
+                    pValue: memory.as_ptr() as *mut std::ffi::c_void,
                     ulValueLen: memory.len().try_into()?,
                 })
             })
@@ -535,16 +534,14 @@ impl Session {
 
         // This should only return OK as all attributes asked should be
         // available. Concurrency problem?
-        if !template.is_empty() {
-            unsafe {
-                Rv::from(get_pkcs11!(self.client(), C_GetAttributeValue)(
-                    self.handle(),
-                    object.handle(),
-                    template.as_mut_ptr(),
-                    template.len().try_into()?,
-                ))
-                .into_result(Function::GetAttributeValue)?;
-            }
+        unsafe {
+            Rv::from(get_pkcs11!(self.client(), C_GetAttributeValue)(
+                self.handle(),
+                object.handle(),
+                template.as_mut_ptr(),
+                template.len().try_into()?,
+            ))
+            .into_result(Function::GetAttributeValue)?;
         }
 
         // Convert from CK_ATTRIBUTE to Attribute


cryptoki/examples/benchmark_attributes.rs
```diff
@@ -0,0 +1,332 @@
...
+        AttributeType::Label,    // Variable length
+        AttributeType::Id,       // Variable length
+        AttributeType::KeyType,  // CK_ULONG (fixed, 8 bytes)
+        AttributeType::Token,    // CK_BBOOL as CK_ULONG (fixed, 8 bytes)
+        AttributeType::Private,  // CK_BBOOL as CK_ULONG (fixed, 8 bytes)
+        AttributeType::EcPoint,  // Variable length (~65 bytes for P-256 uncompressed)
+        AttributeType::EcParams, // Variable length (10 bytes for P-256 OID)
+        AttributeType::Verify,   // CK_BBOOL as CK_ULONG (fixed, 8 bytes)
+        AttributeType::Encrypt,  // CK_BBOOL as CK_ULONG (fixed, 8 bytes)
+        AttributeType::Local,    // CK_BBOOL as CK_ULONG (fixed, 8 bytes)
```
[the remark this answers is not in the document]
keldonin: good catch.

cryptoki/src/session/object_management.rs
```diff
@@ -547,6 +548,183 @@ impl Session {
...
+        for (i, attr) in template1.iter().enumerate() {
+            if attr.ulValueLen == CK_UNAVAILABLE_INFORMATION {
+                // Skip unavailable attributes
+                continue;
+            } else if attr.pValue.is_null() && attr.ulValueLen > 0 {
+                // NULL pointer but has a length - needs fetching in pass2
+                pass2_indices.push(i);
+            } else if !attr.pValue.is_null() {
+                // Has data already - satisfied in pass1
+                pass1_indices.push(i);
```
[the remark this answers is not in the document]
keldonin: This would normally not be expected, but better safe than sorry.

cryptoki/src/session/object_management.rs
```diff
@@ -547,6 +548,179 @@ impl Session {
...
+        let mut pass1_indices: Vec<usize> = Vec::new();
+        let mut pass2_indices: Vec<usize> = Vec::new();
+
+        for (i, attr) in template1.iter().enumerate() {
+            if attr.ulValueLen == CK_UNAVAILABLE_INFORMATION {
+                // Skip unavailable attributes
+                continue;
+            } else if attr.pValue.is_null() && attr.ulValueLen > 0 {
+                // NULL pointer but has a length - needs fetching in pass2
+                pass2_indices.push(i);
```
Jakuje: I think this will bail out on the test added in #324 where the softhsm will happily return the variable-length AllowedMechanisms with length 0. Theoretically, this could happen also with other non-fixed size attributes that might be empty, such as CKA_ID (from brief run through the pkcs11 specs).

cryptoki/src/object.rs
```diff
@@ -322,6 +322,86 @@ impl AttributeType {
...
+            | AttributeType::ValidationLevel
+            | AttributeType::ValidationAuthorityType
+            | AttributeType::ProfileId
+            | AttributeType::KeyGenMechanism => Some(size_of::<CK_ULONG>()),
+
+            // CK_DATE (8 bytes: year[4] + month[2] + day[2])
+            AttributeType::StartDate | AttributeType::EndDate => Some(size_of::<CK_DATE>()),
+
+            // CK_VERSION (2 bytes: major + minor)
+            AttributeType::ValidationVersion => Some(size_of::<CK_VERSION>()),
```
Jakuje, line 417: The `ValidationCountry` says it should be `2 letter ISO country code` so lets include it as fixed-length  attribute too.

cryptoki/src/session/object_management.rs
Jakuje: As I expected, the test from #324 fails with this PR + SoftHSM so it needs some more love:
```
---- import_export stdout ----

thread 'import_export' (2292734) panicked at cryptoki/tests/basic.rs:1120:26:
removal index (is 0) should be < len (is 0)
```

cryptoki/src/session/object_management.rs
```diff
@@ -547,6 +548,179 @@ impl Session {
...
+
+        match rv1 {
+            Rv::Ok
+            | Rv::Error(RvError::BufferTooSmall)
+            | Rv::Error(RvError::AttributeSensitive)
+            | Rv::Error(RvError::AttributeTypeInvalid) => {
+                // acceptable - we'll inspect ulValueLen/pValue
+            }
+            _ => {
+                rv1.into_result(Function::GetAttributeValue)?;
```
Jakuje: The `rv1` is not used anywhere after this block so I would rather keep using `rv` instead of `rv1` and `rv2` as it makes it easier to read and follow.

cryptoki/src/object.rs
keldonin: sure thing. included.

cryptoki/src/session/object_management.rs
keldonin: adjusted.

cryptoki/src/session/object_management.rs
keldonin: This is an interesting problem indeed. I have removed the check on length; I have also ensured that when converting from C to Rust, i.e. `TryFrom<CK_ATTRIBUTE> for Attribute` is able to capture a zero length attribute properly, and adapted the Rust to C conversion to transform the dangling but non-null ptr of Rust into a real NULL for the C layer (that needs it this way).

All tests pass now.

cryptoki/src/session/object_management.rs
```diff
@@ -500,7 +499,7 @@ impl Session {
     /// Get the attributes values of an object.
     /// Ignore the unavailable one. One has to call the get_attribute_info method to check which
     /// ones are unavailable.
-    pub fn get_attributes(
+    pub fn get_attributes_old(
```
Jakuje, line 504: I think we should somehow mark this as `pub(crate)` as this is likely not something we want to expose to our users, don't we?
Or do we want to keep it around as a "backup" in the API?

Jakuje (dismissed):

cryptoki/src/session/object_management.rs
keldonin: TBH I was meaning to remove it entirely. I just left it there to be able to demonstrate the gain in perf. Certainly not something you want to keep/maintain over time. I assume that with proper git history at hand, you don't really need to maintain dead code anymore. 

Note that it also means I will remove the benchmark comparison code as well.

Let me know if you are fine with the approach, and I'll do the cleanup.

Jakuje: That would probably make the most sense. While the benchmark is still interesting to see for comparision, it will be less important in the future. So maybe it might be the best to have this PR as a demonstration with a benchmark and then separate PR with the cleaned up version only that would be ready for merge, while still having this branch/PR for others to chime in and consult the speed difference or run the benchmarks?

Note that other maintainers might have different ideas, but this would be my approach that should not hurt.

keldonin:
I've created another branch that I'll use as a new PR - a cleaned up version of this one, without the benchmarking example and the `get_attributes_old` method.

I have also reverted a few changes that I mistakenly committed in my previous rebase. These improve the performance of the previous implementation, but for one specific case only. This explains also why the last perf comparison table I posted in this PR looks different, as there was apparently no gain for "Single-Nonexist". 

If desired, I can create a separate PR for what has been reverted, but the changes are less useful with the new method IMO.


This is the latest perf table for this PR, to use as the reference. It compares the perf with the original implementation.

```
╔════════════════════════════════════════════════════════════════════════════════════════════════════╗
║                                   BENCHMARK SUMMARY TABLE                                          ║
╠════════════════════╦═════════════╦═════════════╦═════════════╦═════════════╦═══════╦═══════════════╣
║     Test Case      ║   Orig Mean ║    Orig p95 ║    Opt Mean ║     Opt p95 ║ Unit  ║    Speedup    ║
╠════════════════════╬═════════════╬═════════════╬═════════════╬═════════════╬═══════╬═══════════════╣
║ Multiple           ║       29.02 ║       34.21 ║        6.21 ║        7.62 ║    µs ║        x 4.68 ║
║ Single-fixed       ║        4.57 ║        5.46 ║        2.17 ║        2.62 ║    µs ║        x 2.10 ║
║ Single-variable    ║        4.86 ║        5.75 ║        4.61 ║        5.83 ║    µs ║        x 1.05 ║
║ Single-nonexist    ║        4.60 ║        5.46 ║        2.28 ║        2.75 ║    µs ║        x 2.01 ║
║ Single-zero-length ║        4.58 ║        5.46 ║        4.82 ║        5.67 ║    µs ║        x 0.95 ║
╚════════════════════╩═════════════╩═════════════╩═════════════╩═════════════╩═══════╩═══════════════╝
```

I guess we can just "close" the PR and proceed with the new one.

keldonin:
Retiring this PR, #341 is taking over. The branch remains, to provide easy access to banchmarking example.

Issue #323, opened 2025-11-18
Undefined behavior in CK_ATTRIBUTE::try_from or Session::get_attributes


nwalfield:
While testing `sequoia-cryptoki` using SoftHSM2, I encountered a panic:

```
unsafe precondition(s) violated: slice::from_raw_parts requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`
```

The error message suggests that there is undefined behavior.  Examining the backtrace, [`CK_ATTRIBUTE::try_from`](https://github.com/parallaxsecond/rust-cryptoki/blob/fc0eae6/cryptoki/src/object.rs#L955), was executing which is called from [`Session::get_attributes`](https://github.com/parallaxsecond/rust-cryptoki/blob/fc0eae6/cryptoki/src/session/object_management.rs#L547), which is immediately preceded by an unsafe block:

```
        // Allocating a chunk of memory where to put the attributes value.
        let attrs_memory: Vec<(AttributeType, Vec<u8>)> = attrs_info
            .iter()
            .zip(attributes.iter())
            .filter_map(|(attr_info, attr_type)| {
                if let AttributeInfo::Available(size) = attr_info {
                    Some((*attr_type, vec![0; *size]))
                } else {
                    None
                }
            })
            .collect();

        let mut template: Vec<CK_ATTRIBUTE> = attrs_memory
            .iter()
            .map(|(attr_type, memory)| {
                Ok(CK_ATTRIBUTE {
                    type_: (*attr_type).into(),
                    pValue: memory.as_ptr() as *mut std::ffi::c_void,
                    ulValueLen: memory.len().try_into()?,
                })
            })
            .collect::<Result<Vec<CK_ATTRIBUTE>>>()?;

        // This should only return OK as all attributes asked should be
        // available. Concurrency problem?
        unsafe {
            Rv::from(get_pkcs11!(self.client(), C_GetAttributeValue)(
                self.handle(),
                object.handle(),
                template.as_mut_ptr(),
                template.len().try_into()?,
            ))
            .into_result(Function::GetAttributeValue)?;
        }

        // Convert from CK_ATTRIBUTE to Attribute
        template.into_iter().map(|attr| attr.try_into()).collect()
```

In my code, the call to `Session::get_attributes` is [here](https://gitlab.com/sequoia-pgp/sequoia-cryptoki/-/blob/main/tools/src/commands/token/list.rs?ref_type=heads#L685).

[...]

PR #324, opened 2025-11-20
Fix handling empty AllowedMechanisms attribute

Jakuje:
As returned by SoftHSM: https://github.com/softhsm/SoftHSMv2/issues/825
    
Fixes: #323

### DIFF
--- a/cryptoki/tests/basic.rs
+++ b/cryptoki/tests/basic.rs
@@ -1117,6 +1117,20 @@ fn import_export() -> TestResult {
         panic!("Expected the Modulus attribute.");
     }
 
+    let mut attrs =
+        session.get_attributes(is_it_the_public_key, &[AttributeType::AllowedMechanisms])?;
+
+    if is_softhsm() {
+        let attr = attrs.remove(0);
+        if let Attribute::AllowedMechanisms(v) = attr {
+            assert_eq!(v, Vec::<MechanismType>::new());
+        } else {
+            panic!("Expected the AllowedMechanisms attribute.");
+        }
+    } else {
+        assert_eq!(attrs, Vec::<Attribute>::new());
+    }
+
     // delete key
     session.destroy_object(is_it_the_public_key)?;
 


cryptoki/src/object.rs
```diff
@@ -1114,6 +1114,9 @@ impl TryFrom<CK_ATTRIBUTE> for Attribute {
                 Ok(Attribute::ValidationVersion(Version::new(val[0], val[1])))
             }
             AttributeType::AllowedMechanisms => {
+                if attribute.ulValueLen == 0 {
+                    return Ok(Attribute::AllowedMechanisms(Vec::<MechanismType>::new()));
+                }
```
nwalfield, line 1135: This looks correct to me.  I don't know what style rules this project has, but the surrounding code doesn't use return.  So more idiomatic would be to do:

```
                if attribute.ulValueLen == 0 {
                    Ok(Attribute::AllowedMechanisms(Vec::<MechanismType>::new()))
                } else {
                   ...
```

Jakuje: Thank you for having a look into this! I wanted to avoid nesting the rest of the code one more level. But you are right avoiding return would be more systematic with the surrounding code. Will adjust.

Jakuje:
OTOH this could be considered a softhsm bug (or at least oddness) and as a rust-cryptoki, we might want to shield a users from such pkcs11-module-specific issues. But then we would have to filter this out early in get_attributes, which does not sound great either.

nwalfield:
You already pointed out that at least two attributes can definitely be zero length.  It thus seems like a bug in `rust-cryptoki` that it tries to allocate `0` bytes and expects to get a pointer that can be cast to a raw pointer and passed to [`std::slice::from_raw_parts`](https://doc.rust-lang.org/stable/std/slice/fn.from_raw_parts.html):

```
Behavior is undefined if any of the following conditions are violated:

    data must be non-null, valid for reads for len * size_of::<T>() many bytes, and it must be properly aligned
```

cryptoki/src/object.rs
```diff
@@ -1114,18 +1114,22 @@ impl TryFrom<CK_ATTRIBUTE> for Attribute {
...
-                    )
-                };
-                let types: Vec<MechanismType> = val
-                    .iter()
-                    .copied()
-                    .map(|t| t.try_into())
-                    .collect::<Result<Vec<MechanismType>>>()?;
-                Ok(Attribute::AllowedMechanisms(types))
+                if attribute.ulValueLen == 0 {
+                    Ok(Attribute::AllowedMechanisms(Vec::<MechanismType>::new()))
```
neverpanic: Should this come with a comment explaining the special case and a reference to the softhsm ticket, or do you want to rely on the commit message for that?

Jakuje: Technically, we could add here a link if anyone would prefer. But I find it quite self-explanatory to understand what is going on there.

Jakuje: Added the comment here.

cryptoki/src/object.rs
```diff
@@ -1114,18 +1114,25 @@ impl TryFrom<CK_ATTRIBUTE> for Attribute {
...
-                };
-                let types: Vec<MechanismType> = val
-                    .iter()
-                    .copied()
-                    .map(|t| t.try_into())
-                    .collect::<Result<Vec<MechanismType>>>()?;
-                Ok(Attribute::AllowedMechanisms(types))
+                if attribute.ulValueLen == 0 {
+                    /* For zero-length attributes we are getting pointer
+                     * to static buffer of length zero, which can not be to create slices.
```
neverpanic:
```suggestion
                     * to static buffer of length zero, which can not be used to create slices.
```

Jakuje: :facepalm: thanks. Fixed.

hug-dev (approved):
Thanks for the fix and the comment ⭐

cryptoki/src/object.rs
```diff
@@ -1114,18 +1114,25 @@ impl TryFrom<CK_ATTRIBUTE> for Attribute {
...
-                    .iter()
-                    .copied()
-                    .map(|t| t.try_into())
-                    .collect::<Result<Vec<MechanismType>>>()?;
-                Ok(Attribute::AllowedMechanisms(types))
+                if attribute.ulValueLen == 0 {
+                    /* For zero-length attributes we are getting pointer to static
+                     * buffer of length zero, which can not be used to create slices.
+                     * Short-circuit here to avoid crash (#324) */
+                    Ok(Attribute::AllowedMechanisms(Vec::<MechanismType>::new()))
```
wiktor-k: I think the compiler would automatically infer the type here:

```suggestion
                    Ok(Attribute::AllowedMechanisms(Vec::new()))
```

cryptoki/src/object.rs
```diff
@@ -1114,18 +1114,25 @@ impl TryFrom<CK_ATTRIBUTE> for Attribute {
...
+                        std::slice::from_raw_parts(
+                            attribute.pValue as *const CK_MECHANISM_TYPE,
+                            attribute.ulValueLen.try_into()?,
+                        )
+                    };
+                    let types: Vec<MechanismType> = val
+                        .iter()
+                        .copied()
+                        .map(|t| t.try_into())
+                        .collect::<Result<Vec<MechanismType>>>()?;
```
wiktor-k:
```suggestion
                    let types = val
                        .iter()
                        .copied()
                        .map(|t| t.try_into())
                        .collect::<Result<Vec<_>>>()?;
```

cryptoki/src/object.rs
Jakuje: You are obviously right! Adjusted. Thanks for the review!

cryptoki/src/object.rs
Jakuje: Same here. Adjusted. Thanks for the review!

Jakuje:
oh, no. Something in the ubuntu broke the installation of the 32b arch softhsm ...

PR #325, opened 2025-11-20
Don't raise an error at close when it's expected

EliseChouleur:
If session's close raises a `SessionHandleInvalid` error, that just means the session is already closed by other means.
So don't raise an error but just a warn message in this case.

### DIFF
--- a/cryptoki/src/session/session_management.rs
+++ b/cryptoki/src/session/session_management.rs
@@ -107,11 +107,11 @@ impl Session {
     // Helper function to be able to close a session only taking a reference.
     // This is used in the Drop trait function which only takes a reference as input.
     pub(super) fn close_inner(&self) -> Result<()> {
-        self.closed.set(true);
-
         unsafe {
             Rv::from(get_pkcs11!(self.client(), C_CloseSession)(self.handle()))
-                .into_result(Function::CloseSession)
+                .into_result(Function::CloseSession)?;
         }
+        self.closed.set(true);
+        Ok(())
     }
 }


wiktor-k:
> If not, it could be good to still have this message to show that something unexpected happened?

Yep, it sounds as if there would be a double close in this scenario, which would be good to have in logs.

I wonder about the use case of other means of closing the session... :eyes:

cryptoki/examples/thread_local_session.rs
```diff
@@ -0,0 +1,295 @@
...
+    // Initialize user PIN
+    let user_pin = AuthPin::new(USER_PIN.into());
+    {
+        let session = pkcs11.open_rw_session(slot)?;
+        session.login(UserType::So, Some(&so_pin))?;
+        session.init_pin(&user_pin)?;
+    } // Session auto-closes here via Drop
+
+    // Store context in global Mutex
+    *PKCS11_CTX.lock().unwrap() = Some(pkcs11);
```
wiktor-k: I wonder if this would work:

```suggestion
    *PKCS11_CTX.lock()? = Some(pkcs11);
```

cryptoki/examples/thread_local_session.rs
```diff
@@ -0,0 +1,295 @@
...
+    })
+}
+
+/// Generate an RSA key pair and sign data using thread-local session.
+/// Demonstrates that each thread has its own session and can perform
+/// cryptographic operations independently.
+///
+/// This function makes multiple calls to with_session() to demonstrate
+/// session reuse within the same thread.
+fn generate_and_sign(thread_id: usize) -> testresult::TestResult<()> {
```
wiktor-k: `()` should be the default:

```suggestion
fn generate_and_sign(thread_id: usize) -> testresult::TestResult {
```

cryptoki/examples/thread_local_session.rs
```diff
@@ -0,0 +1,295 @@
...
+    println!(
+        "Thread {:?} (worker {}): All operations completed",
+        thread::current().id(),
+        thread_id
+    );
+
+    Ok(())
+}
+
+fn main() -> testresult::TestResult {
```
wiktor-k: I think we could import the test result and just reuse it everywhere:

```suggestion
use testresult::TestResult;

fn main() -> TestResult {
```

hug-dev:
We reverted a commit on main recently that might affect you! How is everything working now on your side?

cryptoki/examples/thread_local_session.rs
```diff
@@ -0,0 +1,294 @@
...
+use cryptoki::mechanism::Mechanism;
+use cryptoki::object::Attribute;
+use cryptoki::session::{Session, UserType};
+use cryptoki::types::AuthPin;
+
+const USER_PIN: &str = "fedcba";
+const SO_PIN: &str = "abcdef";
+
+// Global PKCS11 context shared across all threads using Arc for cheap cloning
+static PKCS11_CTX: OnceLock<Arc<Pkcs11>> = OnceLock::new();
```
hug-dev: We moved back with `Pkcs11` containing an `Arc` so maybe it's not needed to have another Arc here :)

cryptoki/src/session/session_management.rs
```diff
@@ -10,14 +10,17 @@ use crate::types::{AuthPin, RawAuthPin};
...
 use secrecy::ExposeSecret;
 use std::convert::{TryFrom, TryInto};
 
 impl Drop for Session {
     fn drop(&mut self) {
         match self.close_inner() {
             Err(Error::Pkcs11(RvError::SessionClosed, Function::CloseSession)) => (), // the session has already been closed: ignore.
+            Err(Error::Pkcs11(RvError::SessionHandleInvalid, Function::CloseSession)) => {
+                warn!("Failed to close session: The specified session handle is invalid, the session must be already closed.");
+            }
```
hug-dev: With the recent changes you can also directly call `close` on your session and filter on the error returned on your side!

EliseChouleur: This isn't enough because Session::close() takes ownership of self but still triggers Drop after execution. The Drop then calls close_inner() again, which fails and logs the ERROR.
But, I can use std::mem::forget() when I detect an invalid handle to prevent Drop from running and avoide the error log.

EliseChouleur: I've created a test in this PR to help :)

hug-dev:
Thanks for putting the effort in helping us understanding this problem with an example 👏 

I also think that option 2 is good in your list but can also suggest removing the `Drop` implementation on `Session` altogether, the same way as we did for `finalize` on `Pkcs11` in #326. That way there is only one close and no side effect or extra logic needed.

Sure, it will add a few extra steps needed for the creators of session but in the end they might not create that many (and have a long-lived one as you did)? They can also define their own `Session` with `Drop` if they want. Maybe it's best to leave that kind of logic to the user.

What do you think?

hug-dev:
Maybe instead of a Cargo feature we can modify the API to be able to choose if we want the `Session` to close on `Drop` or not? That way both are available without changing build flags :)

To not add breaking changes we could add new constructors, `open_ro(w)_session_no_drop`, that would not close on `Drop` and modify the default constructors so that they do?

> And I also add the close security to not call close twice ?

Yes exactly!

cryptoki/examples/thread_local_session.rs
```diff
@@ -0,0 +1,289 @@
...
+    println!("- Automatic session lifecycle management");
+    println!("- Session reuse within the same thread\n");
+
+    // Initialize global context once
+    println!("Initializing PKCS11 context...");
+    init_pkcs11_context()?;
+    println!();
+
+    // Spawn multiple threads
+    println!("Spawning 3 worker threads...\n");
```
wiktor-k: let max_threads = 3;

```suggestion
    println!("Spawning {max_threads} worker threads...\n");
```

cryptoki/examples/thread_local_session.rs
```diff
@@ -0,0 +1,289 @@
...
+    let mut handles = vec![];
+
+    for i in 0..3 {
```
wiktor-k:
```suggestion
    for i in 0..max_threads {
```

cryptoki/src/context/session_management.rs
```diff
@@ -13,7 +13,12 @@ use super::Function;
 
 impl Pkcs11 {
     #[inline(always)]
-    fn open_session(&self, slot_id: Slot, read_write: bool) -> Result<Session> {
+    fn open_session(
+        &self,
+        slot_id: Slot,
+        read_write: bool,
+        close_on_drop: bool,
```
wiktor-k: I'm not super OK with having so many booleans as then the call sites are looking really ugly. It's not such a super-big deal since the function is private but maybe it'd be worth using an enum?

```rust
#[derive(Debug, Copy, Clone)]
enum CloseOnDrop {
  AutomaticallyCloseSession,
  DoNotClose,
}
```

Of course my naming isn't that great :sweat_smile: 

```suggestion
        close_on_drop: CloseOnDrop,
```

cryptoki/src/context/session_management.rs
```diff
@@ -63,13 +68,31 @@ impl Pkcs11 {
     /// # let _ = session; Ok(()) }
     /// ```
     pub fn open_ro_session(&self, slot_id: Slot) -> Result<Session> {
-        self.open_session(slot_id, false)
+        self.open_session(slot_id, false, true)
```
wiktor-k:
```suggestion
        self.open_session(slot_id, false, CloseOnDrop::AutomaticallyCloseSession)
```

cryptoki/src/session/session_management.rs
```diff
@@ -103,6 +107,14 @@ impl Session {
     // Helper function to be able to close a session only taking a reference.
     // This is used in the Drop trait function which only takes a reference as input.
     pub(super) fn close_inner(&self) -> Result<()> {
+        if self.closed.get() {
```
wiktor-k: :thinking: I don't think this is strictly needed because then it'd call the `C_CloseSession` down below and get the error from there... :thinking:

cryptoki/tests/common/test_logger.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use log::{Level, LevelFilter, Metadata, Record};
+use std::sync::Mutex;
+
+static LOG_MESSAGES: Mutex<Vec<(Level, String)>> = Mutex::new(Vec::new());
+static LOGGER: TestLogger = TestLogger;
+
+struct TestLogger;
+
+impl log::Log for TestLogger {
```
wiktor-k: Hmm... what's the reason for needing the test logger?

I wonder if we could reuse this thing: https://crates.io/crates/test-log

cryptoki/tests/drop_error_handling.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    // Simulate token removal
+    mock.simulate_token_removal();
+
+    // Drop the session WITHOUT calling close()
+    // This should trigger the Drop error
+    drop(session);
+
+    // Verify that an error WAS logged
+    println!("Captured logs:");
```
wiktor-k: Aaah, that's why the logger is here... to verify the test :thinking:

EliseChouleur: [crates.io/crates/test-log](https://crates.io/crates/test-log) crate doesn't seem to do someting similar
But I agree with you that checking the logs isn't the best way to do tests. but that was how I found out about the double close. I'll check for a better way to test.

cryptoki/src/session/mod.rs
```diff
@@ -40,6 +41,8 @@ pub struct Session {
     client: Pkcs11,
     // This is not used but to prevent Session to automatically implement Sync
     _guard: PhantomData<*mut u32>,
+    close_on_drop: bool,
+    closed: Cell<bool>,
```
hug-dev, line 51: Do you *need* a `Cell` here 🤔 ? And if so wondering if using an atomic boolean is "lighter"?

cryptoki/src/session/session_management.rs
hug-dev: Agree! Also I am thinking that we should maybe set `closed` to `true` if the call to `CloseSession` later on succeeded?

cryptoki/tests/drop_error_handling.rs
```diff
@@ -0,0 +1,180 @@
+// Copyright 2024 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+//! Tests for Drop error handling in Session.
+//!
+//! These tests use a mock PKCS#11 library that can simulate token removal,
```
hug-dev: Really appreciate the effort you put into adding a whole mock PKCS11 library to be able to test this but I am wondering if it might not be a bit "too much" for the scope of this change. I am thinking about the tradeoff between testing this and adding a lot of code that we will have to maintain in the future 😅 

Would it be possible somehow to get `drop` to fail with something else than "already close" without the need of the entire mock library?

If not maybe in this case we can say that it's been "tested on your machine" 😅

cryptoki/src/session/mod.rs
```diff
@@ -65,11 +68,13 @@ impl std::fmt::UpperHex for Session {
 unsafe impl Send for Session {}
 
 impl Session {
-    pub(crate) fn new(handle: CK_SESSION_HANDLE, client: Pkcs11) -> Self {
+    pub(crate) fn new(handle: CK_SESSION_HANDLE, client: Pkcs11, close_on_drop: bool) -> Self {
```
hug-dev: If this function was public, `unsafe` and renamed `new_from_raw` (ie similar than what we did with the [`ObjectHandle`](https://docs.rs/cryptoki/latest/cryptoki/object/struct.ObjectHandle.html#method.new_from_raw)), then it would be possible in the tests to create new bad session and see that `drop` fails with `CKR_SESSION_HANDLE_INVALID` in the logs

cryptoki/tests/drop_error_handling.rs
EliseChouleur: In addition to be able to test this case it was also to suggest a mock version of tests to be able to test edge cases later on, but I'm not sure there's really some other edge cases to test ^^

cryptoki/src/session/mod.rs
EliseChouleur: Actually `Cell` is more performant as `AtomicBool` uses memory barriers/fences to ensure thread-safe access but `Session` isn't `Sync`. I think the difference is negligible but using Cell supports the not Sync idea.

hug-dev: Alright thanks for the explanation! Good to know and I learned a thing :)

cryptoki/tests/drop_error_handling.rs
hug-dev: I see, then I think we could maybe do a simple test triggering a fault with a `new_from_raw` method as described below and maybe revisit this mock-PKCS11 idea once we have another use case?

EliseChouleur: Yup, I've done this in 38aaf5afa88783f759403e5b9e8cd22009a16c69 but couldn't find a way to test without checking the logs which isn't a nice test assertion, so I think you're right, it's better to remove.

cryptoki/src/session/session_management.rs
```diff
@@ -103,6 +107,8 @@ impl Session {
     // Helper function to be able to close a session only taking a reference.
     // This is used in the Drop trait function which only takes a reference as input.
     pub(super) fn close_inner(&self) -> Result<()> {
+        self.closed.set(true);
```
wiktor-k: @hug-dev did you mean that the `set` should be positioned after the call to `C_CloseSession`?

Or maybe this one should be done in `drop()` when the `close_inner()` function returns `Ok(())`?

:thinking:

cryptoki/src/session/mod.rs
```diff
@@ -28,6 +29,12 @@ pub use object_management::ObjectHandleIterator;
 pub use session_info::{SessionInfo, SessionState};
 pub use validation::ValidationFlagsType;
 
+#[derive(Debug, PartialEq)]
```
wiktor-k: I think we can also make it `Copy` since it's super small :thinking: 

```suggestion
#[derive(Debug, PartialEq, Eq, Clone, Copy)]
```

wiktor-k (commented):
I think it looks small and good :ok_hand:

cryptoki/src/session/session_management.rs
hug-dev: > @hug-dev did you mean that the set should be positioned after the call to C_CloseSession?

Yes! In case `C_CloseSession` fails and we want to try it again later.

> Or maybe this one should be done in drop() when the close_inner() function returns Ok(())?

I think it's good to have it here since ` pub fn close(self)` also calls `close_inner`

EliseChouleur: I moved the set to true, but I think you can't call twice the `close` function as it's taking the ownership of `Session`, so the only retry will be the drop.

hug-dev: ahh yes good point! At least we have one more chance :)

PR #326, opened 2025-11-24
Refactor initialing/finalizing Pkcs11 and Session structures

### LINKED ISSUE #208: Do not call C_Finalize if not initialized

hug-dev:
As discussed in #208 and related PRs, the lifecycle management of the Pkcs11 structure might probably happen outside of the Rust program and such automatic finalizing on Drop is not wanted.

This PR removes that and also allows manually closing a session, which is needed as sessions take reference of the Pkcs11 structure.
Also removes the double initialization check and adapt the tests to work.

Fix #208

### DIFF
--- a/cryptoki/tests/slh_dsa.rs
+++ b/cryptoki/tests/slh_dsa.rs
@@ -22,6 +22,7 @@ fn slh_dsa() -> TestResult {
     if !pkcs11.is_fn_supported(Function::VerifySignature) {
         /* return Ignore(); */
         println!("SKIP: The PKCS#11 module does not support VerifySignature API");
+        pkcs11.finalize()?;
         return Ok(());
     }
 
@@ -109,6 +110,9 @@ fn slh_dsa() -> TestResult {
     session.destroy_object(public)?;
     session.destroy_object(private)?;
 
+    session.close()?;
+    pkcs11.finalize()?;
+
     Ok(())
 }
 
@@ -120,6 +124,7 @@ fn slh_dsa_multipart() -> TestResult {
     if !pkcs11.is_fn_supported(Function::VerifySignature) {
         /* return Ignore(); */
         println!("SKIP: The PKCS#11 module does not support VerifySignature API");
+        pkcs11.finalize()?;
         return Ok(());
     }
 
@@ -174,6 +179,9 @@ fn slh_dsa_multipart() -> TestResult {
     session.destroy_object(public)?;
     session.destroy_object(private)?;
 
+    session.close()?;
+    pkcs11.finalize()?;
+
     Ok(())
 }
 
@@ -185,6 +193,7 @@ fn slh_dsa_hash() -> TestResult {
     if !pkcs11.is_fn_supported(Function::VerifySignature) {
         /* return Ignore(); */
         println!("SKIP: The PKCS#11 module does not support VerifySignature API");
+        pkcs11.finalize()?;
         return Ok(());
     }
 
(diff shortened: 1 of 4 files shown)


cryptoki/tests/slh_dsa.rs
```diff
@@ -174,6 +177,9 @@ fn slh_dsa_multipart() -> TestResult {
     session.destroy_object(public)?;
     session.destroy_object(private)?;
 
+    session.close()?;
+    pkcs11.finalize()?;
```
wiktor-k, line 183: Correct me if I'm wrong but this change means that if an error happens *before* these lines are hit then no close and finalize will happen. Of course that's only outside of the happy path and maybe not so important (since tests are assumed to always complete successfully) :thinking:

hug-dev: True 😅 I was too lazy to add finalize on all the paths just for the tests

wiktor-k: Naah, I'm just thinking other folks may also be too lazy or just use the code here as an inspiration for their code... I'm not sure how big of a deal would be if there was no finalization on error cases.

(My original PR idea was to have something that finalizes, just like the old code, to have all options available :thinking: )

hug-dev: I think as per the discussion in https://github.com/parallaxsecond/rust-cryptoki/issues/208, seems like initialization/finalization of Pkcs11 mostly happen outside of the application that uses it... So not sure if it's something we should care about that much. But we can also add a new constructor which could finalize on Drop. Maybe on a new PR :nerd_face: ?

hug-dev:
> Do you want to open an issue or should I do that?

I did with https://github.com/latchset/kryoptic/issues/375!

cryptoki/tests/slh_dsa.rs
simo5: While it is necessary to call Initialize, calling Finalize is not such a big deal; when the process terminates, generally, all is gone anyway.

Any pkcs11 driver that needs to keep state has to face the chance the process crashes anyway, so any state saved, or client/server architecture within a pkcs11 driver needs to be able to handle the case where finalize is not called and unclean shutdowns happened.

Jakuje:
Opened #328 for testing against kryoptic main.

hug-dev:
Rebased on `main`

Jakuje:
So this still fails in random tests with kryoptic during the initialization here:

https://github.com/parallaxsecond/rust-cryptoki/blob/979e6314b81813d5c6956616967c1e7cea1854dd/cryptoki/tests/common/mod.rs#L48

```
thread 'object_handle_new_from_raw' (3766) panicked at cryptoki/tests/common/mod.rs:48:51:
called `Result::unwrap()` on an `Err` value: Pkcs11(CryptokiAlreadyInitialized, Initialize)
```

This means the kryoptic is already initialized (in different thread). Should we ignore this error in here now, given that its not fatal?

hug-dev:
> This means the kryoptic is already initialized (in different thread).

Agree! Since all tests are tagged with `serial` and I made sure in this PR that all tests finalize the Pkcs11 structure, I would not expect to see those errrors. But I forgot a few cheeky tests. Hopefully everything is covered now :pray:

Issue #311, opened 2025-09-12
Run tests against kryoptic main


Jakuje:
Currently, we run tests against the Fedora packaged version of kryoptic, which is some months old:

https://src.fedoraproject.org/rpms/kryoptic

We would like to get the new version released, but the code is moving fast and it might be more viable to build the kryoptic from source for the CI as we are hitting the new functionalities (#307, #310, #306) we are unable to test.

Jakuje:
We are getting a new kryoptic release so it migt not be needed. And I do not expect that much of a development from now on.

PR #328, opened 2025-11-26
Run tests against kryoptic main

Jakuje:
Follow-up from #311 and discussion in #326, where we agreed that keeping track of main is less problematic than keeping outdated version.

### DIFF
--- a/.github/workflows/kryoptic.yml
+++ b/.github/workflows/kryoptic.yml
@@ -1,11 +1,14 @@
 ---
-name: Test kryoptic FIPS module
+name: Test kryoptic
 
 on: [push, pull_request, workflow_dispatch]
 
+env:
+  KRYOPTIC_REVISION: main
+
 jobs:
-  build:
-    name: Test kryoptic FIPS module
+  tests-kryoptic:
+    name: Run tests against Kryoptic
     runs-on: ubuntu-22.04
     container: quay.io/fedora/fedora:latest
     steps:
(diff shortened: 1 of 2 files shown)


hug-dev (approved):
Thank you for the quick fix!

.github/workflows/kryoptic.yml
```diff
@@ -132,5 +210,5 @@ jobs:
           TEST_PKCS11_MODULE: /__w/rust-cryptoki/kryoptic/target/debug/libkryoptic_pkcs11.so
           OUT_DIR: /__w/rust-cryptoki/kryoptic/target/debug/deps/
           RUST_BACKTRACE: 1
-        run: cargo test
+        run: cargo build --all-features && cargo test
```
wiktor-k, line 213: I think `check` is enough, since we don't care about the output artifacts:

```suggestion
        run: cargo check --all-features && cargo test
```

The question is whether we want to check `--all-targets` (which include tests and examples and bins...)...

wiktor-k (approved):
I think this looks good :ok_hand:

.github/workflows/kryoptic.yml
Jakuje: I took this from `ci.yml` (where I probably wrote this previously). I think that just `test` should be enough. Not sure if we should need the `check` at all here as the `check` is executed in the other jobs (as well as the `--all-targets` -- the Fedora `cargo` should not be that much different from ubuntu cargo.

wiktor-k: Good point :+1:

Issue #321, opened 2025-10-27
Add a way to pass a pointer for the "reserved" argument in CInitializeArgs()


simo5:
Although the spec says the pReserved pointer is reserved, there are some tokens that make use (and sort-of depend on being able to get an initialization string through that interface.

Softokn from the NSS project and also Kryoptic can be initialized that way, and although they both can use also environment variables to pass the same info, it is often easier (especially in testing environments) to just pass a string straight to the C_Initialize call.

PR #329, opened 2025-11-27
Add pReserved and pkcs11 flags

placintaalexandru:
#321: Add `pReserved` to `CInitializeArgs` and other flags as as stated in PKCS#11 initialize doc

### DIFF
--- a/cryptoki/src/context/locking.rs
+++ b/cryptoki/src/context/locking.rs
@@ -99,7 +99,7 @@ impl From<UnlockError> for CK_RV {
 }
 
 /// Trait to manage lifecycle of mutex objects
-pub trait MutexLifeCycle {
+pub trait MutexLifeCycle: Send + Sync {
     /// Creates a mutex
     fn create() -> Result<Box<Self>, CreateError>;
 
--- a/cryptoki/src/context/locking.rs
+++ b/cryptoki/src/context/locking.rs
@@ -2,24 +2,142 @@
 // SPDX-License-Identifier: Apache-2.0
 //! Locking related type
 
-use cryptoki_sys::{CKF_OS_LOCKING_OK, CK_FLAGS};
+use cryptoki_sys::{CKF_LIBRARY_CANT_CREATE_OS_THREADS, CKF_OS_LOCKING_OK, CK_FLAGS, CK_RV};
 
-use std::ptr;
+use std::{
+    os::raw::c_void,
+    ptr::{self, NonNull},
+};
 
-/// Argument for the initialize function
+/// Function pointer that creates a mutex
+pub type CreateMutexFn = unsafe extern "C" fn(*mut *mut ::std::os::raw::c_void) -> CK_RV;
+
+/// Function pointer that destroys a mutex
+pub type DestroyMutexFn = unsafe extern "C" fn(*mut ::std::os::raw::c_void) -> CK_RV;
+
+/// Function pointer that locks a mutex
+pub type LockMutexFn = unsafe extern "C" fn(*mut ::std::os::raw::c_void) -> CK_RV;
+
+/// Function pointer that unlocks a mutex
+pub type UnlockMutexFn = unsafe extern "C" fn(*mut ::std::os::raw::c_void) -> CK_RV;
+
+/// Provides function pointers for mutex-handling to ensure safe multi-threaded access.
 #[derive(Copy, Clone, Debug)]
-pub enum CInitializeArgs {
-    /// The library can use the native OS library for locking
+pub struct CustomMutexHandling {
+    create_mutex: CreateMutexFn,
+    destroy_mutex: DestroyMutexFn,
+    lock_mutex: LockMutexFn,
+    unlock_mutex: UnlockMutexFn,
+}
+
+impl CustomMutexHandling {
+    /// Create a new `CustomMutexHandling` with the given function pointers
+    /// to handle library's thread safety.
+    ///
+    /// # Safety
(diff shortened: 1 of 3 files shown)


cryptoki/src/context/locking.rs
```diff
@@ -2,32 +2,169 @@
...
+            destroy_mutex: destroy_mutex,
+            lock_mutex: lock_mutex,
+            unlock_mutex: unlock_mutex,
+        }
+    }
+}
+
+/// Flags to set for the initialize function
+#[derive(Copy, Clone, Debug)]
+pub enum CInitializeFlags {
```
hug-dev: I think to start with a new abstraction of this type it would be nice to go with a "dumber" 1-1 mapping with the C type. I think it's nicer to start simple and straightforward and then improve over time depending on usage.

I mean the following:
* making `CInitializeArgs` a struct with the same fields as in the standard but no need to have `pReserved` for now (it can be set to null always in the `From` implementation).
* The `CInitializeFlags` flags could be created as a `bitflags` definition, similar than what is done for `MechanismInfoFlags`. That way it can expend easily with new flags in future versions.
* The `CInitializeArgs` basic constructor would take the five arguments as inputs (`Option` for the mutex functions)

cryptoki/src/context/locking.rs
```diff
@@ -2,32 +2,169 @@
...
 
-use std::ptr;
+use std::{
+    os::raw::c_void,
+    ptr::{self, NonNull},
+};
 
-/// Argument for the initialize function
+/// Function pointer that creates a mutex
+pub type CreateMutexFn = unsafe extern "C" fn(*mut *mut ::std::os::raw::c_void) -> CK_RV;
```
hug-dev: I think it would be nice to also create abstraction types for those functions, which would be safe abstractions over the `CK_LOCKMUTEX`, `CK_UNLOCKMUTEX`, `CK_DESTROYMUTEX` and `CK_CREATEMUTEX` types. They would take in and return abstracted types and be converted to the C function in our code.

hug-dev (commented):
Thanks for the PR! As said with the comments I think we don't have to continue with the enum type as it is now but start from scratch with a more straightforward implementation :)

cryptoki/src/context/locking.rs
placintaalexandru: yes, I agree with that. the thing is I do not know how they would look like (I've never had to pass those functions when i initialize cryptoki)

cryptoki/src/context/locking.rs
placintaalexandru: I agree with the 1st point, but the purpose of the PR was to add the `pReserved` pointer. I see 2 ways here:
- allowing the pointer and using `unsafe`
- restraining the user to some known cryptokis that use `pReserved` and the type of those structs. We would create safe bindings to those structs. The pointer would be then obtained via `Box::into_raw(Box::new(safe_object))` and we would have to manage that pointer

```rust
pub enum Reserved {
    None,
    IBM(IBMInitParams),
    Utimaco(UtimacoInitParams),
}

let p_reserved = Box::into_raw(Box::new(Reserved::IBM(...)));
```

Regarding points 2 and 3 I see the following problems:
- for point 2: one could set both `CKF_OS_LOCKING_OK` and `CKF_LIBRARY_CANT_CREATE_OS_THREADS`. Is that a valid situation?
- for point 3: all 4 mutex functions must be provided according to PKCS#11 docs. Allowing `Option` would add unnecessary runtime validation so that all match `Some`

cryptoki/src/context/locking.rs
hug-dev: Those are defined [here](https://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html#_Toc416959692) but I could not come up with a nice way to do this is pure Rust.
I was thinking of defining a `trait Mutex` with all those methods but not sure how one could define trait methods as types :thinking: 

I am wondering if to not make it easier for now we could ignore those functions? As in always set them to `None` and remove it as parameters. Until someone needs them and then we can think about how to do it?

Unless you come up with a good way to do this.

cryptoki/src/context/locking.rs
hug-dev: > but the purpose of the PR was to add the pReserved pointer.

Yes :facepalm: ! I forgot the whole purpose of the related issue :grimacing: 

> for point 2: one could set both CKF_OS_LOCKING_OK and CKF_LIBRARY_CANT_CREATE_OS_THREADS. Is that a valid situation?

I think the PKCS11 API does not prevent it so I guess we should not limit either... But agree that some combinations of those two are pretty weird.

> for point 3: all 4 mutex functions must be provided according to PKCS#11 docs. Allowing Option would add unnecessary runtime validation that can be avoided

Let's see what we decide with the other comment to see how we proceed here

placintaalexandru: They mention that if `CKF_LIBRARY_CANT_CREATE_OS_THREADS` is set and the lib can't work with it then `C_Initialize` will return with code `CKR_NEED_TO_CREATE_THREADS`. So I guess I can go with `bitflags`

cryptoki/src/context/locking.rs
placintaalexandru: I have a first approach that compiles. The downside is that it exposes an API that resembles the pthread mutex API

hug-dev (commented):
Thank you!! The bitflags and pReserved changes look good to me!
I wanted to ask you if we could start with only those changes in the PR (without the genericity and the mutex functions) and open a new one for the rest?

PR #330, opened 2025-12-03
Add initialize flags and pReserved to CInitializeArgs

placintaalexandru:
#321

### DIFF
--- a/cryptoki/examples/generate_key_pair.rs
+++ b/cryptoki/examples/generate_key_pair.rs
@@ -1,6 +1,6 @@
 // Copyright 2024 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
-use cryptoki::context::{CInitializeArgs, Pkcs11};
+use cryptoki::context::{CInitializeArgs, CInitializeFlags, Pkcs11};
 use cryptoki::mechanism::Mechanism;
 use cryptoki::object::Attribute;
 use cryptoki::session::UserType;
@@ -19,7 +19,7 @@ fn main() -> testresult::TestResult {
             .unwrap_or_else(|_| "/usr/local/lib/softhsm/libsofthsm2.so".to_string()),
     )?;
 
-    pkcs11.initialize(CInitializeArgs::OsThreads)?;
+    pkcs11.initialize(CInitializeArgs::new(CInitializeFlags::OS_LOCKING_OK))?;
 
     let slot = pkcs11.get_slots_with_token()?[0];
 
--- a/cryptoki/src/context/locking.rs
+++ b/cryptoki/src/context/locking.rs
@@ -2,34 +2,107 @@
 // SPDX-License-Identifier: Apache-2.0
 //! Locking related type
 
-use cryptoki_sys::{CKF_OS_LOCKING_OK, CK_FLAGS};
+use bitflags::bitflags;
+use cryptoki_sys::{self, CK_C_INITIALIZE_ARGS, CK_FLAGS};
 
-use std::ptr;
+use std::{
+    os::raw::c_void,
+    ptr::{self, NonNull},
+};
 
-/// Argument for the initialize function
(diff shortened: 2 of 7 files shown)


cryptoki/examples/generate_key_pair.rs
```diff
@@ -19,7 +19,7 @@ fn main() -> testresult::TestResult {
             .unwrap_or_else(|_| "/usr/local/lib/softhsm/libsofthsm2.so".to_string()),
     )?;
 
-    pkcs11.initialize(CInitializeArgs::OsThreads)?;
+    pkcs11.initialize(CInitializeArgs::new(CInitializeFlags::OS_LOCKING_OK))?;
```
hug-dev, line 22: Nice 👌

cryptoki/src/context/locking.rs
```diff
@@ -2,34 +2,107 @@
...
-impl From<CInitializeArgs> for cryptoki_sys::CK_C_INITIALIZE_ARGS {
+impl CInitializeArgs {
+    /// Create a new `CInitializeArgs` with the given flags
+    ///
+    /// # Examples
+    /// ```
+    /// use cryptoki::context::{CInitializeArgs, CInitializeFlags};
+    ///
+    /// let args = CInitializeArgs::new(CInitializeFlags::OS_LOCKING_OK | CInitializeFlags::LIBRARY_CANT_CREATE_OS_THREADS);
+    /// ```
```
hug-dev, line 39: Great with the examples!

cryptoki/src/context/locking.rs
```diff
@@ -2,34 +2,107 @@
...
+        }
+    }
+
+    /// Create a new `CInitializeArgs` with the given flags and reserved pointer.
+    ///
+    /// # Safety
+    /// Considered unsafe due to the user's ability to pass any pointer: the
+    /// caller must ensure that the provided pointer is valid and points
+    /// to a struct that has the same ABI as the one required by cryptoki's
+    /// `C_Initialize`
```
hug-dev, line 53: Agree with the general safety notice but I guess what the user must do to ensure safety is dependent on the specific PKCS11 implementation. Maybe one of them can see the argument as a value instead of a pointer for example. Maybe we should be a bit more vague here saying that the use of `p_reserved` is implementation-specific and can lead to unsafe behaviour?

cryptoki/src/context/locking.rs
placintaalexandru: thank you

cryptoki/src/context/locking.rs
placintaalexandru: It is implementation specific, but at the end of the day, the pointer passed from Rust must point to something that still has the same ABI as the value expected by `C_Initialize`. From my experience with the crate, most of the time (if not all the time), I ended up passing some struct that contained `#[repr(C)]`

hug-dev: Yes ok makes sense!

cryptoki/src/context/locking.rs
```diff
@@ -2,34 +2,107 @@
...
+    /// // Obtain the raw pointer
+    /// let ptr_reserved = NonNull::new(Box::into_raw(boxed_data) as *mut c_void)
+    ///     .expect("Failed to create NonNull pointer");
+    ///
+    /// // SAFETY: since the data was allocated when boxed_data was created, this is safe
+    /// let args = unsafe {
+    ///     CInitializeArgs::new_with_reserved(flags, ptr_reserved)
+    /// };
+    ///
+    /// // Reassemble back the bax to make sure the data is correctly cleaned
```
Jakuje:
```suggestion
    /// // Reassemble back the box to make sure the data is correctly cleaned
```

hug-dev:
Let's merge this before #326 to avoid you the pain of rebasing, as a first-time contributor Christmas gift :gift:

Jakuje (approved):

PR #331, opened 2025-12-08
Fix CI pipelines

Jakuje:
* The Ubuntu image needs to be updated before installing multilib binutil
* The Ed25519Ctx mechanism was removed from OpenSSL FIPS provider


hug-dev (approved):
I am wondering what kind of black magic you conjured to be able to find the solution to that!! Good job ⭐ !!

Wondering if we should add a comment on the lines in the workflow to indicate it's to speed up the upgrade

PR #332, opened 2025-12-09
Truncate intermediate outputs of multi-part cipher operations

satoqz:
The outputs of `encrypt_update` and `decrypt_update` should be truncated, just like already done with `encrypt`, `encrypt_final`, `decrypt` and `decrypt_final`. The additional AES-GCM test section I've added would previously fail with SoftHSM if the decrypted output was not truncated correctly.

### DIFF
--- a/cryptoki/tests/basic.rs
+++ b/cryptoki/tests/basic.rs
@@ -481,7 +481,7 @@ fn encrypt_decrypt_multipart() -> TestResult {
         0x33, 0x99, 0x77,
     ];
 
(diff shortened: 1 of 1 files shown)


wiktor-k:
> Would you prefer I retract the additional test / comment it out / move it into a separate test function that's skipped?

I think there are helper functions so you can add a check and return if we're using softhsm.

But please add a comment with a link to kryoptic issue so that this can be fixed upstream. Also CC @Jakuje 😅

PR #333, opened 2025-12-11
Make `Session` hold a reference to a Pkcs11 or an owned Pkcs11

nwalfield:
Making `Session` take a reference to a `Pkcs11` object means that the `Session`'s lifetime is explicitly coupled to the `Pkcs11` object.  To allow users of the library to decouple these, extend `Session` to hold either a reference to a `Pkcs11` or an `Arc<Pkcs11>`.

A user of the library chooses to use an `Arc<Pkcs11>` by using `Session::open_owned_ro_session` and `Session::open_owned_rw_session` instead of `Session::open_ro_session` and `Session::open_rw_session`. No other parallel APIs are needed.

This effective allows opting into the old behavior that was removed in this commit: https://github.com/parallaxsecond/rust-cryptoki/commit/1994987175423b5508a0a2b75ff01d55fef348eb (part of https://github.com/parallaxsecond/rust-cryptoki/pull/326).

I need these semantics in my project, as my PKCS11 abstraction is separate from Sessions and tying them together would add quite a bit of complexity with no obvious benefits as far as I can see.

I'm unsure of where to put the `MaybeOwned` data structure.  I put it in the same module as the `Pkcs11` data structure.  We should discuss if it belongs there.  Note: we can't use `Cow`, as `Pkcs11` does not implement `Clone`.

### DIFF
--- a/cryptoki/src/context/session_management.rs
+++ b/cryptoki/src/context/session_management.rs
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //! Session management functions
 
-use cryptoki_sys::{CKF_RW_SESSION, CKF_SERIAL_SESSION};
+use cryptoki_sys::{CKF_RW_SESSION, CKF_SERIAL_SESSION, CK_SESSION_HANDLE};
+use std::sync::Arc;
 
 use crate::context::Pkcs11;
 use crate::error::{Result, Rv};
@@ -13,7 +14,7 @@ use super::Function;
 
 impl Pkcs11 {
     #[inline(always)]
-    fn open_session(&self, slot_id: Slot, read_write: bool) -> Result<Session<'_>> {
+    fn open_session(&self, slot_id: Slot, read_write: bool) -> Result<CK_SESSION_HANDLE> {
         let mut session_handle = 0;
 
         let flags = if read_write {
@@ -33,7 +34,7 @@ impl Pkcs11 {
             .into_result(Function::OpenSession)?;
         }
 
-        Ok(Session::new(session_handle, self))
+        Ok(session_handle)
     }
 
     /// Open a new Read-Only session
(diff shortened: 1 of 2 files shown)


hug-dev:
Hello! Thanks for the PR!

@wiktor-k , do you remember why you added the change in https://github.com/parallaxsecond/rust-cryptoki/commit/1994987175423b5508a0a2b75ff01d55fef348eb? I thought in the beginning that it was indeed cleaner but if it is causing more issues than anticipated I am wondering if you had other reasons to add it :)

wiktor-k:
> @wiktor-k , do you remember why you added the change in [1994987](https://github.com/parallaxsecond/rust-cryptoki/commit/1994987175423b5508a0a2b75ff01d55fef348eb)? I thought in the beginning that it was indeed cleaner but if it is causing more issues than anticipated I am wondering if you had other reasons to add it :)

Hmm... it seems I removed the inner Arc as part of making the client manage the lifecycle of the Pkcs11 object.

Are the issues related to the fact that now one cannot hold a single Pkcs11 object and then open multiple sessions in different threads? If so, we may want to revert that commit.

Sorry for the trouble :wave:

wiktor-k:
Okay, that's good, thanks for the reference.

> the behavior of Cryptoki is undefined if multiple threads of an application attempt to access a common Cryptoki session simultaneously

That's not an issue since Session is explicitly not Send and Sync.

> ... This is true even if the Cryptoki library was initialized with locking enabled for thread-safety.

So if the Cryptoki library is not initialized with locking enabled creating sessions on different threads may not be a good idea?

I'm considering the following scenario: Pkcs11 is used to create one session, then the `Pkcs11` object is moved to a different thread to create another session there. Which shouldn't be possible if if wasn't initialized with locking enabled. And if that's not the case why are we using `Arc` instead of regular `Rc`? :thinking: 

Just throwing questions, see if there's any merit in there...

PR #336, opened 2025-12-16
test: Add Wycheproof-based AES-GCM tests

### LINKED ISSUE #187: Add Wycheproof-based tests
For many crypto operations we could just use [Wycheproof](https://crates.io/crates/wycheproof) to verify that we're doing the right thing. We don't necessarily need to go through all tests, since we're not verifying a crypto implementation, just a wrapping layer, but still good to use official test vectors, and have them maintained for us.

testingapisname:
Implements comprehensive AES-GCM testing using official Wycheproof test vectors from Google. Tests 313 valid cryptographic operations across multiple key sizes (128/192/256-bit), nonce lengths, tag sizes, and AAD configurations.

Fixes #187

### DIFF
--- a/cryptoki/tests/wycheproof.rs
+++ b/cryptoki/tests/wycheproof.rs
@@ -1,4 +1,4 @@
-// Copyright 2024 Contributors to the Parsec project.
+// Copyright 2025 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
 //! Wycheproof-based cryptographic tests
 //!
@@ -268,15 +268,6 @@ fn aes_gcm_message_wycheproof() -> TestResult {
     let session = pkcs11.open_rw_session(slot)?;
     session.login(UserType::User, Some(&AuthPin::new(USER_PIN.into())))?;
 
-    // Determine PKCS#11 version to apply appropriate limits
-    let info = pkcs11.get_library_info()?;
-    let cryptoki_version = info.cryptoki_version();
-    let max_nonce_bytes = if cryptoki_version.major() >= 3 {
-        u32::MAX as usize
-    } else {
-        256
-    };
-
     // Load Wycheproof AES-GCM test vectors
     let test_set = wycheproof::aead::TestSet::load(wycheproof::aead::TestName::AesGcm)?;
 
@@ -294,12 +285,6 @@ fn aes_gcm_message_wycheproof() -> TestResult {
         }
 
         for test in &test_group.tests {
-            // Skip tests with nonce sizes that exceed PKCS#11 version-specific limits
-            if test.nonce.len() > max_nonce_bytes {
-                skipped += 1;
-                continue;
-            }
-
             // Skip tests with tag sizes that exceed PKCS#11 limits (max 128 bits)
             if test.tag.len() * 8 > 128 {
                 skipped += 1;


cryptoki/tests/wycheproof.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        // Only test key sizes we support (128, 192, 256 bits)
+        if ![128, 192, 256].contains(&key_size) {
+            skipped += test_group.tests.len();
+            continue;
+        }
+
+        for test in &test_group.tests {
+            // Skip tests with nonce sizes that exceed PKCS#11 limits (max 256 bytes)
+            if test.nonce.len() > 256 {
```
Jakuje: While the PKCS#11 2.40 has this limitation [1], the higher limit is defined in the current specification 3.2:

> The length of the initialization vector can be any number between 1 and (2^32) - 1.

In the tests, you can detect the pkcs11 version and I think we can use the larger ones for the new modules.

Updating the documentation would be good too

[1] https://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/errata01/os/pkcs11-curr-v2.40-errata01-os-complete.html#_Toc441850509
[2] https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/pkcs11-spec-v3.2.html#_Toc195693377

cryptoki/tests/wycheproof.rs
```diff
@@ -0,0 +1,168 @@
...
+                    Ok(params) => params,
+                    Err(e) => {
+                        eprintln!("Test {}: Failed to create GCM params: {:?}", test.tc_id, e);
+                        failed += 1;
+                        continue;
+                    }
+                };
+
+            // Test encryption
+            let encrypt_result = session.encrypt(&Mechanism::AesGcm(gcm_params), key, &test.pt);
```
Jakuje, line 110: Could we test also the PKCS#11 3 API with message-based encryption?

https://github.com/parallaxsecond/rust-cryptoki/blob/36fb3c3a2a0dc9c625c61329322cd0cabbb9d853/cryptoki/tests/basic.rs#L2587

Jakuje:
> @Jakuje I think I addressed your comments in this [b8ec82d](https://github.com/parallaxsecond/rust-cryptoki/commit/b8ec82d3b1bb07a822b6ca0a7b36c0e1f274233c) commit. I tested with softhsm and kyroptic.

Thank you!

> I am not sure what you mean by update the documentation. Can you please review?

Sorry, I meant this part of the doc text, which is having the maximum IV length:

https://github.com/parallaxsecond/rust-cryptoki/blob/main/cryptoki/src/mechanism/aead.rs#L26

cryptoki/tests/wycheproof.rs
```diff
@@ -0,0 +1,502 @@
...
+    session.login(UserType::User, Some(&AuthPin::new(USER_PIN.into())))?;
+
+    // Determine PKCS#11 version to apply appropriate limits
+    let info = pkcs11.get_library_info()?;
+    let cryptoki_version = info.cryptoki_version();
+    let max_nonce_bytes = if cryptoki_version.major() >= 3 {
+        u32::MAX as usize
+    } else {
+        256
+    };
```
Jakuje: The function `MessageEncryptInit` is available only in the PKCS#11 3 API so I think this check is not needed.

cryptoki/tests/wycheproof.rs
```diff
@@ -0,0 +1,502 @@
+// Copyright 2024 Contributors to the Parsec project.
```
Jakuje: nit: 
```suggestion
// Copyright 2025 Contributors to the Parsec project.
```

PR #341, opened 2026-01-07
Accelerate fetching attributes from an object handle - cleaned up version

keldonin:
As suggested by @Jakuje, this is a cleaned up version of PR #322, with legacy code removed and benchmark example stripped off.

See PR #322 for discussions.

### DIFF
--- a/cryptoki/src/object.rs
+++ b/cryptoki/src/object.rs
@@ -419,8 +419,44 @@ impl AttributeType {
             // CK_VALIDATION_COUNTRY (2 CK_UTF8CHAR, typically 2 bytes for ISO country code)
             AttributeType::ValidationCountry => Some(size_of::<[CK_UTF8CHAR; 2]>()),
 
-            // Variable-length attributes (all the others)
-            _ => None,
+            // Variable-length attributes
+            AttributeType::AcIssuer
+            | AttributeType::AllowedMechanisms
+            | AttributeType::Application
+            | AttributeType::AttrTypes
+            | AttributeType::Base
+            | AttributeType::CheckValue
+            | AttributeType::Coefficient
+            | AttributeType::EcParams
+            | AttributeType::EcPoint
+            | AttributeType::Exponent1
+            | AttributeType::Exponent2
+            | AttributeType::HashOfIssuerPublicKey
+            | AttributeType::HashOfSubjectPublicKey
+            | AttributeType::Id
+            | AttributeType::Issuer
+            | AttributeType::Label
+            | AttributeType::Modulus
+            | AttributeType::ObjectId
+            | AttributeType::Owner
+            | AttributeType::Prime
+            | AttributeType::Prime1
+            | AttributeType::Prime2
+            | AttributeType::PrivateExponent
+            | AttributeType::PublicExponent
+            | AttributeType::PublicKeyInfo
+            | AttributeType::Seed
+            | AttributeType::SerialNumber
+            | AttributeType::Subject
+            | AttributeType::UniqueId
+            | AttributeType::Url
+            | AttributeType::ValidationModuleId
+            | AttributeType::ValidationCertificateIdentifier
+            | AttributeType::ValidationCertificateUri
+            | AttributeType::ValidationVendorUri
+            | AttributeType::ValidationProfile
+            | AttributeType::Value
+            | AttributeType::VendorDefined(_) => None,
         }
     }
 }


cryptoki/tests/basic.rs
```diff
@@ -4617,3 +4617,140 @@ fn object_handle_new_from_raw() -> TestResult {
...
+        Attribute::Sign(true),
+        Attribute::Private(true),
+        Attribute::Id(vec![]),
+    ];
+
+    // generate a key pair
+    let (public, private) =
+        session.generate_key_pair(&mechanism, &pub_key_template, &priv_key_template)?;
+
+    // Test get_attributes_fast with various attribute types
```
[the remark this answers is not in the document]
keldonin: fixed in the next commit.

cryptoki/tests/common/mod.rs
```diff
@@ -17,14 +19,24 @@ fn get_pkcs11_path() -> String {
         .unwrap_or_else(|_| "/usr/local/lib/softhsm/libsofthsm2.so".to_string())
 }
 
+// Used to simulate different library behaviors.
+// for SoftHSM, just create the environment variable TEST_PRETEND_LIBRARY with "softhsm"
```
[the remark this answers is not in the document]
keldonin: Fixed in the next commit.

cryptoki/src/session/object_management.rs
```diff
@@ -497,54 +498,178 @@ impl Session {
...
-            })
-            .collect::<Result<Vec<CK_ATTRIBUTE>>>()?;
-
-        // This should only return OK as all attributes asked should be
-        // available. Concurrency problem?
-        unsafe {
+                    pValue: as_cptr!(buffer),
+                    ulValueLen: size as CK_ULONG,
+                });
+                buffers.push(buffer);
```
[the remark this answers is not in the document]
keldonin: The Copilot analysis is incorrect. 

at line 539:
```rust
                let buffer = vec![0u8; size];
                template1.push(CK_ATTRIBUTE {
                    type_: (*attr_type).into(),
                    pValue: as_cptr!(buffer),
                    ulValueLen: size as CK_ULONG,
                });
                buffers.push(buffer);
```


`buffer` is a `Vec` object, and in this instance, its associated payload is allocated at creation. Then the address of buffer's array is captured using the `as_cptr!()`macro. Eventually, `buffer` is pushed into `buffers`. While it is true that `buffer`is moved into `buffers`, the underlying memory array is unaffected and remains the same, and therefore pValue remains valid.

cryptoki/src/session/object_management.rs
```diff
@@ -497,54 +498,178 @@ impl Session {
...
+
+        for (i, attr) in template1.iter().enumerate() {
+            if attr.ulValueLen == CK_UNAVAILABLE_INFORMATION {
+                // Skip unavailable attributes
+                continue;
+            } else if attr.pValue.is_null() {
+                // NULL pointer - needs fetching in pass2
+                // Note: even if ulValueLen is 0, we still need to fetch as 0 length attributes are legit
+                pass2_indices.push(i);
+            } else if !attr.pValue.is_null() {
```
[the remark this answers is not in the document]
keldonin: good catch.

keldonin:
Sure, will do that once the review process is finished. BTW this can be conveniently performed at merge time, so maybe not needed from my end.

cryptoki/src/object.rs
```diff
@@ -322,6 +340,89 @@ impl AttributeType {
...
+            AttributeType::StartDate | AttributeType::EndDate => Some(size_of::<CK_DATE>()),
+
+            // CK_VERSION (2 bytes: major + minor)
+            AttributeType::ValidationVersion => Some(size_of::<CK_VERSION>()),
+
+            // CK_VALIDATION_COUNTRY (2 CK_UTF8CHAR, typically 2 bytes for ISO country code)
+            AttributeType::ValidationCountry => Some(size_of::<[CK_UTF8CHAR; 2]>()),
+
+            // Variable-length attributes (all the others)
+            _ => None,
```
wiktor-k: :thinking: What happens when someone adds a new `AttributeType` and forgets to add a branch here?

(That is I wonder if we should explicitly enumerate all `None` variants here or would that be a small concern...)

cryptoki/tests/common/mod.rs
```diff
@@ -17,14 +19,24 @@ fn get_pkcs11_path() -> String {
         .unwrap_or_else(|_| "/usr/local/lib/softhsm/libsofthsm2.so".to_string())
 }
 
+// Used to simulate different library behaviors.
+// For SoftHSM, just create the environment variable TEST_PRETEND_LIBRARY with "softhsm"
+// This is used to interface a shim library during testing, while appearing to be using the real library.
+#[allow(dead_code)]
+pub fn get_pretend_library() -> String {
```
wiktor-k, line 26: Sorry, I might be missing context but was that used in CI/test scripts that got removed? :thinking:

cryptoki/src/object.rs
Jakuje: I think it will just go through the slow path as before -- first query the size from the token and then query the value. I agree though that having all of them enumerate would make it more nudging for the people adding new attributes to add them correctly here.

cryptoki/tests/common/mod.rs
Jakuje: This is solving the problem I already had but was lazy to address before. When using wrapper like pkcs11-spy [1], the library detection from filename does not work so this is useful to be able to test the correct feature set against the wrapper which logs you the PKCS#11 communication (or does anything else with this like p11-kit can).

[1] https://github.com/OpenSC/OpenSC/wiki/Using-OpenSC#pkcs-11-spy

wiktor-k: Right, right! I guess it'd be good to document all these inner features somewhere but I don't want to drag this PR any longer than necessary. Thanks for the context! :pray:

cryptoki/src/object.rs
wiktor-k: Ack! Let's see what @keldonin thinks about it, I'm fine either way :+1:

hug-dev: Nice that this is future-proof but agree that listing them all could be good!

cryptoki/src/object.rs
```diff
@@ -925,7 +1026,7 @@ impl Attribute {
             | Attribute::ValidationVendorUri(bytes)
             | Attribute::ValidationProfile(bytes)
             | Attribute::VendorDefined((_, bytes))
-            | Attribute::Id(bytes) => bytes.as_ptr() as *mut c_void,
+            | Attribute::Id(bytes) => as_cptr!(bytes),
```
hug-dev, line 1065: Thank for also updating those 🙏

cryptoki/tests/basic.rs
```diff
@@ -4617,3 +4617,140 @@ fn object_handle_new_from_raw() -> TestResult {
 
     Ok(())
 }
+
+#[test]
+#[serial]
+fn get_attributes_test() -> TestResult {
```
hug-dev, line 4623: Great test, thanks!

cryptoki/src/object.rs
keldonin: I think we can make this, no big deal. Removing the catch-all requires the user to adjust this method.

```python
$ python3 -m this | grep Explicit   
Explicit is better than implicit.
```

keldonin:
hmmm, seems like I am facing some erratic errors during testing - and not always with the same test suite. I can't seem to reproduce these locally.

```
 ---- cryptoki/src/context/session_management.rs - context::session_management::Pkcs11::open_ro_session (line 47) stdout ----
Test executable failed (exit status: 101).

stderr:

thread 'main' (4188) panicked at cryptoki/src/context/session_management.rs:14:1:
error: cryptoki::error::Error - Function::Initialize: PKCS11 error: This value can only be returned by C_Login.  It indicates that the normal user’s PIN has not yet been initialized with C_InitPIN.
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /builddir/build/BUILD/rust-1.92.0-build/rustc-1.92.0-src/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /builddir/build/BUILD/rust-1.92.0-build/rustc-1.92.0-src/library/core/src/panicking.rs:80:14
   2: <testresult::TestError as core::convert::From<T>>::from
   3: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
   4: rust_out::main
   5: core::ops::function::FnOnce::call_once
 ```

These issues seem unrelated to the change itself. Something you guys have already experienced?

PR #342, opened 2026-01-07
docs: update CHANGELOG

hug-dev:
Updates the CHANGELOG and also adds releasing steps in the README to remember in the future!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,69 @@
 # Changelog
 
+## [cryptoki-0.11.0](https://github.com/parallaxsecond/rust-cryptoki/tree/cryptoki-0.11.0) (2025-12-19)
+
+[Full Changelog](https://github.com/parallaxsecond/rust-cryptoki/compare/cryptoki-sys-0.5.0...cryptoki-0.11.0)
+
+## [cryptoki-sys-0.5.0](https://github.com/parallaxsecond/rust-cryptoki/tree/cryptoki-sys-0.5.0) (2025-12-19)
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,69 @@
 # Changelog
 
+## [cryptoki-0.11.0](https://github.com/parallaxsecond/rust-cryptoki/tree/cryptoki-0.11.0) (2025-12-19)
+
+[Full Changelog](https://github.com/parallaxsecond/rust-cryptoki/compare/cryptoki-sys-0.5.0...cryptoki-0.11.0)
```
wiktor-k, line 5: These links look weird and when followed github says that "there's nothing to compare as they are identical" (paraphrasing).

Jakuje: Oh, I mised this. It should probably be like this?
https://github.com/parallaxsecond/rust-cryptoki/compare/cryptoki-sys-0.4.0...cryptoki-sys-0.5.0
or even 
https://github.com/parallaxsecond/rust-cryptoki/compare/cryptoki-0.10.0...cryptoki-0.11.0

wiktor-k: And I missed that the "auto merge" was enabled :sweat_smile: well, no big deal, but if the thing was generated by the tool and requires manual fixes this sounds like a recipe for problems :scream:

hug-dev: Yes the automatic generation tools is not smart enough to know what tags to use for the diff :sweat_smile: It's not really made for repos where there are two crates sharing two sets of tags. If we change it manually here then it will be overwritten next time.

I think the best for the future would be to write our changelogs manually but not sure it's worth the time...

PR #345, opened 2026-01-18
fix: return an error if `C_GetFunctionList` couldn't be found

supervacuus:
When a user provides an incorrect library during construction, the interface fallback on V2 would panic instead of returning  an error: https://github.com/NLnetLabs/kmip2pkcs11/issues/19

This PR 
* adds a new error variant (`MissingSymbol`), and
* checks whether required symbol in the V2-fallback, `C_GetFunctionList`, actually exists in `pkcs11_lib` before calling it

### DIFF
--- a/cryptoki/src/context/mod.rs
+++ b/cryptoki/src/context/mod.rs
@@ -430,7 +430,7 @@ mod test {
 
         // Skip if the library doesn't exist (e.g., cross-compilation without target libs)
         if !Path::new(lib_path).exists() {
-            println!("Skipping the test since {lib_path} does not exist.").
+            println!("Skipping the test since {lib_path} does not exist.");
             return;
         }
 


cryptoki/src/context/mod.rs
```diff
@@ -401,3 +404,45 @@ fn v30tov32(f: cryptoki_sys::CK_FUNCTION_LIST_3_0) -> cryptoki_sys::CK_FUNCTION_
...
+        ))]
+        let lib_path = "/lib/x86_64-linux-gnu/libm.so.6";
+        #[cfg(target_os = "freebsd")]
+        let lib_path = "/lib/libm.so.5";
+        #[cfg(target_os = "windows")]
+        let lib_path = "kernel32.dll";
+
+        // Skip if the library doesn't exist (e.g., cross-compilation without target libs)
+        if !Path::new(lib_path).exists() {
+            return;
```
wiktor-k, line 434: Could you insert this so at least from the test output it's clear that the test has been skipped (rather than passed successfully)?

```suggestion
            println!("Skipping the test since {lib_path} does not exist.").
            return;
```

PR #348, opened 2026-01-22
fix(tests): check if SLH DSA can do key pair gen

hug-dev:
The kryoptic backend configured in FIPS mode does not currently support those.

As suggested in https://github.com/latchset/kryoptic/issues/405


wiktor-k (dismissed):
Looks good :ok_hand: any reason this is a draft?

Issue #346, opened 2026-01-20
Allocation-free operations?


bal-e:
I noticed that [`Session::sign()`](https://docs.rs/cryptoki/latest/cryptoki/session/struct.Session.html#method.sign), and the alternative `sign_init()`/`sign_update()`/`sign_final()` methods, always perform heap allocations (`sign()` and `sign_final()` both return `Vec`). At least for my use case, I already know which cryptographic algorithm is being used, so I know exactly how big the resulting signature should be. I'd like to have non-allocating methods, where I can pass in a buffer of a pre-determined size, just for efficiency.

This concern probably extends to all the other PKCS#11 operations. If this feels like a worthwhile change, I'm happy to put up PRs (at least for a few interesting operations). But I wanted to check in before getting started.

hug-dev:
Closing since #349 was merged! We can open new issues for more allocation-free operations, if needed

PR #349, opened 2026-01-26
Allocation-free operations

bal-e:
Starts addressing #346.

Question for the reviewers:

- What do you think of the added documentation? Is it too much? Or should I add more to the related `sign()` functions for consistency?

- I don't have a strict need to implement this for more operations, but I'm happy to do one or two sets more. Are there any in particular you'd like to see? I'm leaning towards key generation (which I need, but is not performance-sensitive) and encryption (which I guess more people would want to use).

- Would you have any tips for adding these methods to the test suite? It'll have to be a bit algorithm-specific.

### DIFF
--- a/cryptoki/src/session/signing_macing.rs
+++ b/cryptoki/src/session/signing_macing.rs
@@ -3,6 +3,8 @@
 //! Signing and authentication functions
 
 use crate::context::Function;
+#[cfg(doc)]
+use crate::error::RvError;
 use crate::error::{Result, Rv};
 use crate::mechanism::Mechanism;
 use crate::object::ObjectHandle;
@@ -10,8 +12,22 @@ use crate::session::Session;
 use cryptoki_sys::*;
 use std::convert::TryInto;
 
+/// # Generating and Verifying Signatures
+///
+/// Several functions are provided for signing data and verifying signatures.
+/// This includes message authentication codes (MACs). The signed data can be
+/// provided in one-shot and streaming modes.
 impl Session {
-    /// Sign data in single-part
+    /// Sign data in one-shot mode.
+    ///
+    /// `data` should be a byte sequence representing the input message. It will
+    /// be signed using the specified key, and the resulting signature will be
+    /// returned in a `Vec`.
+    ///
+    /// Use [`Self::sign_into()`] if (an upper bound for) the size of the
+    /// signature is known, to avoid the heap allocation of `Vec`. Use
+    /// [`Self::sign_init()`] etc. if the input data is being streamed (i.e. it
+    /// is not all immediately available).
     pub fn sign(&self, mechanism: &Mechanism, key: ObjectHandle, data: &[u8]) -> Result<Vec<u8>> {
         self.sign_init(mechanism, key)?;
         self.sign_single(data)
@@ -26,7 +42,7 @@ impl Session {
     pub fn sign_single(&self, data: &[u8]) -> Result<Vec<u8>> {
         let mut signature_len = 0;
 
-        // Get the output buffer length
+        // Get the output buffer length.
         unsafe {
             Rv::from(get_pkcs11!(self.client(), C_Sign)(
                 self.handle(),
@@ -38,9 +54,10 @@ impl Session {
             .into_result(Function::Sign)?;
         }
 
+        // Allocate the output buffer.
         let mut signature = vec![0; signature_len.try_into()?];
 
-        //TODO: we should add a new error instead of those unwrap!
+        // Perform the actual signing.
         unsafe {
             Rv::from(get_pkcs11!(self.client(), C_Sign)(
                 self.handle(),
@@ -52,9 +69,66 @@ impl Session {
             .into_result(Function::Sign)?;
         }
 
+        // Limit the output buffer to the size of the generated signature.
         signature.truncate(signature_len.try_into()?);
 
         Ok(signature)
+    }
+
+    /// Sign data into the given buffer.
+    ///
+    /// `data` should be a byte sequence representing the input message. It will
+    /// be signed using the specified key, and the resulting signature will be
+    /// written to the buffer `sig`.
+    ///
+    /// `sig` should be large enough to store the signature. The number of
+    /// filled bytes will be returned, such that `sig[0..size]` contains the
+    /// prepared signature. `sig[size..]` might be modified.
+    ///
+    /// Use [`Self::sign()`] if (an upper bound for) the size of the signature
+    /// is not known. Use [`Self::sign_init()`] etc. if the input data is being
+    /// streamed (i.e. it is not all immediately available).
+    ///
+    /// ## Errors
+    ///
+    /// Returns [`RvError::BufferTooSmall`] if the generated signature does not
+    /// fit in `sig`. `sig` might be modified. The size of the actual signature
+    /// is **not** returned. This method should only be used if the caller knows
+    /// an upper bound for the signature size.
+    pub fn sign_into(
+        &self,
+        mechanism: &Mechanism,
+        key: ObjectHandle,
+        data: &[u8],
+        sig: &mut [u8],
+    ) -> Result<usize> {
+        // The size of the signature buffer, into which 'C_Sign' will write the
+        // size of the generated signature.
+        let sig_buf_len = sig.len().try_into()?;
+        let mut sig_len = sig_buf_len;
+
+        // Initialize the signing operation.
+        self.sign_init(mechanism, key)?;
+
+        // Perform the actual signing.
+        unsafe {
+            Rv::from(get_pkcs11!(self.client(), C_Sign)(
+                self.handle(),
+                data.as_ptr() as *mut u8,
+                data.len().try_into()?,
+                sig.as_mut_ptr(),
+                &mut sig_len,
+            ))
+            .into_result(Function::Sign)?;
+        }
+
+        assert!(
+            sig_len <= sig_buf_len,
+            "'C_Sign' succeeded but increased 'sig_len', possibly indicating out-of-bounds accesses"
+        );
+
+        // NOTE: As checked above, 'sig_len <= sig_buf_len <= usize::MAX'.
+        Ok(sig_len as usize)
     }
 
     /// Starts new single-part or multi-part signing operation
@@ -88,12 +162,20 @@ impl Session {
         Ok(())
     }
 
-    /// Finalizes ongoing multi-part signing operation,
-    /// returning the signature
+    /// Complete an ongoing streaming signing operation.
+    ///
+    /// This must be preceded by [`Self::sign_init()`] and zero or more calls
+    /// to [`Self::sign_update()`]. This method will terminate the multi-part
+    /// signing operation. The resulting signature will be returned in a `Vec`.
+    ///
+    /// Use [`Self::sign_final_into()`] if (an upper bound for) the size of
+    /// the signature is known, to avoid the heap allocation of `Vec`. Use
+    /// [`Self::sign()`] if the input data is entirely available in a single
+    /// buffer (i.e. does not have to be streamed).
     pub fn sign_final(&self) -> Result<Vec<u8>> {
         let mut signature_len = 0;
 
-        // Get the output buffer length
+        // Get the output buffer length.
         unsafe {
             Rv::from(get_pkcs11!(self.client(), C_SignFinal)(
                 self.handle(),
@@ -103,8 +185,10 @@ impl Session {
             .into_result(Function::SignFinal)?;
         }
 
+        // Allocate the output buffer.
         let mut signature = vec![0; signature_len.try_into()?];
 
+        // Perform the actual signing.
         unsafe {
             Rv::from(get_pkcs11!(self.client(), C_SignFinal)(
                 self.handle(),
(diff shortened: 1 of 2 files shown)


cryptoki/src/session/signing_macing.rs
```diff
@@ -56,6 +61,81 @@ impl Session {
         Ok(signature)
     }
 
+    /// Sign data into the given buffer.
+    ///
+    /// `data` should be a byte sequence representing the input message. It will
+    /// be signed using the specified key, and the resulting signature will be
+    /// written to the given output buffer.
```
hug-dev: maybe add the name of that output buffer to mirror how you did it with `data`, ie `sig`.

cryptoki/src/session/signing_macing.rs
```diff
@@ -56,6 +61,81 @@ impl Session {
...
+                // All reasonable buffer sizes fit in 'usize'.
+                let sig_len: usize = sig_len.try_into()?;
+
+                Err(Error::IncorrectBufferSize(sig_len))
+            }
+
+            // The error seems unrelated to the buffer size, so we don't check
+            // what 'sig_len' has been set to (if anything).
+            Rv::Error(err) => Err(Error::Pkcs11(err, Function::Sign)),
+        }
```
hug-dev: My personal opinion is that we can simply return return `res` as is, without the new error `IncorrectBufferSize`. I think the original PKCS11 error code `BufferTooSmall` is explicit enough for the caller 😸 Same for the multi steps variant!

If needed, we could also add a new method which would return the size of a signature with this specific key and mechanism!

hug-dev (commented):
Thank you!
Really like the added documentation, I think that's good! My only suggestion is to simplify and return the original error as is :)

For testing I think it's fine trying  one or a few variants. You can for example use the same one which is done in the test currently for the `sign` operation!

cryptoki/src/session/signing_macing.rs
bal-e: Right, I had originally added `IncorrectBufferSize` because I wanted to enforce that the caller had provided exactly the right size (and not one that was too big). Since that is no longer the case, the error is not necessary. Good catch, I'll remove it.

cryptoki/src/error/mod.rs
```diff
@@ -31,6 +31,12 @@ pub enum Error {
     /// Error when converting a slice to an array
     TryFromSlice(std::array::TryFromSliceError),
 
+    /// A user-provided buffer has the wrong size.
+    ///
+    /// The buffer may be too small or too large. The expected size (in bytes)
+    /// is returned.
+    IncorrectBufferSize(usize),
```
wiktor-k: I think I'd make it a bit more explicit:

```suggestion
    IncorrectBufferSize {
       expected_size: usize
    },
```

cryptoki/src/session/signing_macing.rs
```diff
@@ -3,13 +3,18 @@
...
 use crate::context::Function;
-use crate::error::{Result, Rv};
+use crate::error::{Error, Result, Rv, RvError};
 use crate::mechanism::Mechanism;
 use crate::object::ObjectHandle;
 use crate::session::Session;
 use cryptoki_sys::*;
 use std::convert::TryInto;
 
+/// # Generating and Verifying Signatures
```
wiktor-k, line 15: Hmm... are these docs rendered anywhere? If not I'd move the docs to the bottom of docs for `struct Session`

cryptoki/src/session/signing_macing.rs
```diff
@@ -56,6 +61,81 @@ impl Session {
...
+    /// `data` should be a byte sequence representing the input message. It will
+    /// be signed using the specified key, and the resulting signature will be
+    /// written to the given output buffer.
+    ///
+    /// The output buffer should be large enough to store the signature.
+    /// If it is large enough, the number of filled bytes is returned
+    /// (i.e. `sig[0..size]` contains the prepared signature). Otherwise,
+    /// [`Error::IncorrectBufferSize`] is returned.
+    ///
+    /// Use [`Self::sign()`] if (an upper bound for) the size of the signature
```
wiktor-k, line 88: I'm not sure if these parens are working 🤔 


```suggestion
    /// Use [`Self::sign`] if (an upper bound for) the size of the signature
```

cryptoki/src/session/signing_macing.rs
```diff
@@ -56,6 +61,81 @@ impl Session {
...
+                data.len().try_into()?,
+                sig.as_mut_ptr(),
+                &mut sig_len,
+            ))
+        };
+
+        // Check the result and buffer size.
+        match res {
+            Rv::Ok => {
+                // TODO: Should this be an Error of some kind?
```
wiktor-k: I don't think so, we promise in the docs to write only up to `sig_len` size so the assert is fine here. I'd add a little bit of docs to the assert...

cryptoki/src/session/signing_macing.rs
```diff
@@ -56,6 +61,81 @@ impl Session {
...
+                assert!(sig_len <= sig_buf_len);
```
wiktor-k: A random suggestion:

```suggestion
                assert!(sig_len <= sig_buf_len, "Signature size is bigger than the buffer size but the PKCS#11 module returned OK, this is not right.");
```

cryptoki/src/session/signing_macing.rs
```diff
@@ -56,6 +61,81 @@ impl Session {
...
+
+                // 'sig_len <= sig_buf_len <= usize::MAX'.
+                Ok(sig_len as usize)
+            }
+
+            Rv::Error(RvError::BufferTooSmall) => {
+                // TODO: Should this be an Error of some kind?
```
wiktor-k: Similarly to the case above with some clever wording :)

wiktor-k (commented):
Looks great, a couple of nitpicks/clarifications. Thanks for your contribution! 🙇

cryptoki/src/session/signing_macing.rs
```diff
@@ -56,6 +61,81 @@ impl Session {
...
+    /// Sign data into the given buffer.
+    ///
+    /// `data` should be a byte sequence representing the input message. It will
+    /// be signed using the specified key, and the resulting signature will be
+    /// written to the given output buffer.
+    ///
+    /// The output buffer should be large enough to store the signature.
+    /// If it is large enough, the number of filled bytes is returned
+    /// (i.e. `sig[0..size]` contains the prepared signature). Otherwise,
+    /// [`Error::IncorrectBufferSize`] is returned.
```
Jakuje: might make sense to be explicit here too that the returned `IncorrectBufferSize` has the expected size.

cryptoki/src/session/signing_macing.rs
```diff
@@ -118,6 +198,66 @@ impl Session {
...
+    ///
+    /// Use [`Self::sign_final()`] if (an upper bound for) the size of the
+    /// signature is not known. Use [`Self::sign_into()`] if the input data
+    /// is entirely available in a single buffer (i.e. does not have to be
+    /// streamed).
+    //
+    // TODO: Is it possible to re-do this call if an incorrect buffer size
+    // was passed in? I think so?
+    pub fn sign_final_into(&mut self, sig: &mut [u8]) -> Result<usize> {
```
Jakuje: Is there a way to reuse the `sign_final()` we have? Or reuse this in the existing `sign_final()` function? It has a lot of duplicated code.

Similarly for the `sign_into()`.

cryptoki/src/session/signing_macing.rs
bal-e: These docs will be rendered in the documentation for `Session`, yes. You can try generating the documentation locally and taking a look.

cryptoki/src/session/signing_macing.rs
bal-e: I'm not sure what you mean. Could you elaborate / give an example?

cryptoki/src/session/signing_macing.rs
bal-e: The parentheses can help RustDoc disambiguate between declarations of the same name. In this case, it's not necessary for disambiguation, but I like leaving the parentheses in as they are visible in the generated docs, and make it more clear that I am referring to a function.

cryptoki/src/session/signing_macing.rs
bal-e: Actually, coming back to this: in case `RvError::BufferTooSmall` is returned, how will the user find out the correct size? Or should we deliberately omit that information, and force them to use the `Vec`-returning alternatives in those cases?

cryptoki/src/session/signing_macing.rs
bal-e: In `sign_final()`, yes, but not in `sign()` -- in `sign()`, the function call to discover the right signature size must happen after `C_SignInit`. In `sign_final()`, only the second `C_SignFinal` call could be reused, which IMO isn't worthwhile.

cryptoki/src/session/signing_macing.rs
hug-dev: I was just thinking to add:

```
and the resulting signature will be written to the given output buffer `sig`.
```

cryptoki/src/session/signing_macing.rs
hug-dev: > how will the user find out the correct size?

I would say that if the user is about to use those optimized functions then they would have done the previous research to know the exact size they want for their signature algorithms. So I think it's fine to leave them use the `Vec` variant otherwise.

cryptoki/src/session/signing_macing.rs
bal-e: Ah, got it. Will add.

cryptoki/src/session/signing_macing.rs
bal-e: Cool, I'll document that.

cryptoki/src/session/signing_macing.rs
```diff
@@ -51,11 +61,62 @@ impl Session {
...
+                sig.as_mut_ptr(),
+                &mut sig_len,
+            ))
+            .into_result(Function::Sign)?;
+        }
+
+        assert!(
+            sig_len <= sig_buf_len,
+            "'C_Sign' succeeded but increased 'sig_len', possibly indicating out-of-bounds accesses"
+        );
```
Jakuje, line 128: I think we should not use `assert!()` throughout the code, just in tests. It is either something that can not happen and then the check should not be there or it can happen and we should report proper error, regardless how much unlikely it would be or how much the pkcs11 module would have to be broken.

cryptoki/src/session/signing_macing.rs
```diff
@@ -113,11 +184,56 @@ impl Session {
...
+    /// (i.e. `sig[0..size]` contains the prepared signature). Otherwise,
+    /// [`RvError::BufferTooSmall`] will be returned.
+    ///
+    /// Use [`Self::sign_final()`] if (an upper bound for) the size of the
+    /// signature is not known. Use [`Self::sign_into()`] if the input data
+    /// is entirely available in a single buffer (i.e. does not have to be
+    /// streamed).
+    //
+    // TODO: Is it possible to re-do this call if an incorrect buffer size
+    // was passed in? I think so?
```
Jakuje: Yes, this should be possible per pkcs11 specs. The signFinal follows the 5.2 semantics:

https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.2/pkcs11-spec-v3.2.html#_Toc195693125

bal-e: I've now removed the TODO comment and documented that users should know an appropriate buffer size upfront (so they don't need to retry). See <https://github.com/parallaxsecond/rust-cryptoki/pull/349#discussion_r2821253089>.

cryptoki/src/session/signing_macing.rs
bal-e: I initially had a TODO comment here discussing whether an `assert` was appropriate. We could add some kind of `ImplementationError` variant to `crate::error::Error`, maybe that would be helpful for other functions too. @hug-dev, what do you think?

hug-dev: I don't think that PKCS11 implementation should increase the `sig_len` value while returning `CK_OK` (as per [the specs](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/os/pkcs11-spec-v3.1-os.html#_Toc111203252)). If they do then I would say that it's an error in the implementation but maybe not something that we should check in runtime.

So I would also be in favor of removing those asserts!

bal-e: My only concern is that user code will assume this property to be upheld. It would be better for `cryptoki` to forcibly crash than to introduce a buffer overflow in user code (e.g. if the user does `&sig[..sig_len]`, which loses bounds checks in release mode).

wiktor-k: AFAIK Rust still inserts bounds checks in release mode (source https://news.ycombinator.com/item?id=39261604) unless the optimizer can deduct that it's safe to remove (eg. in for loops).

Maybe yet another possibility would be to use `debug_assert`? 🤔

Jakuje: I don't know. If the pkcs11 module will return larger value than the buffer length, it likely already wrote beyond the buffer bounds, which is already issue and it means its completely off-specs. Asserting afterward, we just catch this potential mismatch.

Thinking about this further, its probably better to be safe than sorry if this is something we can catch. This additional check in sign operation is likely not something that would kill the performance. I still think returning an error is more appropriate than assert in any production code though.

wiktor-k: > I still think returning an error is more appropriate than assert in any production code though.

Yeah, asserts in production code are especially tricky. For really rare occasions even stdlib has panics but maybe we can "have a cookie and eat it too" here by introducing some error variant with just a string description that'd be a catch-all for all these assert-like bugs? :thinking:

hug-dev: > My only concern is that user code will assume this property to be upheld.

I think that is fair to assume (we can even write a warning notice about that in the readme). If we can't trust that the PKCS11 library implements the spec correctly then there is not much we can do :cry:

bal-e: > Yeah, asserts in production code are especially tricky.

I would normally agree, but I think "we think UB has occurred, but we don't want to panic, maybe everything is fine" is concerning. I don't know what the ideal solution is. I'm happy to implement whatever you (the maintainers) think should happen.

wiktor-k: > I'm happy to implement whatever you (the maintainers) think should happen.

That's great, we just have a little split brain problem here so it may take a while until we have a consensus:)

Personally I'm okay with an assert of a very unlikely scenario. I'm also fine with a new Error enum variant if that makes @Jakuje calm.

The "no check at all" scenario is the worse one IMO.

Jakuje: I agree so don't let me hold the PR back as it is a great improvement! I agree that this case really means the underlying token is broken so maybe the assert is adequate. I just know how annoying is when somewhere deep in the dependency chain of your built package you get assert and there is nothing you can do but rebuild.

hug-dev:
It's been a long time @bal-e but if you have the time you could rebase the PR and then we should merge as is since we agreed on going forward with the assert!

cryptoki/src/session/signing_macing.rs
hug-dev: Resolving this then, let's go with the assert to try to unblock this!

hug-dev (dismissed):

Jakuje (dismissed):
We need rebase, but otherwise I think we are good to merge this!

bal-e:
Awesome, on it!

hug-dev (approved):
Thank you!

wiktor-k (approved):
LGTM 👍

PR #362, opened 2026-03-16
ci: Update kryoptic features to unbreak CI

Jakuje:
The dummy-integrity is gone.


wiktor-k (approved):
It's okay with me. Could you elaborate on why this is using Simo's fork? (I guess this has to do with some patches unmerged upstream but I wonder if they had been proposed and will be available eventually...)

hug-dev (approved):

PR #363, opened 2026-05-07
test: Add Wycheproof-based AES-CBC-PKCS5 tests

### LINKED ISSUE #187: Add Wycheproof-based tests

testingapisname:
Implements comprehensive AES-CBC-PKCS5 testing using official Wycheproof test vectors from Google. Tests 216 valid cryptographic operations across multiple key sizes (128/192/256-bit) with PKCS#7 padding.

Fixes #187

### DIFF
--- a/cryptoki/tests/wycheproof.rs
+++ b/cryptoki/tests/wycheproof.rs
@@ -485,3 +485,151 @@ fn aes_gcm_message_wycheproof() -> TestResult {
 
     Ok(())
 }
+
+/// Test AES-CBC with PKCS5 padding using Wycheproof test vectors
+#[test]
+#[serial]
+fn aes_cbc_pkcs5_wycheproof() -> TestResult {
+    let (pkcs11, slot) = init_pins();
+    let session = pkcs11.open_rw_session(slot)?;
+    session.login(UserType::User, Some(&AuthPin::new(USER_PIN.into())))?;
+
+    // Load Wycheproof AES-CBC-PKCS5 test vectors
+    let test_set = wycheproof::cipher::TestSet::load(wycheproof::cipher::TestName::AesCbcPkcs5)?;
+
+    let mut passed = 0;
+    let mut failed = 0;
+    let mut skipped = 0;
+
+    for test_group in &test_set.test_groups {
+        let key_size = test_group.key_size;
+
+        // Only test key sizes we support (128, 192, 256 bits)
+        if ![128, 192, 256].contains(&key_size) {
+            skipped += test_group.tests.len();
+            continue;
+        }
+
+        for test in &test_group.tests {
+            // Skip tests with nonce (IV) sizes other than 16 bytes (AES block size)
+            if test.nonce.len() != 16 {
+                skipped += 1;
+                continue;
+            }
+
+            // Import the test key
+            let key_template = vec![
+                Attribute::Class(cryptoki::object::ObjectClass::SECRET_KEY),
+                Attribute::KeyType(cryptoki::object::KeyType::AES),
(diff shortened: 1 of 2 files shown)


cryptoki/tests/wycheproof.rs
```diff
@@ -485,3 +485,151 @@ fn aes_gcm_message_wycheproof() -> TestResult {
...
+    println!(
+        "AES-CBC-PKCS5 Wycheproof results: {} passed, {} failed, {} skipped",
+        passed, failed, skipped
+    );
+
+    // The main requirement is that Valid tests pass
+    assert_eq!(failed, 0, "Some valid Wycheproof tests failed");
+
+    session.close()?;
+    pkcs11.finalize()?;
```
wiktor-k: I think it's better to close and *then* assert, since if the assert panics the finalization will not happen (maybe I'm paranoid 😅).

```suggestion
    session.close()?;
    pkcs11.finalize()?;

    // The main requirement is that Valid tests pass
    assert_eq!(failed, 0, "Some valid Wycheproof tests failed");
```

cryptoki/tests/wycheproof.rs
```diff
@@ -485,3 +485,164 @@ fn aes_gcm_message_wycheproof() -> TestResult {
...
+                        test.tc_id,
+                        test.result,
+                        key_size,
+                        test.nonce.len(),
+                        test.pt.len()
+                    );
+                    passed += 1;
+                }
+                // Invalid test that succeeded - Note: HSM may not catch all invalid cases
+                (wycheproof::TestResult::Invalid, Ok(_)) => {
```
hug-dev, line 593: Does that actually happen with some of the test cases? Maybe it would be safe to double-check those and filter-out on invalid cases we do not support?

cryptoki/tests/wycheproof.rs
```diff
@@ -485,3 +485,164 @@ fn aes_gcm_message_wycheproof() -> TestResult {
...
+                        "  Key size: {}, IV len: {}, PT len: {}, CT len: {}",
+                        key_size,
+                        test.nonce.len(),
+                        test.pt.len(),
+                        test.ct.len()
+                    );
+                    failed += 1;
+                }
+                // Acceptable tests can go either way
+                (wycheproof::TestResult::Acceptable, Ok(_)) => {
```
hug-dev, line 617: Can we move the `|` condition on `Acceptable` above and move it here like:

```rust
// Acceptable tests can go either way
(wycheproof::TestResult::Acceptable, Ok(_) | Err(_)) => {
```

hug-dev (commented):
Thank you for adding those!

cryptoki/tests/wycheproof.rs
Jakuje: Going through the source file for this input, I do not see any tests marked Acceptable:

https://github.com/randombit/wycheproof-rs/blob/master/src/data/aes_cbc_pkcs5_test.json

Could we prune the branches that are not executed?

cryptoki/tests/wycheproof.rs
Jakuje: I think most of the Invalid tasks have BadPadding flag. which can be used to filter it more precisely.

https://github.com/randombit/wycheproof-rs/blob/master/src/data/aes_cbc_pkcs5_test.json#L424

PR #365, opened 2026-08-20
Add Mechanism::AesCtr

sietseringers:
Adds support for the `CKM_AES_CTR` mechanism, which was already exposed as a `MechanismType` but could not be used as a `Mechanism` because its parameter type had no safe wrapper.

I am interested in this because I am thinking of basing an implementation of AES-SIV ([RFC 5297](https://datatracker.ietf.org/doc/html/rfc5297)) on top of this, which is based on AES-CTR.

## Changes

- New module defining `AesCtrParams`, which is a `#[repr(transparent)]` wrapper around `CK_AES_CTR_PARAMS`.
- New `Mechanism::AesCtr(AesCtrParams)` variant, wired into `Mechanism::mechanism_type()`.

### DIFF
--- a/cryptoki/src/mechanism/aes_ctr.rs
+++ b/cryptoki/src/mechanism/aes_ctr.rs
@@ -28,6 +28,7 @@ impl AesCtrParams {
     /// bits.
     ///
     /// # Errors
+    ///
     /// This function returns [`Error::InvalidValue`] if `counter_bits` is zero
     /// or larger than the AES block size of 128 bits.
     pub fn new(counter_bits: u8, block: [u8; 16]) -> Result<Self, Error> {
--- a/cryptoki/src/mechanism/aes_ctr.rs
+++ b/cryptoki/src/mechanism/aes_ctr.rs
@@ -1,0 +1,45 @@
+// Copyright 2026 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+//! AES-CTR mechanism types
+
+use crate::error::Error;
+use cryptoki_sys::*;
+
+/// Parameters for AES-CTR.
+///
+/// This structure wraps a `CK_AES_CTR_PARAMS` structure.
+#[derive(Debug, Clone, Copy)]
+#[repr(transparent)]
+pub struct AesCtrParams {
+    inner: CK_AES_CTR_PARAMS,
+}
+
+impl AesCtrParams {
+    /// Construct AES-CTR parameters.
+    ///
+    /// # Arguments
+    ///
+    /// `counter_bits` - The number of least significant bits of `block` that
+    /// make up the counter, i.e. that are incremented for every block that is
+    /// processed. Must be between 1 and 128.
+    ///
+    /// `block` - The initial value of the counter block: the nonce, followed by
+    /// the initial value of the counter in the `counter_bits` least significant
+    /// bits.
+    ///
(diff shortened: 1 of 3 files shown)


cryptoki/src/mechanism/aes_ctr.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// `block` - The initial value of the counter block: the nonce, followed by
+    /// the initial value of the counter in the `counter_bits` least significant
+    /// bits.
+    ///
+    /// # Errors
+    /// This function returns [`Error::InvalidValue`] if `counter_bits` is zero
+    /// or larger than the AES block size of 128 bits.
+    pub fn new(counter_bits: u8, block: [u8; 16]) -> Result<Self, Error> {
+        if counter_bits == 0 || counter_bits > 128 {
+            return Err(Error::InvalidValue);
```
hug-dev, line 36: Maybe here we could return the wrapped `CKR_MECHANISM_PARAM_INVALID` as stated in the documentation, what do you think?

hug-dev (commented):
Thank you! Looks good with a small comment. You might need to sign-off your commit  (`git commit --signoff`) for the DCO check to pass :)

cryptoki/src/mechanism/aes_ctr.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// `counter_bits` - The number of least significant bits of `block` that
+    /// make up the counter, i.e. that are incremented for every block that is
+    /// processed. Must be between 1 and 128.
+    ///
+    /// `block` - The initial value of the counter block: the nonce, followed by
+    /// the initial value of the counter in the `counter_bits` least significant
+    /// bits.
+    ///
+    /// # Errors
+    /// This function returns [`Error::InvalidValue`] if `counter_bits` is zero
```
wiktor-k, line 32:
```suggestion
    ///
    /// This function returns [`Error::InvalidValue`] if `counter_bits` is zero
```

wiktor-k (dismissed):
Looks nice, thanks! Of course @hug-dev remarks stand :)

cryptoki/src/mechanism/aes_ctr.rs
sietseringers: Do you mean returning e.g. `Error::Pkcs11(RvError::MechanismParamInvalid, Function::EncryptInit)` instead of `Error::InvalidValue`?

That is certainly a possibility, and I'd be happy to do that if you'd prefer it. However, that second parameter would in this case not correspond with any function that we've actually called at this point, whereas every other `Error::Pkcs11` in the crate seems to come from an actual return value via `Rv::into_result`. Also, this constructor would also be called during decryption, so `Function::DecryptInit` for example might be just as applicable; in other words, it's not clear to me which `Function` variant would be most applicable. Finally,  `Error::InvalidValue` is what the crate already uses for client-side validation elsewhere (`mechanism/rsa.rs:70`, `object.rs:1974`, `types.rs:34`).

hug-dev: This is what I meant and your explanation completely makes sense, let's keep it this way!

hug-dev (approved):

wiktor-k (approved):
Let's go 🚀

sietseringers:
Great, thanks!

